PR #323, opened 2021-04-21
Optimize the logo

rictic:
- fewer paths
- all straight lines, no arcs or beziers, just triangles

a combination of `svgo` and hand optimizations

renders identically for me in Chrome, Firefox, and Safari desktop


justinfagnani:
@rictic @aomarks what do we want to do with this PR?

Regarding the technique, is there a resource on writing simple SVGs without gaps between adjacent shapes? Is adding a bulge a common thing?

aomarks:
I think https://stackoverflow.com/a/53312630 was probably where I found the bulge technique. There are some other solutions proposed there, but none of them worked for our situation.

We could definitely optimize the bulges, I just did them quickly in inkscape. Maybe close this PR and file an issue?

aomarks:
Closing for now.

PR #343, opened 2021-04-23
Update directives.md

Christian24:
Added assertion on ref directive examples. RefCallback can be undefined, so added optional chaining in the second example.

### DIFF
--- a/packages/lit-dev-content/site/docs/templates/directives.md
+++ b/packages/lit-dev-content/site/docs/templates/directives.md
@@ -810,7 +810,7 @@ class MyElement extends LitElement {
   }
 
   firstUpdated() {
-    const input = this.inputRef.value;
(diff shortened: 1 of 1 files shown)


Christian24:
Is there something I have to do here still or can this only be done by maintainers?

e111077:
cc: @arthurevans

arthurevans:
Hi @Christian24. Sorry for the delay. A lot of us were either on vacation or sick the week after the release and we dropped the ball here.  Looks good to me but I'm going to ask @kevinpschaaf to take a look.

e111077:
It seems like the new github actions rules to curb github actions abuse might be interfering with how our rules are set up

arthurevans:
I don't know what the CI bot's problem with this is, but I'm going to give up and merge it...

PR #353, opened 2021-04-24
Fix typos in examples on shadow-dom page

lucaseverett:
The two "equivalent to" examples are missing an underscore.


lucaseverett:
@googlebot I signed it!

aomarks (approved):

aomarks:
Thanks

PR #354, opened 2021-04-26
Fix typo in polyfill-support sample code

butterflyhug:
I assume this is a typo because the surrounding context consistently calls this file "polyfill-support.js" instead of "platform-support.js".


justinfagnani (approved):
Thanks for catching this!

PR #358, opened 2021-04-29
Update static expression docs, include example imports

### LINKED ISSUE #357: Document how to import static template support
https://lit.dev/docs/templates/expressions/#static-expressions mentions a `literal` tag and `unsafeStatic` but I can't find anywhere how to import it.

It seems to be accessible via `import {literal, unsafeStatic} from 'lit/static-html.js`.

Should it be exported and added to the docs?

justinfagnani:
Fixes #357 by including some import examples. I also try to clarify a few things in the section.

### DIFF
--- a/packages/lit-dev-content/site/docs/templates/expressions.md
+++ b/packages/lit-dev-content/site/docs/templates/expressions.md
@@ -400,7 +400,7 @@ After static values have been interpolated, the template must be well-formed lik
 
 ### Non-literal statics
 
-In rare cases, you may need to interpolate static HTML into a template that is not defined in your script, and thus cannot be tagged with the `literal` function. For these cases, the `unsafeStatic()` function can be used to create static HTML based on strings from non-script source.
+In rare cases, you may need to interpolate static HTML into a template that is not defined in your script, and thus cannot be tagged with the `literal` function. For these cases, the `unsafeStatic()` function can be used to create static HTML based on strings from non-script sources.
 
 ```ts
 import {html, unsafeStatic} from 'lit/static-html.js';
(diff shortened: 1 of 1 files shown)


justinfagnani:
I just noticed that a big customer is planning on using static templates everywhere in order to parameterize what are just constant tag names. I think we need to add a bigger perf warning to the docs and API docs.

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -375,18 +386,26 @@ class MyAnchor extends MyButton {
...
 
 For these reasons, it's a good idea keep changes to expressions using `literal` to a minimum and avoid using reactive properties to change `literal` values, since reactive properties are intended to change.
 
+### Template structure
+
+After static values have been interpolated, the template must be well-formed like normal Lit templates, otherwise the dynamic expressions in the template might not function properly. See the [Well-formed HTML](#well-formed-html) section for more information.
+
 ### Non-literal statics
 
 In rare cases, you may need to interpolate static HTML into a template that is not defined in your script, and thus cannot be tagged with the `literal` function. For these cases, the `unsafeStatic()` function can be used to create static HTML based on strings from non-script source.
```
arthurevans, line 403:
```suggestion
In rare cases, you may need to interpolate static HTML into a template that is not defined in your script, and thus cannot be tagged with the `literal` function. For these cases, the `unsafeStatic()` function can be used to create static HTML based on strings from non-script sources.
```
should be "a non-script source" or "non-script sources" I think.

arthurevans (approved):
One minor comment, otherwise LGTM. Did you want to add a warning to this PR?

I'd suggest adding a new L3 heading called "Performance impact" or something like that so it's right in the TOC.

justinfagnani:
Thanks! I'll add the warning in another PR

PR #374, opened 2021-05-13
doc: correct shadowRootOptions spread syntax

colbywhite:
`super` can't be used in the format indicated in the doc. this corrects the example to properly reference the parent class.


aomarks (approved):
Good catch thanks!

PR #377, opened 2021-05-17
Fix broken import examples in upgrade.md

Haprog:
Some of the import examples were using backticks (template literal) instead of quotes (string literal) which doesn't work for imports.

With the backticks in import path TypeScript gives this kind of error:
```
src/my-element.ts:8:39 - error TS2307: Cannot find module 'lit/decorators.js'.
8 import {customElement, property} from `lit/decorators.js`;
```

And VS Code also highlight the problem with a message "String literal expected".


simi (approved):
:muscle:

AndrewJakubowicz (approved):
Thank you!

PR #383, opened 2021-05-20
Use timeout in clock-controller

FelixSchuSi:
In the clock-controller example the setTimeout function is used to regularly update the current time.
It seems that setTimeout is unintentionally being used without the timout parameter despite a separate `timeout` filed being present.


arthurevans (approved):
Looks like an oversight. Tested and confirmed it's running crazy without the timeout there.

PR #386, opened 2021-05-26
Remove invalid closing tag from input element

Haprog:
`<input>` is a void element.

> Void elements only have a start tag; end tags must not be specified for void elements.

https://html.spec.whatwg.org/multipage/syntax.html#elements-2

Fixes invalid HTML in an example shown at https://lit.dev/docs/templates/expressions/#property-expressions

Changes:
```ts
html`<input .value=${this.itemCount}></input>`;
```

to:
```ts
html`<input .value=${this.itemCount}>`;
```

This also makes the example consistent with the few other examples that contain an `<input>` element on the same page.


aomarks (approved):
Thanks!

PR #387, opened 2021-05-28
Update upgrade.md

bschlenk:
Mention that `ifDefined` now removes attributes for both `null` and `undefined`, and that `nothing` can be rendered in an attribute expression.

Closes https://github.com/lit/lit/issues/1927


arthurevans (approved):
LGTM, thanks.

PR #390, opened 2021-06-04
doc: Fix typo.

ejez:
Fix typo in `packages/lit-dev-content/site/docs/components/properties.md`.


arthurevans:
Thanks for the PR. This whole list was really wrong--"defined and not-null" makes sense for properties, not so much for attributes. The actual test here is whether the element has the given attribute. Fixed this description in a separate PR.

PR #401, opened 2021-07-05
A couple of sample-fixes

filipbech:
Seems the clock-controller uses .value and not .time to store the time. Also the dual-clock controller probably wants to save the controller instances on the element-instance. 

For reference the clock-controller code is here: https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/samples/docs/controllers/overview/clock-controller.ts


arthurevans (approved):
LGTM. Thanks for the fix!

PR #402, opened 2021-07-08
docs: add Pharos Design System to who-is-using

Niznikr:
Add Pharos Design System to who-is-using with logo


Niznikr:
@googlebot I signed it!

packages/lit-dev-content/site/home/5-who-is-using.html
```diff
@@ -151,6 +151,16 @@ <h1>
           <span>Wired Elements</span>
         </div>
       </a>
+
+      <a target="_blank" rel="noopener" href="https://github.com/ithaka/pharos#:~:text=lit-element">
```
arthurevans, line 155: This is very cool--thanks for the PR! One thing we ask is that the link points to a mention of Lit or LitElement. This could be a mention in the docs, a blog post, or even a pointer to something on GitHub, like this:
```suggestion
      <a target="_blank" rel="noopener" href="https://github.com/ithaka/pharos/blob/main/packages/pharos/CHANGELOG.md#:~:text=Lit">
```
I thought I'd found a section on your docs that mentioned Lit, but I can't find it anymore.

Niznikr: Makes sense! We mention in our developer docs [here](https://github.com/ithaka/pharos/blob/70f42ef0776d1b0ff0ea32ae6560deef92757d8d/docs/development/anatomy-of-a-component.md#lit) if that's better.

Niznikr: The one currently in the PR points to a `lit-element` tag we have on the repo

arthurevans: Perfect! That's what I was looking for.

arthurevans: I missed both your second comment and the tag on the repo... But I think the other link is better, anyway. This should be live now.

Issue #371, opened 2021-05-12
Unify styles of single/multi file inline examples


aomarks:
Currently there are two kinds of inline playground examples: single file, and multi file. Currently, the two look quite different, which is a bit odd.

multi file:
<img src="https://user-images.githubusercontent.com/48894/118025611-e3f2f380-b314-11eb-91a3-b5250caba879.png" width="70%">

single file:
<img src="https://user-images.githubusercontent.com/48894/118025652-f0774c00-b314-11eb-86bf-cefa2208ef6b.png" width="70%">

Probably the multi-file version should look like the single file version, with the preview below, but just with tabs on the top. We could also consider an arrangement where multiple files are visible at once, e.g. HTML and CSS next to each other, with preview below.

PR #404, opened 2021-07-08
Unify styles of single/multi file inline examples

AndrewJakubowicz:
Address issue #371, using our `litdev-example` component for both `playground-ide` and `playground-example` shortcodes.

### Context

The multi-file and single file examples have very different layouts. This PR updates the multi-file example to have the same style and layout as the single file example.

An additional benefit is there is now a playground link button in the multi file examples.

### How

Unify the style by using the same underlying component for both example shortcodes. The primary difference between the two shortcodes is whether or not there is a file tab bar displayed. Thus we handle both cases through the presence of the parameter `filename`.

This PR also removed a line-number parameter which isn't used, and updated some example config heights to remove a tiny amount of scrolling.

### Testing

Tested manually with `npm run dev`. Went page by page tweaking some of the heights to better reflect the new layout. Firefox has a larger font line-height, so most of the height changes were minor. Most of the changes were to allow a larger preview. 

I also ran `npm run test` but I don't believe we test the playground examples.
Please let me know if there are any automated tests I can add or include in this PR.

Manually tested in browsers:
 - [x] Chrome
 - [x] Firefox
 - [x] Safari

Manually tested narrow screen width with horizontal scrolling of file tabs.

### Screenshots of change

Screen Shot 2021-07-08 at 4 42 36 PM

Screen Shot 2021-07-08 at 4 43 13 PM

Screen Shot 2021-07-08 at 5 13 47 PM

### DIFF
--- a/packages/lit-dev-content/samples/docs/components/events/comm/project.json
+++ b/packages/lit-dev-content/samples/docs/components/events/comm/project.json
@@ -5,5 +5,6 @@
     "my-listener.ts": {},
     "index.html": {}
   },
-  "editorHeight": "920px"
+  "editorHeight": "760px",
+  "previewHeight": "180px"
 }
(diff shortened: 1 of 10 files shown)


aomarks (approved):
Looks great! LGTM

PR #407, opened 2021-07-14
Bump playground/MWC versions

aomarks:
Brings in these improvements: https://github.com/PolymerLabs/playground-elements/blob/main/CHANGELOG.md#0100---2021-07-01

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,29 +17,38 @@
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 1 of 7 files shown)


e111077 (approved):

arthurevans:
The service worker doesn't seem to be updating smoothly here. As @justinfagnani reported in chat, seems to require manually unregistering the service worker.

PR #408, opened 2021-07-15
Update package-lock with newest patch versions

AndrewJakubowicz:
## Issue

Originally on node 14 I was getting issues with `npm i` and `npm run build` due to the playground having `-pre.1` in the lock.

`^0.10.1` in package.json wasn't finding the `-pre.1` version in the package lock.

## Fix

Run `npm run nuke`, with environment:

```
$ node -v
v16.5.0

$ npm -v
7.19.1
```

## Testing

This is a package-lock change, so will check manually in the preview that playground works and looks correct.

 - [x] Checked manually.

## Risks

There are some other libraries updated in the package locks.

Open questions, I am not sure why my package-lock has updated in order to build. This PR may not be required at all.

### DIFF
--- a/packages/lit-dev-content/package-lock.json
+++ b/packages/lit-dev-content/package-lock.json
@@ -10,17 +10,17 @@
 			"dependencies": {
 				"@lit-labs/task": "1.0.0-pre.2",
 				"@lit/localize": "^0.10.0",
-				"@material/mwc-button": "0.22.0-canary.8cba18fa.0",
-				"@material/mwc-drawer": "0.22.0-canary.8cba18fa.0",
-				"@material/mwc-formfield": "0.22.0-canary.8cba18fa.0",
-				"@material/mwc-icon-button": "0.22.0-canary.8cba18fa.0",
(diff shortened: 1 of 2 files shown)


aomarks (approved):
Nice, I think this was my bad in https://github.com/lit/lit.dev/pull/407. I was testing with some pre-release versions initially, and then after I published MWC I bumped to the final releases, but forgot to regenerate the package locks.

I think this means the current site is using the pre-releases, because `npm ci` will take the versions from the `package-lock.json`, rather than the latest compatible.

AndrewJakubowicz:
In lit.dev/docs the example is not loading:
Screen Shot 2021-07-15 at 2 45 30 PM

It loads in the preview site, so going to merge and hope it's all good.

aomarks:
> It loads in the preview site, so going to merge and hope it's all good.

Actually this is because of some problem where service workers don't get updated in a timely manner, and there's a breaking change between the previous and current versions of the elements/service worker. Kinda bad! Will need to look into this very soon. For now you can open up dev tools/Application and click "Unregister" on the service worker to force it to reload.

But you're still good to merge this!

Screen Shot 2021-07-15 at 2 48 24 PM

Issue #411, opened 2021-07-16
typing in lit.dev playground can lock up entire tab or browser


aomarks:
1. Open https://lit.dev/playground/ to load the default "Hello World (TypeScript)" example.

2. Paste the following code into the class (after the `name` property):
```ts
constructor() {
  super();
  this.go();
}
```

3. *Type* (don't paste) the following code after that:
```ts
async go() {
  for |
}
```

4. If the browser does not lock up, try deleting/re-typing the code around the `|`. Eventually the tab will lock up. For @justinfagnani this also caused the entire *origin* to lock up, even after closing/opening tabs!

AndrewJakubowicz:
The rollback also fixes the two styling issues #409 and #410 . Should I close these issues since they are no longer relevant and just need to be kept in mind when we re-upgrade.

aomarks:
> The rollback also fixes the two styling issues #409 and #410 . Should I close these issues since they are no longer relevant and just need to be kept in mind when we re-upgrade.

Let's keep them because they need to be addressed still. Thanks for filling!

aomarks:
Simpler repro:

https://lit.dev/playground/#project=W3sibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPHNjcmlwdD5cbiAgLy8gdW5jb21tZW50IHRvIGxvY2tcbiAgLy8gZm9yKDs7KTtcbjwvc2NyaXB0PiJ9XQ

The main issue here seems to be that if the iframe locks up, the parent frame does too. This shouldn't happen, because on lit.dev we run the playground iframe on a separate origin (playground.lit.dev), so it should get its own thread.

The reason the specific repro in the issue description caused this error is that TypeScript compiles `{ for }` to `{ for (;;); }` -- so it transforms a runtime syntax error into a valid infinite loop (https://www.typescriptlang.org/play?#code/N4AgZg9gTiC+Q)!

aomarks:
> This shouldn't happen, because on lit.dev we run the playground iframe on a separate origin (playground.lit.dev), so it should get its own thread.

Turns out process isolation is per *site*, not per *origin*, by default (https://www.chromium.org/developers/design-documents/site-isolation#TOC-Threat-Model), so lit.dev and playground.lit.dev share a process.

However there is a new `Origin-Agent-Cluster` header that we should be able to use to switch to per *origin*, so that they would be isolated: https://web.dev/origin-agent-cluster/. Trying that out now!

PR #418, opened 2021-07-23
Set Origin-Agent-Cluster on playground.lit.dev

aomarks:
This allows Chrome to use a separate process to run Playground iframes. This should improve performance overall, and will also prevent e.g. an infinite loop from locking up the entire site.

See https://github.com/PolymerLabs/playground-elements/pull/179 and https://web.dev/origin-agent-cluster/ for more context.

Fixes https://github.com/lit/lit.dev/issues/411

### DIFF
--- a/packages/lit-dev-server/src/index.ts
+++ b/packages/lit-dev-server/src/index.ts
@@ -33,6 +33,15 @@ console.log(`port: ${port}`);
 console.log(`static root: ${staticRoot}`);
 
 const app = new Koa();
+
+if (mode === 'playground') {
+  // See https://github.com/PolymerLabs/playground-elements#process-isolation
+  app.use(async (ctx, next) => {
+    ctx.set('Origin-Agent-Cluster', '?1');
+    await next();
+  });
+}
+
 app.use(async (ctx, next) => {
   if (ctx.path.match(/\/[^\/\.]+$/)) {
     // Canonicalize paths to have a trailing slash, except for files with
@@ -54,12 +63,13 @@ app.use(async (ctx, next) => {
     ctx.status = 301;
     ctx.redirect(
       pageRedirects.get(ctx.path) +
-      (ctx.querystring ? '?' + ctx.querystring : '')
+        (ctx.querystring ? '?' + ctx.querystring : '')
     );
   } else {
     await next();
   }
 });
+
 app.use(koaConditionalGet()); // Needed for etag
 app.use(koaEtag());
 app.use(


AndrewJakubowicz (approved):
This is far better than the alternative - a whole page lock up. What an amazing find! I played around with it on the preview site.

Some questions whilst playing with this (recording linked below).

 - Is there a solution that can work on other browsers?
 - Would it be possible to make the refresh button save the user from a possible loop? Currently it seems that a full page refresh is required to get the preview to respond again.

Awesome!!! And TIL about Origin-Agent-Cluster!

https://user-images.githubusercontent.com/15080861/126744756-077d8448-8eb2-4483-ab06-a35da5938e61.mp4

justinfagnani (approved):

aomarks:
> * Is there a solution that can work on other browsers?

Neither Firefox nor Safari currently support out-of-process iframes at all (even with different-site).

Firefox does have experimental support behind a flag (`about:config / fission.autostart`) to switch on Project Fission, which adds support (https://wiki.mozilla.org/Project_Fission, https://bugzilla.mozilla.org/show_bug.cgi?id=1451850). I tested it out, and it works! There's also a bug filed to implement `Origin-Agent-Cluster` (https://bugzilla.mozilla.org/show_bug.cgi?id=1665474). Firefox does have a nice related feature, which is that after a minute or so of any page being blocked, it will prompt you to stop execution.

Safari doesn't seem to support out-of-process iframes at all, and I couldn't find any bug filed for it at https://bugs.webkit.org/ (but I definitely could have missed it).

I also was curious how Glitch works in these cases, and it's exactly the same as Playground will now behave (Glitch uses a different-site for the preview iframe): https://glitch.com/edit/#!/humdrum-aware-kitty?path=index.html%3A14%3A9

> * Would it be possible to make the refresh button save the user from a possible loop? Currently it seems that a full page refresh is required to get the preview to respond again.

I think we can *detect* a lockup like this:

1. Maintain a second prober iframe with the same origin as the preview, which serves a static document that simply acknowledges `postMessage` pings.
2. Ping the prober iframe every N seconds. If we don't get a response after some timeout, assume that the preview must have locked up that origin's process/thread.

From some experimenting, I don't *think* there is any way to kill the runaway process once it's started. It will lock up that origin until the parent window reloads. Even if the culprit iframe is removed from the DOM, the process continues running (so if you create a new iframe with the same origin, it will never load). Something to look into, and ask about!

Maybe the best we can do after detecting a lockup is to show an alert to the user with a button that saves the current playground state, and reloads the tab.

There is a "solution" I thought of, but it has its own fatal problem:

1. Set up wildcard DNS for `*.lit.dev` (or maybe register a new TLD for this purpose).
2. Use a sandbox origin like `playground-0.lit.dev` instead of the static `playground.lit.dev` we have now.
3. Every time we detect a lockup (using the prober iframe approach above), increment the number in the origin (e.g. `playground-1.lit.dev`) and re-initialize the project again on the new origin. The new origin should get a separate process

The big problem with this is that each runaway process would just continue in the background forever, so we'll end up consuming all CPU resources eventually. (Barring any interventions the browser might do to detect runaway processes? Firefox does actually prompt to stop processes after a minute or so).

Issue #409, opened 2021-07-15
[Styles] Floating resize bar issue on tutorial page


AndrewJakubowicz:
## Issue

The background above the preview window doesn't extend up enough making it counter-intuitive to locate the resize bar.

## Repro

Navigate to  https://lit.dev/tutorial/ and hover over the horizontal resize for the preview. Note that it is located in the right place, but there is a transparent patch beneath it.

See image, the light blue bar is the resize bar:

![Screen Shot 2021-07-15 at 3 44 29 PM](https://user-images.githubusercontent.com/15080861/125867164-8425b2bc-4fed-42ea-b0e8-b3a5d8644f3f.png)

Issue #410, opened 2021-07-15
[Styles] Documentation Playground Examples slight height mismatch


AndrewJakubowicz:
## Issue

Within the editable code example, the div `class="CodeMirror cm-s-default"` is slightly taller than its parent and leaks into the preview pane. It's just enough to cover the border slightly.

## Repro

Navigate to https://lit.dev/docs/ and scroll down to the example.
![Screen Shot 2021-07-15 at 3 47 58 PM](https://user-images.githubusercontent.com/15080861/125867565-2af5a786-ea39-4cf9-acb7-a65ef932605b.png)

Note the border is truncated by something.

I think the playground styles may have very subtly changed in the last update causing this minor style regression.

PR #419, opened 2021-07-28
Bump playground and fix styling issues

aomarks:
- Bump playground-elements from `0.9.3` to `0.11.0`. This includes the change that should hopefully make service worker upgrades immediate (https://github.com/PolymerLabs/playground-elements/pull/178). See https://github.com/PolymerLabs/playground-elements/blob/main/CHANGELOG.md#094---2021-05-18 onwards for all changes.

- Bump MWC elements to match playground's version.

- Fix https://github.com/lit/lit.dev/issues/409 and https://github.com/lit/lit.dev/issues/410 styling regressions which occurred the last time I tried to upgrade playground-elements. Turns out this was because a new CSS custom property `--playground-code-padding` was added in https://github.com/PolymerLabs/playground-elements/pull/161, but lit.dev was already using a property with that exact name to set the padding of static SSR'd code samples, and the two behave slightly differently. The fix was simply to rename the lit.dev specific property so that it didn't collide.

- Changes dev mode to always use the locally installed version of the playground service worker, so that we can test local changes to the playground service worker.

- Also tweaks some borders and typography in the tutorial/playground to look a bit sharper.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,29 +17,38 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.12.13",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
-      "integrity": "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==",
+      "version": "7.14.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
+      "integrity": "sha512-9pzDqyc6OLDaqe+zbACgFkb6fKMNG6CObKpnYXChRsvYGyEdc7CA2BaqeOM+vOtCS5ndmJicPJhKAwYRI6UfFw==",
       "dev": true,
       "dependencies": {
-        "@babel/highlight": "^7.12.13"
+        "@babel/highlight": "^7.14.5"
+      },
+      "engines": {
+        "node": ">=6.9.0"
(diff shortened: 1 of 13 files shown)


AndrewJakubowicz (approved):
Sanity checked the playground page. Tested by adding a `while(true) {};` loop. Was able to keep editing and a refresh fixed it immediately.

Also checked some documentation pages and a quick switch to mobile view.  Tested lightly on Chrome and Firefox.

Noted this warning in the console:
`The page requested an origin-keyed agent cluster using the Origin-Agent-Cluster header, but could not be origin-keyed since the origin '<URL>' had previously been placed in a site-keyed agent cluster. Update your headers to uniformly request origin-keying for all pages on the origin.`.
Unsure if that is due to the change in headers when testing the preview from what is currently on stable lit.dev. Doesn't seem to have an adverse impact.

LGTM

aomarks:
> Noted this warning in the console:
> `The page requested an origin-keyed agent cluster using the Origin-Agent-Cluster header, but could not be origin-keyed since the origin '<URL>' had previously been placed in a site-keyed agent cluster. Update your headers to uniformly request origin-keying for all pages on the origin.`.
> Unsure if that is due to the change in headers when testing the preview from what is currently on stable lit.dev. Doesn't seem to have an adverse impact.

Oops. I forgot to pull main before creating this branch, so this branch wasn't including https://github.com/lit/lit.dev/pull/418 which sets the `origin-agent-cluster` header to server responses. So while service worker responses had the header (because of the playground-elements version bump), the actual server responses didn't, hence the warning about inconsistencies.

Just merged main into this branch to confirm. The new preview URL doesn't show the warning. After merging this, the live lit.dev site might show the warning for some time -- possibly until the next time the user reloads their browser (not quite sure of the timing yet).

PR #422, opened 2021-08-05
Remove 404 MDN ParentNode link

aomarks:
MDN doesn't have a page for ParentNode. In the HTML spec it's an interface mixin (https://dom.spec.whatwg.org/#parentnode), and in TypeScript it's an interface type. Neither of these seem like very friendly or important things to link to, so I've just removed the link.


arthurevans (approved):
LGTM. Thanks Al!

PR #427, opened 2021-08-09
POC syndicated blog entries, tags.

arthurevans:
Includes one of Rody's blog posts. There are some issues with the inline samples, and we may want to eliminate the separate "all tags" page in favor of just linking back to the blog index.


arthurevans:
Grr. Missed some files in the first commit.

arthurevans:
There's a link on the left under the author credit: "Reposted from rodydavis.com" ...

There are several ways we could make this more prominent: move it up, change the typeface, color, etc.

I thought about putting it a the head of the main column, but the way the blog is designed, all the metadata is on the left, so that kind of felt like the right place for it.

Issue #428, opened 2021-08-09
Lit.dev search should close drawer on mobile after pressing a result


AndrewJakubowicz:
Justin identified that on mobile the mobile nav drawer doesn't close when you press on a search result. Identified as a shipping blocker.

AndrewJakubowicz:
Fixed by #430

PR #430, opened 2021-08-09
Close mobile nav drawer and clear input on search result navigation.

AndrewJakubowicz:
## Context

Closes issue #428 as well as fixing the input not properly clearing after navigation.

### Why?

When navigating between page fragment identifiers the page doesn't do a full reload. Thus the search input or mobile drawer don't reset to their initial states.

This change manually cleans up the mobile drawer and search input when a navigation has occurred.

### How?

Although we were setting `this.input.value = ''`, this is not a reactive property and thus the input wasn't being cleared. Requesting an update commits the input clearing.

Similarly we also query for the mobile drawer and close it.

### Testing

To reproduce the issue, navigate to https://lit.dev?mods=search and search between two fragments on the same page.
See video reproducing issue (also sorry for the wild aspect ratio stretching):

https://user-images.githubusercontent.com/15080861/128763409-e492b1ae-34dd-4d56-9830-c9f67b5e227f.mp4

This was manually tested.

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-search.ts
+++ b/packages/lit-dev-content/src/components/litdev-search.ts
@@ -17,6 +17,7 @@ import Minisearch from 'minisearch';
 
 import '@lion/combobox/define';
 import {LionOption} from '@lion/listbox';
+import type {Drawer} from '@material/mwc-drawer';
 
 /**
  * Representation of each document indexed by Minisearch.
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -289,6 +290,18 @@ class LitDevSearch extends LitElement {
...
+    // open when navigating between fragment identifiers.
+    const navDrawer = document.querySelector(
+      'mwc-drawer#mobileDrawer'
+    ) as Drawer;
+    if (navDrawer) {
+      navDrawer.open = false;
+    }
+
+    // Rerender to visually clear the value on the input.
+    this.requestUpdate();
```
aomarks, line 304: Instead of calling `requestUpdate` here, could you make `searchText` a `@state` property so that when you set it above, it automatically re-renders?

AndrewJakubowicz: Great suggestion. I don't know why I didn't reach for the declarative approach first. Thank you!

aomarks (approved):
Nice, LGTM! Tried it out with the latest deploy link and the reported issue is fixed.

PR #436, opened 2021-08-17
Sync disconnection-related docs with changes in lit/#2034

kevinpschaaf:
See https://github.com/lit/lit/pull/2034, which reverts the change in Lit 2.0 that caused LitElement lifecycle to be paused while the element was disconnected. This was deemed too risky of a change for existing code, and was reverted, hence it is no longer an upgrade-guide topic, and the docs on the lifecycle and custom directives pages are updated.

Note that the implementation change shifts the API contract such that the onus is now on `AsyncDirective` authors to check the `isConnected` flag during `update`, and this responsibility is given special treatment with an info box.

Should maybe wait to deploy this change until the next RC release?

### DIFF
--- a/packages/lit-dev-content/site/docs/templates/custom-directives.md
+++ b/packages/lit-dev-content/site/docs/templates/custom-directives.md
@@ -248,10 +248,19 @@ export const resolvePromise = directive(ResolvePromise);
 
 Here, the rendered template shows "Waiting for promise to resolve," followed by the resolved value of the promise, whenever it resolves.
 
-Async directives often need to subscribe to external resources. To prevent memory leaks async directives often need to unsubscribe or dispose of resources when the directive instance is no longer in use.  For this purpose, `AsyncDirective` provides the following extra lifecycle callbacks:
+Async directives often need to subscribe to external resources. To prevent memory leaks async directives should unsubscribe or dispose of resources when the directive instance is no longer in use.  For this purpose, `AsyncDirective` provides the following extra lifecycle callbacks and API:
 
-* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected when the value of a given expression no longer resolves to the same directive, or if the subtree the directive was contained in was removed from the DOM.
+* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected when the value of a given expression no longer resolves to the same directive, or if the subtree the directive was contained in was removed from the DOM. After a directive receives a `disconnected` callback, it should release all resources it may have subscribed to during `update` or `render` to prevent memory leaks.
+
 * `reconnected()`: Because DOM subtrees can be temporarily disconnected and then reconnected again later (for example, when DOM is moved or cached for later use) a disconnected directive may need to react to being re-connected. So the `reconnected()` callback should always be implemented alongside `disconnected()`, in order to restore a disconnected directive back to its working state.
+
+* `isConnected`: Reflects the current connection state of the directive.
+
+<div class="alert alert-info">
+
+Note that it is possible for an `AsyncDirective` to continue receiving updates while it is disconnected if its containing tree is re-rendered. As such, `update` and/or `render` should always check the `this.isConnected` flag before subscribing to any long-held resources to prevent memory leaks.
+
+</div>
 
 Below is an example of a directive that subscribes to an `Observable` and handles disconnection and reconnection appropriately:
 
@@ -265,7 +274,9 @@ class ObserveDirective extends AsyncDirective {
     if (this.observable !== observable) {
       this.unsubscribe?.();
       this.observable = observable
-      this.subscribe(observable);
+      if (this.isConnected)  {
+        this.subscribe(observable);
+      }
     }
     return noChange;
   }
--- a/packages/lit-dev-content/site/docs/templates/custom-directives.md
+++ b/packages/lit-dev-content/site/docs/templates/custom-directives.md
@@ -250,7 +250,11 @@ Here, the rendered template shows "Waiting for promise to resolve," followed by 
 
 Async directives often need to subscribe to external resources. To prevent memory leaks async directives should unsubscribe or dispose of resources when the directive instance is no longer in use.  For this purpose, `AsyncDirective` provides the following extra lifecycle callbacks and API:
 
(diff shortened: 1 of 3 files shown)


packages/lit-dev-content/site/docs/templates/custom-directives.md
```diff
@@ -248,11 +248,20 @@ export const resolvePromise = directive(ResolvePromise);
 
 Here, the rendered template shows "Waiting for promise to resolve," followed by the resolved value of the promise, whenever it resolves.
 
-Async directives often need to subscribe to external resources. To prevent memory leaks async directives often need to unsubscribe or dispose of resources when the directive instance is no longer in use.  For this purpose, `AsyncDirective` provides the following extra lifecycle callbacks:
+Async directives often need to subscribe to external resources. To prevent memory leaks async directives should unsubscribe or dispose of resources when the directive instance is no longer in use.  For this purpose, `AsyncDirective` provides the following extra lifecycle callbacks and API:
+
+* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected when the value of a given expression no longer resolves to the same directive, or if the subtree the directive was contained in was removed from the DOM. After a directive receives a `disconnected` callback, it should release all resources it may have subscribed to during `update` or `render` to prevent memory leaks.
```
justinfagnani, line 253: I'd mention host element disconnection, and since there are three items a list might be more readable. I'd order the list with the more common causes of disconnection first (subtree removal, element disconnection, resolved directive). Use consistent tense:

```suggestion
* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected in three cases:
  - When the DOM tree the directive is contained in is removed from the DOM
  - When the directive's host element is disconnected
  - When the expression that produced the directive no longer resolves to the same directive.
  After a directive receives a `disconnected` callback, it should release all resources it may have subscribed to during `update` or `render` to prevent memory leaks.
```

packages/lit-dev-content/site/docs/templates/custom-directives.md
```diff
@@ -248,11 +248,20 @@ export const resolvePromise = directive(ResolvePromise);
...
 
-* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected when the value of a given expression no longer resolves to the same directive, or if the subtree the directive was contained in was removed from the DOM.
 * `reconnected()`: Because DOM subtrees can be temporarily disconnected and then reconnected again later (for example, when DOM is moved or cached for later use) a disconnected directive may need to react to being re-connected. So the `reconnected()` callback should always be implemented alongside `disconnected()`, in order to restore a disconnected directive back to its working state.
 
+* `isConnected`: Reflects the current connection state of the directive.
+
+<div class="alert alert-info">
+
+Note that it is possible for an `AsyncDirective` to continue receiving updates while it is disconnected if its containing tree is re-rendered. As such, `update` and/or `render` should always check the `this.isConnected` flag before subscribing to any long-held resources to prevent memory leaks.
```
justinfagnani, line 261: I think you don't need the "As such"?

packages/lit-dev-content/site/docs/templates/custom-directives.md
justinfagnani: I would also update the reconnected section to mention host element re-connection.

justinfagnani (changes requested):

packages/lit-dev-content/site/docs/templates/custom-directives.md
```diff
@@ -256,13 +256,13 @@ Async directives often need to subscribe to external resources. To prevent memor
...
 
 * `isConnected`: Reflects the current connection state of the directive.
 
 <div class="alert alert-info">
 
-Note that it is possible for an `AsyncDirective` to continue receiving updates while it is disconnected if its containing tree is re-rendered. As such, `update` and/or `render` should always check the `this.isConnected` flag before subscribing to any long-held resources to prevent memory leaks.
+Note that it is possible for an `AsyncDirective` to continue receiving updates while it is disconnected if its containing tree is re-rendered. Bacause of this, `update` and/or `render` should always check the `this.isConnected` flag before subscribing to any long-held resources to prevent memory leaks.
```
justinfagnani, line 265: Typo in "Bacause", but I think you can lose this too. Just say "`update` and/or `render` should always check the `this.isConnected` flag..." Not sure @arthurevans preference on streamlining the style this way...

packages/lit-dev-content/site/docs/templates/custom-directives.md
```diff
@@ -248,10 +248,23 @@ export const resolvePromise = directive(ResolvePromise);
...
 
-* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected when the value of a given expression no longer resolves to the same directive, or if the subtree the directive was contained in was removed from the DOM.
-* `reconnected()`: Because DOM subtrees can be temporarily disconnected and then reconnected again later (for example, when DOM is moved or cached for later use) a disconnected directive may need to react to being re-connected. So the `reconnected()` callback should always be implemented alongside `disconnected()`, in order to restore a disconnected directive back to its working state.
+* `disconnected()`: Called when a directive is no longer in use.  Directive instances are disconnected in three cases:
+  - When the DOM tree the directive is contained in is removed from the DOM
+  - When the directive's host element is disconnected
+  - When the expression that produced the directive no longer resolves to the same directive.
+  After a directive receives a `disconnected` callback, it should release all resources it may have subscribed to during `update` or `render` to prevent memory leaks.
```
justinfagnani, line 257: Looks like there's a formatting issue with the last sentence. It's combined into the previous bullet.

packages/lit-dev-content/site/docs/templates/custom-directives.md
kevinpschaaf: Seemed abrupt to not connect the two sentences. When I read it without a connecting clause, it's not obvious the 2nd follows from the first.

packages/lit-dev-content/site/docs/templates/custom-directives.md
kevinpschaaf: Fixed

justinfagnani (approved):

PR #438, opened 2021-08-17
Fix check external link test timing out

AndrewJakubowicz:
Note to reviewer. There are three commits that track my attempts to solve this elegantly as I figured out the root cause. The root cause is redirect.ts that makes an internal link redirect to slack domains. This seems to be a bug specific with the "--exclude-internal" CLI argument (which I typo in the second commit). Thus the second commit identifies the cause, but provides the wrong fix.

tl;dr: The `--exclude-internal` flag changes some behavior such that the /slack-invite redirect causes the whole internet to be crawled. Without the flag we don't filter internal links out of the check, but crawl the website correctly without redirecting to slack domains.

### Context

Currently the broken link checker starts to crawl the entire internet until it eventually times out. This happens because the broken link checker queues up pages it thinks are on the `localhost` domain for external link checking. This means that it also queues up redirect urls such as `/slack-invite/` and treats them as valid urls to crawl. This results in slack being crawled by the test which eventually leads to the test timing out after a couple hours of crawling.

### Fix

I was not able to filter out the `/slack-invite/` redirect that causes this issue. The `--exclude-internal` CLI filter changes the behavior resulting in slack domain urls queueing up. In order to timebox this issue and get us to a better place I've removed the flag. This means there is some duplication of link checking as the external command now checks both internal & external links. The crawler no longer crawls onto the Slack domain fixing the timeout.

Also cleaned up two other exclusions that no longer seem to be required.
Filed issue https://github.com/lit/lit.dev/issues/439 for following up on filtering out internal links.

### Testing

This can be tested locally by running `npm run test` in the lit.dev root directory. Running on `main` will lead to the script endlessly crawling, whilst on this branch it eventually ends after traversing the entire lit.dev website.

The github action associated with this PR should also complete rather than timing out. Note I'm not addressing any broken links in this change, just fixing the test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "test:links:internal": "run-p -r start check-links:internal",
     "test:links:external": "run-p -r start check-links:external",
     "check-links:internal": "wait-on tcp:8080 && blc http://localhost:8080 --recursive --exclude-external --ordered",
-    "check-links:external": "wait-on tcp:8080 && blc http://localhost:8080 --exclude 'http://localhost:8000/dev/' --exclude 'https://github.com/Polymer/internal*' --exclude 'https://github.com/PolymerLabs/lit.dev*' --exclude 'https://www.npmjs.com/*' --get --recursive --host-requests 1"
+    "check-links:external": "wait-on tcp:8080 && blc http://localhost:8080 --exclude 'http://localhost:8000/dev/' --exclude 'https://www.npmjs.com/*' --exclude 'http://localhost:8000/slack-invite/' --get --recursive --exlude-internal --host-requests 1"
   },
   "author": "Google LLC",
   "license": "BSD-3-Clause",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "test:links:internal": "run-p -r start check-links:internal",
     "test:links:external": "run-p -r start check-links:external",
     "check-links:internal": "wait-on tcp:8080 && blc http://localhost:8080 --recursive --exclude-external --ordered",
-    "check-links:external": "wait-on tcp:8080 && blc http://localhost:8080 --exclude 'http://localhost:8000/dev/' --exclude 'https://www.npmjs.com/*' --exclude 'http://localhost:8000/slack-invite/' --get --recursive --exlude-internal --host-requests 1"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "test:links:internal": "run-p -r start check-links:internal",
     "test:links:external": "run-p -r start check-links:external",
     "check-links:internal": "wait-on tcp:8080 && blc http://localhost:8080 --recursive --exclude-external --ordered",
-    "check-links:external": "wait-on tcp:8080 && blc http://localhost:8080 --exclude 'http://localhost:8000/dev/' --exclude 'https://github.com/Polymer/internal*' --exclude 'https://github.com/PolymerLabs/lit.dev*' --exclude 'https://www.npmjs.com/*' --get --recursive --host-requests 1"
+    "check-links:external": "wait-on tcp:8080 && blc http://localhost:8080 --exclude 'http://localhost:8000/dev/' --exclude 'https://www.npmjs.com/*' --exclude 'http://localhost:8000/slack-invite/' --get --recursive --exlude-internal --host-requests 1"
```
aomarks: Sorry if I missed this from the PR description, but still not sure I understand when we need the exclusions, or why.  E.g. why is npm here but not mdn, which we also link to a lot? It seems like a reasonable default behavior for the link checker would be to follow external links, but not any transitive links? Also what does the `--host-requests` flag do? I couldn't find it mentioned at https://github.com/stevenvachon/broken-link-checker.

aomarks: Also just noticed there's a typo: `--exlude-internal`

AndrewJakubowicz: Sorry for the rambling PR description.
The typo you noticed hints to the "fix" I ended up using. It tricked me.

The --host-requests flag is documented in the `--help` text as: `Concurrent requests limit per host.`.
We need to exclude npm because it returns status code 429 (too many requests) and breaks the test.


The broken link checker has two behaviors when parsing a page. For each link we check whether or not it is broken. Then we also decide whether or not to queue up that link to check its links. This second behavior is changed by the `--exclude-internal` flag. The localhost:8080/slack-invite/ link is marked as internal so we don't check to see if it is broken, but we do queue it so that we can check if it contains external links that are broken. This results in us popping off the slack domain and checking for broken external links. Now the crawler uses the current domain (slack) to decide what is an internal link. This leads to the crawler walking slack. This eventually leads to another redirect which continues for hours.

aomarks: Ok thanks for the explanation!

aomarks (approved):

PR #442, opened 2021-08-20
Update expressions.md

lpoulter:
State that ifDefined is a function and link to it.


lpoulter:
@googlebot I signed it!

Issue #424, opened 2021-08-05
Lit.dev Integration testing


AndrewJakubowicz:
As we are building up lit.dev it would be useful to have screenshot tests that we could run over complex parts of the site.

Acceptance criteria:

 - A way to spin up a browser and run screenshot tests programmatically.
 - Easy way to create screenshot tests.
 - If not flakey, add to integration testing flow.

AndrewJakubowicz:
Made a more specific issue to track screenshot testing. https://github.com/lit/lit.dev/issues/469

PR #463, opened 2021-09-01
Trying out Playwright for lit.dev integration testing

AndrewJakubowicz:
Closes issue: https://github.com/lit/lit.dev/issues/424

### Context

This adds initial integration tests using Playwright. These run in CI with the built website and can be run locally.

Screen Shot 2021-09-01 at 2 17 07 PM

### Tests

Added four tests to check some assumptions.

 - Home page tests
       - Basic test to check some state on the page.
       - Using the site search and navigating to a search suggestion.
 - Playground tests
       - Check that default playground page loads first typescript example.
       - Test that changes the example style, and then tests the computed style in the preview is correct.

### How

For the playground example used: `npx playwright codegen lit.dev/playground` which opened chrome and a debug window that was code generating my interactions. This was really nice to quickly get the selectors.

### Punted

 - Playwright provides a pretty easy way to create and test screenshot tests. However I haven't yet figured out how to generate the screenshots in CI. Tracking: https://github.com/lit/lit.dev/issues/469


playwright.config.ts
```diff
@@ -0,0 +1,5 @@
+import {PlaywrightTestConfig} from '@playwright/test';
```
aomarks, line 1: Add license header

tests/home.spec.ts
```diff
@@ -0,0 +1,38 @@
+import {test, expect} from '@playwright/test';
```
aomarks, line 1: Add license header

tests/home.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+      'heading'
+    );
+    const homePageImg = page.locator('#splashLogo > img');
+    expect(await homePageImg.getAttribute('aria-label')).toBe('Lit');
+  });
+
+  test('homepage screenshot test', async ({page}) => {
+    await page.goto(LOCAL_DOMAIN);
+    const homePageIntro = page.locator('#intro');
+    expect(await homePageIntro.screenshot()).toMatchSnapshot('home-intro.png');
```
aomarks, line 18: There will be rendering differences between platforms right? Looks like the screenshot we have here is for mac OS. How will a developer update the linux screenshot that would be tested on CI?

aomarks (commented):
Love the idea, and @playwright/test looks super nice and simple.

I know this is still draft, just a couple of early comments.

tests/home.spec.ts
AndrewJakubowicz: I temporarily removed screenshot tests. I think there is a way to make it work.

tests/playground.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    // Double click text=blue
+    await page.dblclick('text=blue');
+
+    // Change the text to red
+    await page.keyboard.type('red');
+
+    // Wait for iframe to reload
+    await page.waitForSelector(
+      'playground-preview mwc-linear-progress[aria-hidden="true"]'
```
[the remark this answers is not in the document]
AndrewJakubowicz: Done! I also tried to find a way to listen to the event but I couldn't find a path forward with a short timeboxed investigation.

Instead updated to use your `part` suggestion and it works fantastic!

Thank you!

aomarks (approved):

aomarks (approved):

PR #465, opened 2021-09-01
Redirects for dev mode short links

aomarks:
Adds a `https://lit.dev/msg/<code>` redirect for all warnings logged by Lit's dev mode.

Half of https://github.com/lit/lit/issues/2078. Other half is at https://github.com/lit/lit/pull/2119.

### DIFF
--- a/packages/lit-dev-server/src/index.ts
+++ b/packages/lit-dev-server/src/index.ts
@@ -43,28 +43,21 @@ if (mode === 'playground') {
 }
 
 app.use(async (ctx, next) => {
-  if (ctx.path.match(/\/[^\/\.]+$/)) {
+  let path = ctx.path;
+  if (path.match(/\/[^\/\.]+$/)) {
     // Canonicalize paths to have a trailing slash, except for files with
     // extensions.
-    ctx.status = 301;
-    ctx.redirect(
-      ctx.path + '/' + (ctx.querystring ? '?' + ctx.querystring : '')
-    );
-  } else if (ctx.path.endsWith('//')) {
+    path += '/';
+  }
+  if (path.endsWith('//')) {
     // Koa static doesn't care if there are any number of trailing slashes.
     // Normalize this too.
+    path = path.replace(/\/+$/, '/');
+  }
+  path = pageRedirects.get(path) ?? path;
+  if (path !== ctx.path) {
     ctx.status = 301;
-    ctx.redirect(
-      ctx.path.replace(/\/+$/, '/') +
-        (ctx.querystring ? '?' + ctx.querystring : '')
-    );
-  } else if (pageRedirects.has(ctx.path)) {
-    // Handle any 1:1 page redirects
-    ctx.status = 301;
-    ctx.redirect(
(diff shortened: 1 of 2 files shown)


packages/lit-dev-server/src/index.ts
```diff
@@ -43,28 +43,21 @@ if (mode === 'playground') {
...
-  } else if (ctx.path.endsWith('//')) {
+    path += '/';
+  }
+  if (path.endsWith('//')) {
     // Koa static doesn't care if there are any number of trailing slashes.
     // Normalize this too.
+    path = path.replace(/\/+$/, '/');
+  }
+  path = pageRedirects.get(path) ?? path;
+  if (path !== ctx.path) {
```
AndrewJakubowicz, line 58: Elegant!

Especially like how this check handles the three different `301` cases.

AndrewJakubowicz (approved):
This looks really great and a huge ergonomic improvement using Lit. I tried out manually some short urls:

These work great:
 - https://pr465-57420cb---lit-dev-5ftespv5na-uc.a.run.app/msg/removed-api
 - https://pr465-57420cb---lit-dev-5ftespv5na-uc.a.run.app/msg/removed-api/

Query params are not handled but can probably punt since we have control of the short urls:

https://pr465-57420cb---lit-dev-5ftespv5na-uc.a.run.app/msg/removed-api?mods=doesquerywork
 

Is there a way to test the redirect so we don't accidentally break a link. For example updating a content heading could change the heading id and break the fragment redirect.

packages/lit-dev-server/src/index.ts
justinfagnani: Nice, but... can we add a comment? :)

arthurevans:
Looks good to me. I hadn't though of the redirect testing that @AndrewJakubowicz mentioned, but I think we'll (at a minimum) need to add comments to each of the docs sections where these links land, so we don't accidentally break the intended association between the current URL and the answer we're looking for (for example, when refactoring one section into two). I opened #467 for that, because it doesn't seem like it needs to be part of this PR.

justinfagnani (approved):

packages/lit-dev-server/src/index.ts
aomarks: Added a comment about how it's to flatten redirect chains so we don't actually serve a chain of them.

aomarks:
> Good idea. It would be nice if we could use the link checker we already have for this, but it doesn't support checking for anchors (see [stevenvachon/broken-link-checker#108](https://github.com/stevenvachon/broken-link-checker/issues/108) -- understandable since it would require DOM parsing). I've written a little custom script to do this, will send it in a followup PR.

Done in https://github.com/lit/lit.dev/pull/468

arthurevans:
Link checker is :fire:. I'd still like to add comments to at least the sections in the main docs, because those sections serve multiple purposes. We could (and probably should) add a more specific section on class field shadowing, say, or change-during-update, but in doing so we probably wouldn't remove the existing anchor, so the link wouldn't break, even though the relevant content might have moved to another section.

Issue #467, opened 2021-09-01
Add comments to docs that cover dev mode warnings


arthurevans:
The shortlinks in https://github.com/lit/lit.dev/pull/465 redirect to sections that explain specific issues that devmode warns about. There should be a comment in each of these locations so the hapless editor (ahem) knows to consider this factor rearranging docs.

We also probably need a policy about how to handle these over time. i.e., if Lit 3 eliminates a problem or adds a different solution for it, we still need to keep the Lit 2 shortlink (and associated doc) alive.

arthurevans:
It may be overkill, but I want to add comments where we link to the main docs. Reopening to track those.

PR #468, opened 2021-09-01
Add script for checking lit.dev redirects

aomarks:
Custom script for checking lit.dev redirects.

image

It would be nice if we could use the 3rd party link checker we already have for this somehow, but it doesn't support checking for anchors (see stevenvachon/broken-link-checker#108 -- understandable since it would require DOM parsing) which is one of the main failure cases.

Fixes https://github.com/lit/lit.dev/issues/467 (since we shouldn't need comments if we have the redirects checked in CI).

As part of this, I created a new `lit-dev-tools-esm` package.

The existing `lit-dev-tools` package is currently CommonJS, because mostly it is used for Eleventy plugins, and Eleventy doesn't support ES modules (11ty/eleventy#836).

We want ES modules for this new redirect checker script, because it needs to import some ES modules, and that is difficult to do with TypeScript, because TypeScript doesn't allow emitting an actual `import` statement, which is how CommonJS -> ESM interop works (microsoft/TypeScript#43329).

We also can't really have a mix of CommonJS and ESM in the same package, because the `{"type": "module"}` field has to be set to one or the other in the `package.json`. We could use `.mjs` extensions, but TypeScript won't emit those.

So the simplest solution seems to be to just have two packages!


aomarks:
Actually I'm going to create an ESM package for this, will reopen when ready.

aomarks:
CI should be working now.

PR #472, opened 2021-09-02
Update caching to use setup-node v2

AndrewJakubowicz:
### Context

This updates our github actions to cache using setup-node@v2.

This change preserves caching and simplifies the github actions file. It follows: https://github.com/lit/lit/pull/2130


aomarks (approved):
Yay for simplicity

PR #473, opened 2021-09-02
Add new warnings from code

kevinpschaaf:


packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -333,15 +333,45 @@ Element expressions can occur inside the opening tag after the tag name:
...
+  <!-- ERROR -->
+  <template>${content}</template>
+
+  <!-- OK -->
+  <template id="${attrValue}">static content ok</template>
+  ```
+
+* Inside `<textarea>` element content (attribute expressions on the textarea element itself are allowed). Note that Lit can render content into textarea, however editing the textarea will break references to the DOM that Lit uses to dynamically update. Instead, bind to the `.value` property of textarea instead.
+  ```html
+  <!-- ERROR -->
```
justinfagnani, line 362: Instead of `ERROR` I'd say `BEWARE`

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -333,15 +333,45 @@ Element expressions can occur inside the opening tag after the tag name:
...
-```
+* Where tag or attribute names would appear. Lit does not support dynamically changing values in this position.
+
+  ```html
+  <!-- ERROR -->
+  <${tagName}></${tagName}>
+
+  <!-- ERROR -->
+  <div ${attrName}=true></div>
+  ```
```
justinfagnani, line 348: Link to static docs

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -333,15 +333,45 @@ Element expressions can occur inside the opening tag after the tag name:
 <div ${ref(elementReference)}></div>
 ```
 
-Expressions **_cannot_** appear where tag or attribute names would appear; however [static expressions](#static-expressions) can.
+### Invalid locations { #invalid-locations }
 
-```html
-<!-- ERROR -->
-<${tagName}></${tagName}>
+Expressions should not appear in the following locations and will issue warnings/errors in development mode:
```
justinfagnani, line 338: Also mention comments

justinfagnani (changes requested):

sorvell (approved):

justinfagnani (approved):

packages/lit-dev-server/src/redirects.ts
```diff
@@ -5,25 +5,47 @@
  */
 
 // prettier-ignore
```
aomarks, line 7: Did you intentionally unformat this? I felt it was much more reasonable to do column alignment, hence the `// prettier-ignore` comment.

kevinpschaaf: Not intentionally; apparently the prettier-ignore didn't work 🤷

packages/lit-dev-content/site/docs/templates/expressions.md
kevinpschaaf: Done

packages/lit-dev-content/site/docs/templates/expressions.md
kevinpschaaf: The static reference is at the end of the list.

packages/lit-dev-content/site/docs/templates/expressions.md
kevinpschaaf: Done

packages/lit-dev-server/src/redirects.ts
kevinpschaaf: Will try and fix.

kevinpschaaf:
Merging with failed link: https://beautifier.io/ is currently down, it appears. Assuming it's temporary.

PR #474, opened 2021-09-03
Bump dependencies and API docs

aomarks:
- Bumps all dependencies to their latest major versions.

- Separates `nuke` and `upgrade` commands, mirroring what we do in the Lit/lit repo. `nuke` forces a full re-install using the current package locks, whereas `upgrade` also deletes the package locks.

- Fix new errors from TypeScript 4.4. I had to turn `skipLibCheck` on for now, because of https://github.com/material-components/material-web/issues/2676. Can turn it back on once that is fixed.

- Removed the `serve` dependency that we don't actually use.

- Upgrade to the latest `markdown-it-anchor` package, which now creates accessible permalinks out of the box, so we can delete the custom script we had before. We do still need to do a little manipulation of the DOM it generates, but it's much less now.

- Bump the API docs SHA.

- Remove "reference" nodes from API docs, since they contain no useful information.


AndrewJakubowicz:
A bunch of the external API links to declaration files seem broken (in the external link checker).

For example: https://pr474-82638a4---lit-dev-5ftespv5na-uc.a.run.app/docs/api/directives/#AsyncAppendDirective.constructor (view source button for the AsyncAppendDirective)

This declaration link isn't there on prod: https://lit.dev/docs/api/directives/#AsyncAppendDirective.constructor

AndrewJakubowicz (approved):
LGTM! This is fantastic!

Minor nitpick that I haven't been able to track down. There are new "view source" links in some of the api pages that return 404 status codes.

Could it be that the api documentation needs the sha bumped?

aomarks:
> LGTM! This is fantastic!
> 
> Minor nitpick that I haven't been able to track down. There are new "view source" links in some of the api pages that return 404 status codes.
> 
> Could it be that the api documentation needs the sha bumped?

I bumped the SHA. This did seem to fix a few of the links, but not all. Worth doing anyway.

It seems like TypeDoc occasionally has trouble resolving the source file of certain things to the `.ts` file, and links to the `d.ts` file instead. The case I saw was a constructor that was inherited from a superclass. Since it only affects a few things, I've just added a check so make sure we don't try to link to GitHub when we don't have a `.ts` source.

I also noticed there were a bunch of "Reference" nodes in the API docs (even on the current site) at https://lit.dev/docs/api/custom-directives/#directive. They don't have any useful information, and mostly seem to refer to things that already have proper documentation (e.g. there's "directive" the documented function, and then "directive" the reference with no information). We'll now filter these out.

packages/lit-dev-tools/src/generate-api-docs-data.ts
```diff
@@ -833,6 +835,18 @@ class Transformer {
...
+    if (!source.fileName.endsWith('.ts')) {
+      throw new Error(
+        `Unexpected source.fileName extension: ${source.fileName}`
+      );
+    }
+    if (source.fileName.endsWith('.d.ts')) {
+      // TODO(aomarks) For an unknown reason, TypeDoc sometimes resolves to d.ts
+      // files instead of original .ts source files, e.g. when a class inherits
+      // a constructor from a superclass. We can't link to these, since d.ts
+      // files aren't checked into GitHub.
```
AndrewJakubowicz, line 847: This is awesome! Great find and comment!

Issue #332, opened 2021-04-21
JavaScript docs with TS/JS switch


aomarks:
Similar to what we have on the homepage, add a global language switch that applies to all the docs examples, and store the preference.

aomarks:
We've got a preview ready of the new switch that lets you toggle between TypeScript and vanilla JavaScript code samples on lit.dev:

https://lit.dev/docs/?mods=jsSamples

This covers all the code: inline snippets, interactive samples, full examples on the playground, and the tutorial.

Your preference is saved in localStorage, so it will remember across pages, and the next time you visit the site.

There's a little more polish to do here before we make the toggle visible by default, but feel free to check it out now using that special URL parameter, and file any issues you notice.

(If you're curious, most of this code is automatically generated using the new @lit/ts-transformers package: https://github.com/lit/lit/tree/main/packages/ts-transformers. These transformers automatically convert TypeScript code with decorators into the idiomatic vanilla JavaScript code we currently recommend.)

aomarks:
JS docs are now available on lit.dev.

Look for this button to toggle between JavaScript and TypeScript anywhere on the site, and your preference will be saved across sessions with localStorage:

![image](https://user-images.githubusercontent.com/48894/133345460-9d7d855f-2321-412e-bfdc-be2f168dde12.png)

PR #477, opened 2021-09-07
Generate JS versions of all samples

aomarks:
Adds a new script that generates JS versions of every lit.dev sample.

**Note:** this does not yet integrate into the UI. This will come in a followup PR. It *is* possible to view the output directly now, though, by adding the `js/` prefix to any sample in a playground URL, e.g. https://pr477-979ab34---lit-dev-5ftespv5na-uc.a.run.app/playground/#sample=js/examples/full-component

- Outputs to `samples/js`
- Compiles TypeScript with [`@lit/ts-transformers`](https://github.com/lit/lit/tree/main/packages/ts-transformers) so that decorators are converted to idiomatic JavaScript style
- Updates `.ts` -> `.js` file references in `project.json` files
- Symlinks all other kinds of files (html, css, etc.)
- Supports one-time and `--watch` mode for fast dev

Part of https://github.com/lit/lit.dev/issues/332


justinfagnani (approved):

packages/lit-dev-content/samples/tsconfig.json
```diff
@@ -10,16 +10,9 @@
...
-  "include": [
-    "./**/*.ts",
-  ],
-  "exclude": [
-    "node_modules/**/*",
-    // TODO(kschaaf): Remove once lit released with these directives
-    "examples/directive-async-append/**/*",
-    "examples/directive-async-replace/**/*",
-    // (arthure) This tutorial step is intentionally incomplete
-    "tutorial/06-template-logic/before/**/*",
```
AndrewJakubowicz, line 23: Clarifying my understanding. This was being skipped since originally there was a type error (i.e. missing `addToDo` method in this example.
You've fixed the error by adding a stub so can remove from the exclude. And we want this to be "type valid" code such that we can transform it into javascript?

aomarks: Exactly

aomarks: Clarifying: we could actually still emit the code even with a type checking error, but it could be a bit complicated to distinguish between intentional vs unintentional errors. We want the build to fail if you introduce an unintentional type error. Given that this was the *only* case like that in the entire site, it seemed reasonable to just make this one spot type-correct, since it doesn't affect that tutorial step too much.

AndrewJakubowicz (approved):
LGTM!
I manually tested a bunch of examples using the url. Also tested the mouse-controller docs example.

Small note on the ["Full component" example](https://pr477-979ab34---lit-dev-5ftespv5na-uc.a.run.app/playground/#sample=js/examples/full-component). We may need to handle documentation that calls language specific features. In this example we mention the `@property` decorator which is possibly confusing in the JS example. Non blocking!

This is fantastic! 🎉 :shipit:

packages/lit-dev-content/samples/tsconfig.json
AndrewJakubowicz: Thank you for the answer! I really love that the build fails in the error case.

aomarks: > Out of scope questions: Is there a way to manually test the JS variant of the tutorial? Does the solve button work the same?

In the next PR, I'll integrate a JS/TS switch into all the playgrounds, including the tutorial. I'll put the switch behind a `?mods` parameter. Then we'll be able to test the tutorial along with all the inline docs samples.

> Do the code samples in the inline tutorial exposition get transformed to JS (Asking since I'm not sure if the code snippets in the tutorial exposition are references to live samples or duplicated in the content)?

Inline code snippets won't get automatically transformed, since they are usually not complete programs (e.g. just the content of a class, which won't parse). Instead I'm going to add a mechanism for defining both a TS and JS version of any snippet inside the markdown, and we'll just show/hide one or the other depending on the setting.

packages/lit-dev-server/package.json
```diff
@@ -17,8 +17,7 @@
     "koa": "^2.13.0",
     "koa-conditional-get": "^3.0.0",
     "koa-etag": "^4.0.0",
-    "koa-static": "^5.0.0",
-    "lit-dev-content": "^0.0.0"
```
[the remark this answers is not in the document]
aomarks: This dependency is only needed at run time, not at build time, so it does not need to be declared in the dependency graph.

Removing this dependency is helpful because it breaks the cycle:

  server -> content -> tools-esm -> server

Cycles in Lerna are bad because they create undefined build ordering.

In this case, we want to add the content -> tools-esm dependency so that we can run esm tools from the content package's build step. Breaking the cycle by removing server -> content lets us do this.

aomarks:
> Small note on the ["Full component" example](https://pr477-979ab34---lit-dev-5ftespv5na-uc.a.run.app/playground/#sample=js/examples/full-component). We may need to handle documentation that calls language specific features. In this example we mention the `@property` decorator which is possibly confusing in the JS example. Non blocking!

Very true! We'll need to do a sweep through all the samples to find cases like this. In some cases we can probably rephrase to make it TS-neutral, but we might end up needing to have a way to override certain transforms for special cases by checking in a file that takes precedence.

PR #478, opened 2021-09-07
Adds JS/TS toggle to interactive docs samples

aomarks:
Adds a JS/TS toggle to all interactive samples in the docs, behind the `?mods=jsSamples` feature flag.

**Note:** The playground page, tutorial page, and static samples will come in a followup PR.

**Also note:** There are a some issues with the JS samples that I've seen already, mostly around `/*playground-fold*/` and `/*playground-hide*/` comments not getting preserved quite right. I'll address those in a followup too.

Also fixes layout shift issues with inline playgrounds.

### DIFF
--- a/packages/lit-dev-content/site/css/mods.css
+++ b/packages/lit-dev-content/site/css/mods.css
@@ -22,3 +22,9 @@
   --playground-code-attribute-color: #9cdcfe;
   --playground-code-callee-color: #dcdcaa;
 }
+
+.jsSamples {
+  /* TODO(aomarks) Remove when JS samples launch, and update the
+  litdev-typescript-switch display property to match this. */
+  --litdev-typescript-switch-display: inline-flex;
+}
--- a/packages/lit-dev-content/src/components/litdev-example-controls.ts
+++ b/packages/lit-dev-content/src/components/litdev-example-controls.ts
@@ -1,0 +1,66 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LitElement, html, css, customElement, property} from 'lit-element';
+import './litdev-typescript-switch.js';
+
+/**
+ * Controls for lit.dev code examples.
+ */
+@customElement('litdev-example-controls')
+export class LitDevExampleControls extends LitElement {
+  static override styles = css`
+    :host {
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+      justify-content: center;
+      margin-left: auto;
+    }
+
+    #openInPlayground {
+      display: flex;
+      color: inherit;
+      opacity: 70%;
+      margin-left: 10px;
+    }
+
+    #openInPlayground:hover {
+      opacity: 100%;
+    }
+  `;
+
+  /**
+   * Path to the project dir from `samples/PROJECT/project.json`.
+   */
+  @property()
+  project?: string;
+
+  override render() {
+    return html`
+      <litdev-typescript-switch></litdev-typescript-switch>
+      <a
+        id="openInPlayground"
+        title="Open this example in the playground"
+        target="_blank"
+        href="/playground/#sample=${this.project}"
+      >
+        <!-- Source: https://material.io/resources/icons/?icon=launch&style=baseline -->
+        <svg width="22px" height="22px" viewBox="0 0 24 24" fill="#5f5f5f">
+          <path
+            d="M19 19H5V5h7V3H5c-1.11 0-2 .9-2 2v14c0 1.1.89 2 2 2h14c1.1 0 2-.9 2-2v-7h-2v7zM14 3v2h3.59l-9.83 9.83 1.41 1.41L19 6.41V10h2V3h-7z"
+          />
+        </svg>
+      </a>
+    `;
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'litdev-typescript-controls': LitDevExampleControls;
+  }
+}
--- a/packages/lit-dev-content/src/components/litdev-typescript-switch.ts
+++ b/packages/lit-dev-content/src/components/litdev-typescript-switch.ts
@@ -1,0 +1,124 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LitElement, html, css, customElement} from 'lit-element';
+import {
+  getTypeScriptPreference,
+  setTypeScriptPreference,
+  TYPESCRIPT_PREFERENCE_EVENT_NAME,
+} from '../typescript-preference.js';
+
+/**
+ * A switch that toggles between TypeScript and JavaScript preferences.
+ *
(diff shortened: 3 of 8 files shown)


AndrewJakubowicz (commented):
Before looking at code, some quick nits from manual testing.

See screenshot comparing https://lit.dev/docs/components/properties/ and https://pr478-ff9f2da---lit-dev-5ftespv5na-uc.a.run.app/docs/components/properties/

Screenshot is of the last interactive example on the page using Chrome.

Screen Shot 2021-09-07 at 4 26 48 PM

 - `<div id="bar">` is possibly missing border-radius. Visual artifact is the rounded corner has a tiny break in it.
 - Last example has a scroll bar. This is fine, but maybe could add some padding beneath the "open example in playground" and the top of the scroll bar.

packages/lit-dev-content/site/css/mods.css
```diff
@@ -22,3 +22,9 @@
   --playground-code-attribute-color: #9cdcfe;
   --playground-code-callee-color: #dcdcaa;
 }
+
+.jsSamples {
+  /* TODO(aomarks) Remove when JS samples launch, and update the
+  litdev-typescript-switch display property to match this. */
+  --litdev-typescript-switch-display: inline-flex;
```
AndrewJakubowicz, line 29: Genius. The mod system is brilliant.

packages/lit-dev-content/src/components/litdev-example-controls.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+import {LitElement, html, css, customElement, property} from 'lit-element';
+import './litdev-typescript-switch.js';
+
+/**
+ * Controls for lit.dev code examples.
+ */
+@customElement('litdev-example-controls')
+export class LitDevExampleControls extends LitElement {
```
AndrewJakubowicz, line 14: Minor style nitpick. The switch and openInPlayground link have different heights, 25px and 22px respectively.

This is non blocking since the switch isn't even enabled. Instead commenting that it currently feels like the switch is sitting lower in the menu bar than the open in playground image. The two menu items don't feel vertically aligned.

aomarks:
> * `<div id="bar">` is possibly missing border-radius. Visual artifact is the rounded corner has a tiny break in it.

Good catch, that was it! Fixed.

> * Last example has a scroll bar. This is fine, but maybe could add some padding beneath the "open example in playground" and the top of the scroll bar.

Good catch. I think actually a lot of the samples are a weird size right now, even for TypeScript. I'll do a sweep through all of them as a followup to improve the sizes. (I've been planning to automatically compute the sizes during build, but still haven't gotten around to it, so right now they are all defined manually and they get out of sync easily as the code changes.)

> * The `JS` side of the toggle is `~40px` while the `TS` side is `~45px`. This is a very micro nit, but it feels like it should be equal size.

Oh right, it was using the intrinsic size of the content, so it was very slightly different. Fixed so they are equal size.

packages/lit-dev-content/src/components/litdev-typescript-switch.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    button:not(:last-of-type) {
+      border-right: none;
+    }
+  `;
+
+  override connectedCallback() {
+    // TODO(aomarks) After we upgrade to Lit 2, this and similar code in other
+    // components could be refactored into a controller.
```
AndrewJakubowicz, line 71: Exceptional comment. Was wondering this.

AndrewJakubowicz (approved):
Brilliant! LGTM!

packages/lit-dev-content/src/components/litdev-example-controls.ts
aomarks: Here are some options:

Before:


Top-aligned:


Enlarged:


Is "top-aligned" what you were thinking? Or more like "enlarged"? The problem with "enlarged" to my eye is that the stroke width of the icon becomes too heavy, and that clashes with the stroke width of the JS/TS switch. We could edit the SVG to drop the stroke width, though. WDYT?

AndrewJakubowicz: Let's keep it as before. I think my eye is getting impacted by the arrow at the top of the icon making it feel like it's higher up.

I was thinking more along the lines of enlarged but 100% agree about the stroke width making it awkward.

packages/lit-dev-content/src/components/litdev-example-controls.ts
```diff
@@ -49,7 +46,7 @@ export class LitDevExampleControls extends LitElement {
         href="/playground/#sample=${this.project}"
       >
         <!-- Source: https://material.io/resources/icons/?icon=launch&style=baseline -->
-        <svg width="22px" height="22px" viewBox="0 0 24 24" fill="#5f5f5f">
+        <svg width="22px" height="22px" viewBox="0 2 24 24" fill="#5f5f5f">
```
AndrewJakubowicz, line 49: I really like this 2px bump down.

packages/lit-dev-content/src/components/litdev-example-controls.ts
```diff
@@ -15,9 +15,6 @@ export class LitDevExampleControls extends LitElement {
   static override styles = css`
     :host {
       display: flex;
-      flex-direction: row;
-      align-items: center;
-      justify-content: center;
```
AndrewJakubowicz: I personally prefer this with the centering especially when there are file tabs.

Screen Shot 2021-09-07 at 5 22 40 PM

justinfagnani:
Is there a "switch" component that could be used? I think that would be more intuitive/obvious.

This switch doesn't look that switchable at first glance to me, in that the options are mutually exclusive. Maybe some styling tweaks could help - like making the non-selected option more grayed out, and/or share a background color with its container?

packages/lit-dev-content/src/components/litdev-example-controls.ts
justinfagnani: I like "before" of these... my understanding of the material sizing system (likely imperfect) is that there's an outer size, like 24px, but the icon doesn't fully fill it, which some components, like this switch might. So the container's hsould have the same size, but the icon-based widgets will appear a bit smaller.

aomarks: Done!

packages/lit-dev-content/src/components/litdev-example-controls.ts
aomarks: Oh yeah, definitely looks wrong here :D

aomarks:
> Is there a "switch" component that could be used? I think that would be more intuitive/obvious.
> 
> This switch doesn't look that switchable at first glance to me, in that the options are mutually exclusive. Maybe some styling tweaks could help - like making the non-selected option more grayed out, and/or share a background color with its container?

Yeah, I could see that. The current design is what I think a Material-ish to this problem would look like. A totally different option (no reason we have to stick with Material) is something like this:



I'm going to punt on this for now, since the switch feature is behind a feature flag, but I'll make sure to revisit and potentially implement something like above before the public launch.

PR #479, opened 2021-09-08
Documented useDefineForClassFields

KeithHenry:
`useDefineForClassFields` is incompatible with `experimentalDecorators`, see https://github.com/lit/lit/issues/1985

There are two solutions, both should be covered by Lit's documentation:

- Use `useDefineForClassFields: false` to force this new feature off.
- Use `declare` to tell TS not to emit inline class fields that block the decorator define function.

### DIFF
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -79,6 +79,17 @@ To use decorators with [TypeScript](https://www.typescriptlang.org/docs/handbook
 
 Enabling `emitDecoratorMetadata` is not required and not recommended.
 
+For TypeScript 3.7 and above a new flag `useDefineForClassFields` is incompatible with property decorators, and for 4.3 when the `target` is `es2020` or higher it defaults to true. 
+
+There are two ways to work around this incompatability: either add a compiler option to force it off:
+
+```json
+"experimentalDecorators": true,
+"useDefineForClassFields": false,
+```
+
+Alternatively `useDefineForClassFields` _can_ be used if every decorator annotated property uses `declare`.
+
 ### Using decorators with Babel  { #decorators-babel }
 
 If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins:


aomarks:
Thank you for the PR! We've actually just landed a bunch of improvements to the docs around this issue in https://github.com/lit/lit.dev/pull/516 which you can see live at https://lit.dev/docs/components/properties/#avoiding-issues-with-class-fields.

KeithHenry:
@aomarks 

> Thank you for the PR! We've actually just landed a bunch of improvements to the docs around this issue in https://github.com/lit/lit.dev/pull/516 which you can see live at https://lit.dev/docs/components/properties/#avoiding-issues-with-class-fields.

Thanks and that's a lot more in depth, but it doesn't mention the `declare` option at all, and I think that might be a better option in new code than the compiler flag, as it will remain compatible with future TS changes (which seems unlikely for the flag).

arthurevans:
Hi @KeithHenry. The engineer who wrote that section left the `declare` out intentionally. My understanding is that adding `declare` prevents you from initializing the property. For example:

```
@property()
declare name = 'No one';
```

Gives an error. Our preferred style is to initialize the field where we first declare it, and adding the `declare` keyword would break our sample code that uses this style. Hope that explains the thought process.

KeithHenry:
@arthurevans @aomarks I'm aware of that, it should be in the documentation too. Yes, `declare` prevents inline initialisation, but `useDefineForClassFields: false` affects that too, just in a different way.

So `useDefineForClassFields: false` lets you do this:

```ts
@property() name = 'No one';
```

However, this must not be initialised as a class field - that's the bit that breaks prototype setter decorators like `@property`, so the actual JS output (with `useDefineForClassFields: false` set) is something like:

```js
constructor() { 
  this.name = 'No one'; 
}
...
__decorate([property()], ExampleElement.prototype, 'name');
```

This makes the `constructor` vs `prototype` initialisation done by TS something you're _relying_ on. It's not broken (so not bad practice as such) and definitely the easiest course updating existing Lit projects to TS 4.2, but it's also likely to change again in future and potentially have issues with TS's field outputs (they have good reasons for their switch to class fields by default).

Meanwhile `declare` syntax stops you from initialising fields, but then also doesn't have _side effects_ that are in the JS output but not your TS working code.

Or to put it another way - if you want to initialise a decorated property with `declare` you need to set it in the `constructor`, but _this is what TS outputs anyway_.

I think it _might_ be _better_ practice to use `declare` rather than `useDefineForClassFields: false` because you get fewer potential issues with JS output, but honestly I'm not certain.

Either way `declare` should be in the documentation, with the explicit caveat that inline initialisation won't work.

Meanwhile `useDefineForClassFields: false` should have the caveat that inline initialisation will work but won't actually be inline anymore in the output JS.

aomarks:
```ts
// TS visible code
@property() declare name: string;
```

I would be worried about what this syntax will mean when decorators are finally standardized. This seems like it should be a TypeScript error, because the `declare` syntax means there would be no class field information in JS for the `@property` decorator to receive. Thoughts @justinfagnani ?

PR #480, opened 2021-09-08
Add JS/TS toggle to tutorial

aomarks:
Adds the JS/TS toggle to the tutorial page (still behind the `?mods=jsSamples` feature flag).

Doesn't yet affect the LHS content, will do that in a followup.

Part of #332

image


aomarks:
Looks like external links error is due to beautifier.io being down.

```
├─BROKEN─ https://beautifier.io/ (ERRNO_ECONNRESET)
```

packages/lit-dev-content/src/global/apply-mods.ts
```diff
@@ -0,0 +1,23 @@
...
+
+if (MODS) {
+  document.body.classList.add(
+    ...MODS.split(' ')
+      .map((v) => v.trim())
+      .filter((v) => v !== '')
+  );
+  window.addEventListener('click', (event) => {
+    for (const el of event.composedPath()) {
+      if (el instanceof HTMLAnchorElement) {
```
AndrewJakubowicz, line 17: wow! Nice

AndrewJakubowicz (approved):
Amazing. Some tiny notes, none blocking - all only happen behind mod flag:

 - On [step 3 of the tutorial](https://pr480-548b423---lit-dev-5ftespv5na-uc.a.run.app/tutorial/?mods=jsSamples#03-properties) the html template gets one space on either side. Just something I noticed. Is the TS template trimmed?
 - I added some code then switched the language which resulted in me losing some progress. I think this may need to be communicated or it could be frustrating. While playing with this I found that you can _undo_ through the language change. So you can get your progress back, but now you have JS content in a TS file. Repro:
     - Add some changes.
     - Switch the language and notice your changes are gone in the new language.
     - Press `ctrl` or `cmd` + `Z`.
     - Your changes are back but the file and toggle didn't undo.
     - The result of this can be TS content when toggle is set to JS. This seems to break the preview.
 - On a slower network it's still possible to edit the code before it vanishes.


I also tested out the search. The mod URL param refactor is really nice.

aomarks:
> * On [step 3 of the tutorial](https://pr480-548b423---lit-dev-5ftespv5na-uc.a.run.app/tutorial/?mods=jsSamples#03-properties) the html template gets one space on either side. Just something I noticed. Is the TS template trimmed?

The JS templates are formatted by Prettier, which reformats some HTML. Maybe we could turn that off, or we could format all TS with Prettier so that it's consistent. Will address in followup.

> * I added some code then switched the language which resulted in me losing some progress. I think this may need to be communicated or it could be frustrating. While playing with this I found that you can _undo_ through the language change. So you can get your progress back, but now you have JS content in a TS file. Repro:
>   
>   * Add some changes.
>   * Switch the language and notice your changes are gone in the new language.
>   * Press `ctrl` or `cmd` + `Z`.
>   * Your changes are back but the file and toggle didn't undo.
>   * The result of this can be TS content when toggle is set to JS. This seems to break the preview.
> * On a slower network it's still possible to edit the code before it vanishes.

Good catch, that is something we should improve. Will address in followup. Any thoughts on how it should work -- maybe a popup that says "Toggling languages will lose your changes [CONTINUE] [CANCEL]"?

AndrewJakubowicz:
That popup seems like a great stopgap solution. Definitely useful if the workspace has been modified and things will be lost.

aomarks:
> That popup seems like a great stopgap solution. Definitely useful if the workspace has been modified and things will be lost.

Filed https://github.com/lit/lit.dev/issues/485

PR #481, opened 2021-09-08
Add JS/TS toggle to playground

aomarks:
Adds the JS/TS toggle to the playground page page (still behind the `?mods=jsSamples` feature flag).

One special thing about this integration is that we remove the toggle when the project is shared, or loaded from the base64 URL, since we only have JS versions available for our pre-defined samples.

Part of #332

image


AndrewJakubowicz (approved):
Changes look amazing. Also manually tested out the sharing.

LGTM!

PR #482, opened 2021-09-08
Add switchable static samples shortcode

aomarks:
Adds a `{% switchable-sample %}` shortcode which can be used in markdown files to create a static code snippet that the user can switch between TypeScript and JavaScript. (Still behind the `?mods=jsSamples` feature flag).

It's used in a markdown file like this. (Note these can't be automatically generated, since they are typically incomplete programs):

````

{% switchable-sample %}

```ts
  @property()
  name: string;
}
```

```js
  static properties = {
    name: {},
  };
}
```

{% endswitchable-sample %}

````

And produces this:

Part of #332

### CODE REVIEW

**Comment by aomarks** [comment]
Pushed an update that removes the anti-layout measures (so we allow the snippets to take up different amounts of space and reflow the document, simplifying things greatly), and that floats the toggle switch on the first line of code, instead of having space reserved above.

File: packages/lit-dev-content/site/docs/components/properties.md [hunk 1]
```diff
@@ -30,11 +42,26 @@ The component shouldn't change its own public properties, except in response to
...
 
+{% switchable-sample %}
+
 ```ts
 @state()
 private _counter = 0;
 ```
 
+```js
+static properties = {
```
AndrewJakubowicz, line 53: Are `@state` decorators handled with `static properties` as well?

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -405,6 +516,24 @@ set prop(val: number) {
 get prop() { return this._prop; }
 ```
 
+```js
+static properties = {
+  prop: {},
+};
+
+_prop = 0;
```
AndrewJakubowicz, line 524: This example is the first time in the examples where the member initialization isn't in the constructor.  Am I correct in my understanding that it's because this `_prop` isn't in the `static properties` object?

AndrewJakubowicz (approved):
LGTM! Small unrelated thought, we can remove the two separate `Hello World (TypeScript)` and `Hello World (JavaScript)` playground examples when we release the toggle.

This is awesome.

packages/lit-dev-content/rollup.config.js
```diff
@@ -88,11 +89,18 @@ export default [
     ],
   },
 
-  // Build the mods script separately from the rest because it includes a local
-  // import that we actually DO want to duplicate. This script is inlined and
-  // must run before first render, so an asynchronous import is not acceptable.
+  // These scripts are inlined and must run before first render because they set
+  // global CSS classes/atributes that would otherwise cause restyle/relayout.
```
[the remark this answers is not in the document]
aomarks: Fixed

packages/lit-dev-content/site/docs/components/properties.md
aomarks: Yep, but great question because I forgot the `{state: true}` setting here! Thanks, fixed!

packages/lit-dev-content/site/docs/components/properties.md
aomarks: Yep, only reactive properties need to be initialized in the constructor.

Reactive properties need to be initialized in the constructor instead of declared as a class field because otherwise the property will be directly defined on the instance, shadowing the automatically-generated setter on the prototype, thus breaking reactivity.

I could certainly imagine it could be confusing to show examples of class fields like this, as opposed to the simpler message "always initialize Lit properties in the constructor", but I think it's the most correct approach since the "reactive" distinction is the key thing.

packages/lit-dev-content/src/typescript-preference.ts
```diff
@@ -8,6 +8,8 @@ type TypeScriptPreference = 'ts' | 'js';
 
 const LOCAL_STORAGE_KEY = 'typescript-preference';
 
+const BODY_ATTRIBUTE = 'typescript-preference';
```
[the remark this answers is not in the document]
aomarks: Done. Also renamed `typescript-switch` to `code-language-switch` to match.

aomarks:
> LGTM! Small unrelated thought, we can remove the two separate `Hello World (TypeScript)` and `Hello World (JavaScript)` playground examples when we release the toggle.

True! We should do that.

PR #484, opened 2021-09-09
Add JS versions of tutorial code samples

aomarks:
Adds JS versions of code samples in the left-hand-side HTML content of the tutorial.

Note there is some associated *text* in some places (e.g. in the screenshot above) that are decorator-specific. I think we should update the text to be more neutral (either to let the switchable code speak for itself, or to explain the two possibilities). We could also have switches over *text* content, but I think neutral language is simpler and probably better.

Part of #332

### DIFF
--- a/packages/lit-dev-content/site/tutorial/content/06-template-logic.md
+++ b/packages/lit-dev-content/site/tutorial/content/06-template-logic.md
@@ -27,8 +27,8 @@ You can use standard JavaScript in Lit expressions to create conditional or repe
 
     addToDo() {
       this.listItems.push({text: this.input.value, completed: false});
-      this.requestUpdate();
       this.input.value = '';
+      this.requestUpdate();
     }
     ```
 
@@ -39,8 +39,8 @@ You can use standard JavaScript in Lit expressions to create conditional or repe
 
     addToDo() {
       this.listItems.push({text: this.input.value, completed: false});
-      this.requestUpdate();
       this.input.value = '';
+      this.requestUpdate();
     }
     ```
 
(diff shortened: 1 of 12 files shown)


packages/lit-dev-content/site/tutorial/content/06-template-logic.md
```diff
@@ -29,6 +32,20 @@ You can use standard JavaScript in Lit expressions to create conditional or repe
...
     ```
 
+    ```js
+    get input() {
+      return this.renderRoot?.querySelector('#newitem') ?? null;
+    }
+
+    addToDo() {
+      this.listItems.push({text: this.input.value, completed: false});
+      this.requestUpdate();
```
justinfagnani, line 42: while you're here, can you put `requestUpdate()` last? It being in the middle looks more significant than it is to me, while the order doesn't actually matter at all.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -185,7 +190,13 @@ export class LitDevTutorial extends LitElement {
   }
 
   private _onCodeLanguagePreferenceChanged = () => {
-    this._loadStep();
+    // TODO(aomarks) If the user has modified the code, we should show a dialog
```
justinfagnani, line 193: File an issue?

justinfagnani (approved):

packages/lit-dev-content/src/components/litdev-tutorial.ts
aomarks: https://github.com/lit/lit.dev/issues/485

packages/lit-dev-content/site/tutorial/content/06-template-logic.md
aomarks: Done. Also updated all the RHS project files in the same way from this step on.

PR #486, opened 2021-09-09
Add JS versions of all docs code snippets

aomarks:
For this change I didn't want to modify any of the existing TS code or prose, because the language switch is still behind the `?mods=jsSamples` feature flag, so there will be a second pass with a few more changes that we'll merge as part of the public release.

Part of #332

### DIFF
--- a/packages/lit-dev-content/site/docs/components/shadow-dom.md
+++ b/packages/lit-dev-content/site/docs/components/shadow-dom.md
@@ -84,7 +84,7 @@ This decorator is equivalent to:
 
 ```js
 get _first() {
-  return this.renderRoot.querySelector('#first');
+  return this.renderRoot?.querySelector('#first');
 }
 ```
 
--- a/packages/lit-dev-content/site/docs/components/shadow-dom.md
+++ b/packages/lit-dev-content/site/docs/components/shadow-dom.md
@@ -84,7 +84,7 @@ This decorator is equivalent to:
 
 ```js
 get _first() {
-  return this.renderRoot?.querySelector('#first');
+  return this.renderRoot?.querySelector('#first') ?? null;
 }
 ```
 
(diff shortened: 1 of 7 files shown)


packages/lit-dev-content/site/docs/composition/controllers.md
```diff
@@ -70,8 +72,23 @@ class ClockController implements ReactiveController {
 }
 ```
 
+```js
+class ClockController {
+  constructor(host) {
+    // Store a reference to the host
+    this.host = host;
```
justinfagnani, line 79: I wonder if we want to use class fields in JS examples? They're supported in all current browsers now.

justinfagnani (approved):

packages/lit-dev-content/site/docs/components/shadow-dom.md
```diff
@@ -84,7 +84,7 @@ This decorator is equivalent to:
 
 ```js
 get _first() {
-  return this.renderRoot.querySelector('#first');
+  return this.renderRoot?.querySelector('#first');
```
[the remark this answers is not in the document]
aomarks: Done.

packages/lit-dev-content/site/docs/composition/controllers.md
aomarks: In this case do you mean?:

```js
class ClockController {
  host;
  constructor(host) {
    // Store a reference to the host
    this.host = host;
    // Register for lifecycle updates
    host.addController(this);
  }
}
```

justinfagnani: Yeah. Though I'm not even sure myself whether that's better or not...

aomarks: I think I'll leave it off for now, but can revisit.

edimitchel: Or may be more simple: 
```js
class ClockController {
  constructor(private host) {
    // Register for lifecycle updates
    host.addController(this);
  }
}
```

PR #487, opened 2021-09-09
Hand-roll the JS output for the 2nd tutorial step

aomarks:
This file intentionally contains unused imports, and we don't want TypeScript to remove them.

Part of #332


justinfagnani (approved):

aomarks:
Oops had a bug where we weren't using the TypeScript custom JSON parser for parsing the config file -- it's different because it allows comments and trailing commas! (Was using it for `--watch` mode, but not regular build mode, hence why I missed it). Fixed.

PR #491, opened 2021-09-09
Add TOC link for page headers

### LINKED ISSUE #459: Page table of contents does not allow navigating to the page overview
The main heading of the page doesn't appear in the page TOC, so clicking on the first link in the TOC doesn't navigate to the top of the page.

We should include the page header in the TOC, possibly like: **Contents: Defining a component** or **Defining a component: Contents**

justinfagnani:
Fixes #459 

This lets users navigate to the top of the page with the TOC.

It's not perfect, as the current section indicator still tops out at the first `<h2>`, but it's better than before IMO.


aomarks (approved):
Nice

PR #494, opened 2021-09-10
Fixes for code sample fold regions

aomarks:
- Fix broken folding range. In this example, the `/* playground-fold */` comment was associated with the decorators `import` statement. Since that import gets removed as part of our transform, the comment was getting removed too. The quick fix here is to just shift the comment down. This means the decorator import is now visible in TS mode, but that seems fine.

- Don't auto-format embedded HTML/CSS. By default, Prettier automatically formats HTML and CSS inside html and css backtick functions. This could be good with some tweaks, but right now it seems to just be worse, since we already took care to manually format our HTML and CSS.

- Improve visible line calculation. Previously we did not account for the fact that a folded region gets a clickable `...`, which is slightly taller than a normal code line.

### DIFF
--- a/packages/lit-dev-tools/src/playground-plugin/plugin.ts
+++ b/packages/lit-dev-tools/src/playground-plugin/plugin.ts
@@ -34,37 +34,39 @@ const getNumVisibleLinesForProjectFile = async (
   project: string,
   filename: string
 ): Promise<{ts: number; js: number}> => {
-  const tsData = removeHiddenRegions(
-    filename,
-    await fs.readFile(`samples/${project}/${filename}`, 'utf8')
-  );
-  const tsLines = tsData.split('\n').length;
+  const tsData = await fs.readFile(`samples/${project}/${filename}`, 'utf8');
+  const tsLines = countVisibleLines(filename, tsData);
   let jsLines;
   if (filename.endsWith('.ts')) {
     const jsFilename = filename.replace(/\.ts$/, '.js');
-    const jsData = removeHiddenRegions(
-      filename,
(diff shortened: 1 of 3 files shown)


packages/lit-dev-tools/src/playground-plugin/plugin.ts
```diff
@@ -34,37 +34,39 @@ const getNumVisibleLinesForProjectFile = async (
...
+  const pattern = filename.endsWith('.html')
+    ? /( *<!-- *playground-(?<kind>hide|fold) *-->\n?)(?:(.*?)( *<!-- *playground-\k<kind>-end *-->\n?))?/gs
+    : /( *\/\* *playground-(?<kind>hide|fold) *\*\/\n?)(?:(.*?)( *\/\* *playground-\k<kind>-end *\*\/\n?))?/gs;
+  for (const [content, , kind] of code.matchAll(pattern)) {
+    // Remove the lines within the hidden/folded region.
+    count -= content.split('\n').length;
+    if (kind === 'fold') {
+      // In the fold case a clickable "..." is put in its place. For some
+      // reason, these lines have slightly more height than a normal code line.
+      count += 1.03;
```
justinfagnani, line 66: it'd be nice to fix this in CSS. line-height? Fine for now

justinfagnani (approved):

packages/lit-dev-tools/src/playground-plugin/plugin.ts
aomarks: Yeah, added TODO for now.

PR #495, opened 2021-09-10
Convert static styles/properties getter to class fields

aomarks:
- Convert static styles getters to static class fields
- Convert static properties getters to static class fields
- Replace invalid curly quotes with straight quotes in code samples
- Update localization docs for JS

### DIFF
--- a/packages/lit-dev-content/samples/docs/mixins/highlightable/element-two.ts
+++ b/packages/lit-dev-content/samples/docs/mixins/highlightable/element-two.ts
@@ -4,12 +4,10 @@ import {Highlightable} from './highlightable.js'
 
 @customElement('element-two')
 export class ElementTwo extends Highlightable(LitElement) {
-  static get styles() {
-    return [
-      super.styles || [],
-      css`:host { display: block; }`
-    ];
-  }
+  static styles = [
+    super.styles || [],
+    css`:host { display: block; }`
+  ];
   render(){
     return this.renderHighlight(html`
       <label>
--- a/packages/lit-dev-content/site/docs/components/properties.md
+++ b/packages/lit-dev-content/site/docs/components/properties.md
@@ -97,16 +97,14 @@ The argument to the `@property`  decorators is an [options object](#property-opt
 
 ### Declaring properties in a static properties field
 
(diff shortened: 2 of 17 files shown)


packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -97,16 +97,14 @@ The argument to the `@property`  decorators is an [options object](#property-opt
 
 ### Declaring properties in a static properties field
 
-To declare properties in a static `properties` field:
+To declare properties in a static `properties` class field:
```
AndrewJakubowicz, line 100: nit: Add class to the heading?

aomarks: Done.

packages/lit-dev-content/samples/docs/mixins/highlightable/element-two.ts
```diff
@@ -4,12 +4,10 @@ import {Highlightable} from './highlightable.js'
...
 @customElement('element-two')
 export class ElementTwo extends Highlightable(LitElement) {
-  static get styles() {
-    return [
-      super.styles || [],
-      css`:host { display: block; }`
-    ];
-  }
+  static styles = [
+    super.styles || [],
```
justinfagnani, line 8:
```suggestion
    super.styles ?? [],
```

justinfagnani (approved):

arthurevans:
Have we decided that this style is OK now? When we released the first RC, the call was still to use the static getters because static class fields were problematic... somewhere. Yeah, I wish I remembered, too. Maybe @justinfagnani does.

aomarks:
> Have we decided that this style is OK now? When we released the first RC, the call was still to use the static getters because static class fields were problematic... somewhere. Yeah, I wish I remembered, too. Maybe @justinfagnani does.

Static class fields were added in Chrome 72 (January 2019), Firefox 75 (April 2020), Safari 14.1 (May 2021).

So I assume the decision was made because Safari didn't have them yet when the lit.dev docs were first written, but now it does.

arthurevans:
Yep--that was it. Thanks, Al!

PR #496, opened 2021-09-11
Update TS/JS toggle button design

aomarks:
Trying to make it look more interactive. Open to suggestions/tweaks.

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-code-language-switch.ts
+++ b/packages/lit-dev-content/src/components/litdev-code-language-switch.ts
@@ -9,7 +9,6 @@ import {
   getCodeLanguagePreference,
   setCodeLanguagePreference,
   CODE_LANGUAGE_PREFERENCE_EVENT_NAME,
-  CodeLanguagePreference,
 } from '../code-language-preference.js';
 
 /**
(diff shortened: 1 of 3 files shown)


AndrewJakubowicz (commented):
Haven't gone through code yet. Just some quick initial feedback.
 - It looks amazing! Tried with keyboard controls and it works well.
 - Tested on Windows: Firefox & Chrome

Chrome suggests a slightly larger contrast between the white text and background:
image

packages/lit-dev-content/src/components/litdev-code-language-switch.ts
```diff
@@ -25,45 +24,70 @@ export class LitDevCodeLanguageSwitch extends LitElement {
...
     }
 
-    button:not([disabled]) {
-      cursor: pointer;
+    button:hover > #toggle {
+      background: #005cc5bd;
+    }
+
+    [position='js'] > #toggle {
+      left: 0;
```
AndrewJakubowicz, line 70: elegant!

packages/lit-dev-content/src/components/litdev-code-language-switch.ts
```diff
@@ -25,45 +24,70 @@ export class LitDevCodeLanguageSwitch extends LitElement {
...
-    button[disabled],
-    button:hover {
-      background: #ebeeff;
-      border-color: #7589ff;
+    #toggle {
+      position: absolute;
+      width: 50%;
+      height: 100%;
+      top: 0;
+      transition: left 100ms;
```
AndrewJakubowicz, line 58: Not required but maybe could add [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)? You can sometimes get many moving toggles on the same view.

packages/lit-dev-content/src/components/litdev-code-language-switch.ts
```diff
@@ -93,33 +117,20 @@ export class LitDevCodeLanguageSwitch extends LitElement {
...
-
-      <button
-        title="Display code as TypeScript"
-        aria-label="Display code as TypeScript"
-        ?disabled=${mode === 'ts'}
-        @click=${this._onClickTs}
+        role="switch"
+        aria-checked=${mode == 'ts' ? 'true' : 'false'}
+        aria-title="Toggle TypeScript"
+        title=${mode === 'ts' ? 'Disable TypeScript' : 'Enable TypeScript'}
```
AndrewJakubowicz, line 123: No action required nit: Should the titles be more along the lines of `Toggle between TypeScript and JavaScript`, rather than enabling/disabling TypeScript?

AndrewJakubowicz (approved):
This looks fantastic! LGTM

packages/lit-dev-content/src/components/litdev-code-language-switch.ts
aomarks: Cool, TIL. Done.

packages/lit-dev-content/src/components/litdev-code-language-switch.ts
aomarks: > Chrome suggests a slightly larger contrast between the white text and background:

Fixed!

aomarks: > No action required nit: Should the titles be more along the lines of `Toggle between TypeScript and JavaScript`, rather than enabling/disabling TypeScript?

Good question, not 100% sure about this but ... the reason I used "Toggle TypeScript" for the `aria-label` is that it seemed like `role=switch` was the best role to use, which means that the state of the button is expressed in the `aria-checked` attribute. It seemed like "Toggle between TypeScript and JavaScript" would be confusing in that case, because it's not clear whether TypeScript or JavaScript maps to true vs false. When using voiceover, this reads as "Toggle TypeScript" and then "on"/"off" as you switch between them. 

And for `title`, it also seemed good to make it the same boolean choice for visual users, too.

justinfagnani:
This looks great!

A couple of design notes:

- The vertical alignment needs a small adjustment to make it visually centered I think. I added a 1px bottom padding to the <button> which helped.
- The horizontal centering of the toggle could maybe use an adjustment too. I see there's a .25em left padding, which could be bumped up slightly (1px, but in em I suppose) on the JS label, but reduced slightly on the TS label.
- The toggle drop shadow makes it look a little blurry to me, especially in the gray / non-hover state. Maybe drop the opacity to 20%? Another option is to only show the drop shadows on hover.
- Maybe the button could have a transparent background, then white on hover?

aomarks:
> * The vertical alignment needs a small adjustment to make it visually centered I think. I added a 1px bottom padding to the  which helped.
> * The horizontal centering of the toggle could maybe use an adjustment too. I see there's a .25em left padding, which could be bumped up slightly (1px, but in em I suppose) on the JS label, but reduced slightly on the TS label.

I made some changes to help with centering, including some new padding. PTAL.

Also, I'm trying a technique with units that just occurred to me. Pick some default `font-size`, and then use `calc(<px>em / <default-font-size>)` everywhere else, where `<px>` is the pixel measurement you want at the default font size. That way you can easily write pixel-perfect units for the default size, but still allow arbitrary scaling.

> * The toggle drop shadow makes it look a little blurry to me, especially in the gray / non-hover state. Maybe drop the opacity to 20%? Another option is to only show the drop shadows on hover.

I removed all the drop shadows (also the inset one). Definitely crisper, I think I prefer this look. WDYT?

> * Maybe the button could have a transparent background, then white on hover?

I think anything we do with opacity would drop the contrast so that it's no longer accessible. The white-on-gray is already the minimum contrast that passes Chrome's a11y check.

justinfagnani:
> I think anything we do with opacity would drop the contrast so that it's no longer accessible. The white-on-gray is already the minimum contrast that passes Chrome's a11y check.

Are there any a11y guidelines for hover-over effects like this? Like, does the non-hovered state have to have the same contrast as the hovered state? Certainly if we had decide to only show the switch at all on hover, the contrast of the non-hovered state wouldn't be an issue.

justinfagnani (approved):

aomarks:
> Are there any a11y guidelines for hover-over effects like this? Like, does the non-hovered state have to have the same contrast as the hovered state? Certainly if we had decide to only show the switch at all on hover, the contrast of the non-hovered state wouldn't be an issue.

Good question. I would think it would be important for the non-hover state to meet contrast guidelines -- otherwise how would somebody who couldn't perceive it clearly even know to hover over it? And lighthouse would definitely dock our score regardless.

PR #499, opened 2021-09-13
Fix the style accessor unblocking broken search

AndrewJakubowicz:
### Context

Search is currently broken on lit.dev.

This fixes the style. Didn't have time to investigate why this is happening.
Also tried using `LionOption.style` but it didn't work.

Reverted back to a static style().

Also ran Prettier on the file.

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-search.ts
+++ b/packages/lit-dev-content/src/components/litdev-search.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {LitElement, html, css, TemplateResult, nothing} from 'lit';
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -355,12 +355,13 @@ class LitdevSearchOption extends LionOption {
...
   heading = '';
 
-  @property({type: Boolean})
+  @property({ type: Boolean })
   isSubsection = false;
 
-  static styles = [
-    ...super.styles,
-    css`
+  static get styles() {
```
AndrewJakubowicz, line 361: This is the change that fixed the bug.

aomarks (approved):

Issue #485, opened 2021-09-09
Alert user about losing changes when switching playground languages


aomarks:
This could happen both in the Playground page and the Tutorial page. When the user switches between TypeScript/JavaScript, we reload a new project, blowing away any changes we might have made.

Something along the lines of "Switching languages will lose your changes [Accept] [Cancel]".

Part of this should include a new API from playground-elements for checking whether a project has been modified.

PR #500, opened 2021-09-14
Guard against losing changes when switching languages

aomarks:
If the user has modified code on the Tutorial or Playground pages, and then clicks the change-language toggle, a dialog will now pop up telling them that they'll lose changes, and gives them a chance to cancel.

image

Also brings in the new package exports support (https://github.com/PolymerLabs/playground-elements/pull/200), so Lit will now run in development mode, and more packages will be importable (like Lion elements).

Fixes https://github.com/lit/lit.dev/issues/485

### DIFF
--- a/packages/lit-dev-content/src/code-language-preference.ts
+++ b/packages/lit-dev-content/src/code-language-preference.ts
@@ -14,8 +14,34 @@ const BODY_ATTRIBUTE = 'code-language-preference';
  * Name of the event that is fired on window whenever the global user TypeScript
  * preference changes.
  */
-export const CODE_LANGUAGE_PREFERENCE_EVENT_NAME =
-  'code-language-preference-changed';
+export const CODE_LANGUAGE_CHANGE = 'code-language-change';
+
+/**
+ * Name of the event that is fired on window before the global user code
+ * language changes, providing an opportunity to cancel it.
+ */
+export const BEFORE_CODE_LANGUAGE_CHANGE = 'before-code-language-change';
+
+/**
+ * Details of the before-code-language-change event.
+ */
+interface BeforeCodeLanguageChangeDetail {
+  /**
+   * The new language that will be changed to.
+   */
+  pendingLanguage: CodeLanguagePreference;
+
+  /**
+   * If this function is called, this language change will be aborted.
+   */
+  cancel: () => void;
+}
+
+declare global {
+  interface WindowEventMap {
+    [BEFORE_CODE_LANGUAGE_CHANGE]: CustomEvent<BeforeCodeLanguageChangeDetail>;
+  }
+}
 
 /**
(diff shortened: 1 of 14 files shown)


packages/lit-dev-content/src/code-language-preference.ts
```diff
@@ -30,10 +56,26 @@ export const getCodeLanguagePreference = (): CodeLanguagePreference =>
...
+  preference: CodeLanguagePreference,
+  force = false
 ): void => {
+  if (!force) {
+    let cancelled = false;
+    const detail: BeforeCodeLanguageChangeDetail = {
+      pendingLanguage: preference,
+      cancel: () => {
+        cancelled = true;
+      },
```
AndrewJakubowicz, line 68: Very clever synchronous message passing. It's very nice and avoids any async code.

packages/lit-dev-content/src/components/litdev-playground-change-guard.ts
```diff
@@ -0,0 +1,107 @@
...
+import {
+  BEFORE_CODE_LANGUAGE_CHANGE,
+  setCodeLanguagePreference,
+} from '../code-language-preference.js';
+
+import type {CodeLanguagePreference} from '../code-language-preference.js';
+
+/**
+ * Prompts the user to continue or cancel if the global code language preference
+ * is about to change if the associated Playground has been modified.
```
AndrewJakubowicz, line 22: Nit: "is about to change ~if~ __and__ the associated"

packages/lit-dev-content/src/code-language-preference.ts
```diff
@@ -30,10 +56,26 @@ export const getCodeLanguagePreference = (): CodeLanguagePreference =>
  * a change event on window.
```
AndrewJakubowicz, line 56: May be worth adding some docs that this function now has two behaviors: what is written, and initiating a user prompted language change.

AndrewJakubowicz (approved):
LGTM. Played around a little manually.

It's a very cool design where the modal can take ownership of the event. Cancel it. Then force the change via user action. All with synchronous events.

May be worth a tiny integration test in the future.

packages/lit-dev-content/src/code-language-preference.ts
aomarks: Good idea, done!

packages/lit-dev-content/src/components/litdev-playground-change-guard.ts
aomarks: Thanks! Done.

aomarks (commented):
> May be worth a tiny integration test in the future.

Agreed! Filed https://github.com/lit/lit.dev/issues/502 for myself.

Issue #490, opened 2021-09-09
Unify home-page JS/TS switch with samples switch


justinfagnani:
Should the main example on the homepage use the same JS/TS preference as the other samples? And should the switch be styled the same to be consistent?

aomarks:
Switch should be styled the same yeah.

I'm not 100% sure about making it sync'd with the localStorage setting though. I would be concerned that somebody would toggle it on the homepage just because it's a fun / interesting thing to do, and then not realize they were missing out on the TypeScript docs as our default/recommended? Just sort of feels like an independent toy on the homepage, I guess.

AndrewJakubowicz:
Regarding our recommended option being Typescript, we could also save the state in sessionStorage so it's a little less sticky.  I have no preference since the switch is pretty easy (and fun) to change.

PR #501, opened 2021-09-14
Final JS docs fixes and launch

aomarks:
- Replace home-page-specific language toggle with the new global one.

- Replace the separate JavaScript vs TypeScript "Hello World" Playground examples with a unified one. I'm keeping the JS output hand-rolled for this example for now, because the `static properties` block isn't in the perfect position, and this is a very prominent example. Will fix in https://github.com/lit/lit/issues/2159 as followup.

  

- Fix home page highlight positions when in JS mode. This actually broke in https://github.com/lit/lit.dev/pull/495 where I forgot to update the ranges for the yellow-highlight effect.

- Added a check for generated samples that have lost or re-ordered playground fold/hide comments. This should help detect weird cases that might need hand-rolling going forward.

- Hand-roll two examples that are losing fold comments, and move one comment to be trailing so that it's not lost, that I found via the above check. Filed https://github.com/lit/lit/issues/2158 to hopefully address in the transformers in the future.

- Remove some unnecessary TypeScript vs JavaScript language, and add a few more switchable samples.

- Remove `jsSamples` mod feature flag to launch JS docs.

Fixes https://github.com/lit/lit.dev/issues/332
Fixes https://github.com/lit/lit.dev/issues/490


packages/lit-dev-content/site/docs/index.md
```diff
@@ -42,7 +42,7 @@ Some things to note:
 * Lit’s [expressive, declarative templates](/docs/templates/overview/) (utilizing JavaScript tagged template literals) make it easy to describe how a component should be rendered.
 * [Reactive properties](/docs/components/properties/) represent a component’s public API and/or internal state; your component automatically re-renders whenever a reactive property changes.
 * [Styles](/docs/components/styles) are scoped by default, keeping your CSS selectors simple and ensuring that your component’s styling won’t pollute (or be polluted by) the surrounding context.
-* This example is in TypeScript, but TypeScript is 100% optional–Lit works great in JavaScript, too.
+* Lit works great in vanilla JavaScript, or you can use TypeScript for more even better ergonomics by using decorators and strong types.
```
justinfagnani, line 45: maybe "strong types" -> "type declarations"?

justinfagnani (approved):
🎉

justinfagnani:
Thanks for the easy-to-review commits!

packages/lit-dev-content/site/docs/index.md
AndrewJakubowicz, line 45: nit: "for ~more~ even better"

aomarks: Done.

aomarks: But it **is** more even better!

(Fixed.)

AndrewJakubowicz (approved):
Code looks good. Did some manual testing on home page, tutorial & playground.

Nit: You can undo the language change clear which leaves you in the opposite language. Fine as a follow up.

aomarks:
> Nit: You can undo the language change clear which leaves you in the opposite language. Fine as a follow up.

Oh yeah, that's very related to a bug that I've been meaning to address for a long time! You can actually undo *between* files too, which is totally weird! It's because we use a single CodeMirror instance for each Playground, and just swap out the code + language highlight mode. We need to either manage undo history if possible, otherwise maintain a separate CodeMirror instance for each file: https://github.com/PolymerLabs/playground-elements/issues/154

PR #506, opened 2021-09-18
Update lit.dev to Lit 2

aomarks:
Shaves off 41kB of transferred byes.

Note we had both versions before, because we are using a Lion component, which was already upgraded to Lit 2.

I did need to include a small `postinstall` hack for an MWC element that had a missing `.js` extension on its import. I've got a change out to fix that, though, so we should be able to remove that pretty soon.

# Before

# After


justinfagnani (approved):
🎉

PR #508, opened 2021-09-21
Fix outdated eventOptions import

OttoWinter:
Noticed this while browsing the docs, it seems like this section was using an outdated import path

(new to lit, please let me know if the current is indeed correct)


OttoWinter:
@googlebot I signed it!

PR #513, opened 2021-09-23
Factor out redirect middleware and run in dev server

aomarks:
Previously, our redirect logic did not work when running in dev mode, since that code was only present in the production server. Now that logic is factored out into a Koa middleware, so that it can run in both places.

I also made a few other changes to make working on the server easier going forward.

In an upcoming PR, I'll be adding a middleware to deal with GitHub auth tokens, so now it will be easier to run that both in production and during local development.


packages/lit-dev-server/tsconfig.json
```diff
@@ -1,13 +1,13 @@
 {
   "compilerOptions": {
     "incremental": true,
-    "tsBuildInfoFile": "./tsconfig.tsbuildinfo",
+    "tsBuildInfoFile": "./lib/.tsbuildinfo",
```
AndrewJakubowicz, line 4: Tiny nit: Other packages kept this as `"./lib/tsconfig.tsbuildinfo"`.

AndrewJakubowicz (approved):
LGTM. Also checked it in manually and navigated to a redirect. 👏

packages/lit-dev-server/tsconfig.json
aomarks: Hm interesting. The default is actually `<outdir>/.tsbuildinfo` which matches this path (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html). So we could also just omit it.

aomarks: I might go through and standardize these in a followup.

Issue #469, opened 2021-09-01
Screenshot Integration tests CI feature


AndrewJakubowicz:
Follow up issue from https://github.com/lit/lit.dev/pull/463

We need a way to generate playwright screenshot goldens using the CI environment. Otherwise there are render issues between operating systems.

AndrewJakubowicz:
Done!

PR #515, opened 2021-09-23
Downloading golden snapshots from Github Actions

AndrewJakubowicz:
Addresses issue: https://github.com/lit/lit.dev/issues/469

### Context

This provides a Github action that will initially run on every PR. It provides a download link of just the screenshots (see the action artifact in this PR below).

Screen Shot 2021-09-23 at 4 31 35 PM

These can be pasted into the directory and uploaded and should pass the integration tests.

I've also changed the functionality such that playwright will fail if a snapshot is missing so it's a failing test to not upload a snapshot.

### Features

 - `npm run test:update-goldens` will regenerate all the snapshots.
 - New github action: `Artifacts / Download link for updated snapshots`. Clicking on this action provides a link for downloading the screenshots generated in Github Actions.
 - Added a home page screenshot test and a playground test. Playground snapshot requires waiting for the loading bar to transition out or it is flaky.

### Future work

 - Not generate a `zip` archive of snapshots on every PR.


.gitignore
```diff
@@ -15,5 +15,9 @@ packages/lit-dev-api/lit/
 packages/lit-dev-api/api-data/**/*
 !packages/lit-dev-api/api-data/.gitkeep
 
+tests/**/*.png
```
aomarks, line 18: Why do we have this git pattern -- shouldn't the golden screenshots get checked in?

README.md
```diff
@@ -114,3 +114,9 @@ docker build -t litdev . --build-arg PLAYGROUND_SANDBOX=http://localhost:8081/
 docker run --rm --name litdev -p 8080:8080 -e PORT=8080 -e MODE=main litdev
 docker run --rm --name litdev-playground -p 8081:8081 -e PORT=8081 -e MODE=playground litdev
 ```
+
+### Updating snapshot tests
+
+Snapshots that are committed to the repository need to be created through the
+"Artifacts / Download link for updated snapshots" Github Action. This action
+generates a zip archive `golden-results` which can be extracted into `tests/`.
```
aomarks, line 122: Maybe provide a sample command? `unzip golden-results.zip -d tests/` or similar?

aomarks (changes requested):
This is great! Really simple and elegant solution.

.github/workflows/generate-goldens.yml
```diff
@@ -0,0 +1,37 @@
...
+# downloaded and added to your PR.
+name: Artifacts
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  generate-goldens:
+    name: Download link for updated snapshots
```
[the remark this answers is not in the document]
AndrewJakubowicz: I agree with you! Will update!

.gitignore
AndrewJakubowicz: Great question. Added this so we ignore any created locally. For example `-darwin.png`. This is paired with the gitignore beneath it that allows the `-linux.png` screenshots to be added.

Could make it more explicit with a single rule rather than two.

.github/workflows/generate-goldens.yml
```diff
@@ -0,0 +1,37 @@
+# GitHub Action that generates screenshots for integration tests. These can be
+# downloaded and added to your PR.
+name: Artifacts
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
AndrewJakubowicz: I haven't yet been able to use it since I think the workflow needs to be merged to `main`. I was thinking once this is merged we can remove the pull and push request trigger (although the size of the artifact is currently tiny).

README.md
AndrewJakubowicz: Great suggestion. Manually tested and added command to readme.

AndrewJakubowicz:
Hmmm. An integration test failed and there's no way to debug it.

Going to add the artifact upload to the integration test if it fails so it's possible to debug the failure quickly.

.gitignore
aomarks: Gotcha! Maybe just add a brief `#` comment about how we only want to keep linux screenshots since that's what we test on CI.

aomarks (approved):
Great, LGTM! Assume you'll figure out the integration test failure.

Issue #462, opened 2021-08-31
Move class field discussion from Decorators to Reactive properties


arthurevans:
This section (https://lit.dev/docs/components/decorators/#avoiding-issues-with-class-fields) is not specific to decorators, and should be in the section on properties, not the section on class fields.

arthurevans:
Update this section per discussions.

PR #516, opened 2021-09-24
Beef up docs around using class fields

sorvell:
Fixes https://github.com/lit/lit.dev/issues/462

### DIFF
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -91,7 +91,7 @@ If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you ca
 
 Note, the `@babel/plugin-proposal-class-properties` may not be required with the latest versions of Babel.
 
-To enable these settings, add code like this to your Babel configuration:
+To set up the plugins, add code like this to your Babel configuration:
 
 ```js
 assumptions = {
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -84,7 +84,7 @@ Enabling `emitDecoratorMetadata` is not required and not recommended.
 
 ### Using decorators with Babel  { #decorators-babel }
 
-If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins and assumptions:
+If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins and settings:
 
 *   [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/en/babel-plugin-proposal-decorators)
 *   [`@babel/plugin-proposal-class-properties`](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties)
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -73,7 +73,7 @@ In the future when decorators become a native web platform feature, this may no 
 
 To use decorators with [TypeScript](https://www.typescriptlang.org/docs/handbook/decorators.html), enable the `experimentalDecorators` compiler option.
 
-You should also ensure that the `useDefineForClassFields` setting is `false`. Note, this should only be required when the `target` is set to `es2020` or greater, but it's recommended to explicitly ensure this setting is `false`.
+You should also ensure that the `useDefineForClassFields` setting is `false`. Note, this should only be required when the `target` is set to `esnext` or greater, but it's recommended to explicitly ensure this setting is `false`.
 
 ```json
 "experimentalDecorators": true,
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -73,25 +73,34 @@ In the future when decorators become a native web platform feature, this may no 
 
 To use decorators with [TypeScript](https://www.typescriptlang.org/docs/handbook/decorators.html), enable the `experimentalDecorators` compiler option.
 
+You should also ensure that the `useDefineForClassFields` setting is `false`. Note, this should only be required when the `target` is set to `es2020` or greater, but it's recommended to explicitly ensure this setting is `false`.
+
 ```json
 "experimentalDecorators": true,
+"useDefineForClassFields": false,
 ```
 
 Enabling `emitDecoratorMetadata` is not required and not recommended.
 
 ### Using decorators with Babel  { #decorators-babel }
 
-If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins:
+If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins and assumptions:
 
-*   [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/en/babel-plugin-proposal-decorators).
+*   [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/en/babel-plugin-proposal-decorators)
 *   [`@babel/plugin-proposal-class-properties`](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties)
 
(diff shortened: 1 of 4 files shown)


packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -115,11 +115,36 @@ class MyElement extends LitElement {
 
 An empty option object is equivalent to specifying the default value for all options.
 
-<div class="alert alert-info">
+### Avoiding issues with class fields when declaring properties {#avoiding-issues-with-class-fields}
 
-**If you're using the static properties field, initialize properties in the constructor**. Class field initializers won't work in this case.
+[Class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields) have a problematic interaction with declaring reactive properties. Class fields are defined on the element instance. Reactive properties are defined as accessors on the element prototype. According to the rules of JavaScript, an instance property takes precedence over and effectively hides a prototype property. This means that reactive property accessors do not function when class fields are used. When a property is set, the element does not update.
```
aomarks, line 120:
```suggestion
[Class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields) have a problematic interaction with reactive properties. Class fields are defined on the element instance. Reactive properties are defined as accessors on the element prototype. According to the rules of JavaScript, an instance property takes precedence over and effectively hides a prototype property. This means that reactive property accessors do not function when class fields are used. When a property is set, the element does not update.
```

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -115,11 +115,36 @@ class MyElement extends LitElement {
...
 
-</div>
+**You can use class fields** with reactive properties when transpiling code with Babel or TypeScript and the proper settings are enabled. Both Babel and TypeScript support transpiling class fields into properties set in the constructor rather than defined on the instance.
```
aomarks, line 122: Suggest removing this paragraph, kinda seems like it's mostly redundant with the next two paragraphs, which are more focused.

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -115,11 +115,36 @@ class MyElement extends LitElement {
...
+
+For TypeScript, set the `useDefineForClassFields` setting in your `tsconfig` is set to `false`. Note, this is not required for some configurations of TypeScript, but it's recommended to explicitly set it to `false`.
```
aomarks, line 124:
```suggestion
For **TypeScript**, you may use class fields for declaring reactive properties as long as the `useDefineForClassFields` setting in your `tsconfig` is set to `false`. Note, this is not required for some configurations of TypeScript, but it's recommended to explicitly set it to `false`.
```

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -115,11 +115,36 @@ class MyElement extends LitElement {
...
+
+For Babel, set `setPublicClassFields` to `true` in the `assumptions` config of your `babelrc`. Note, for older versions of Babel, you also need to include the plugin `@babel/plugin-proposal-class-properties`:
```
aomarks, line 126:
```suggestion
When compiling JavaScript with **Babel**, you may use class fields for declaring reactive properties as long as you set `setPublicClassFields` to `true` in the `assumptions` config of your `babelrc`. Note, for older versions of Babel, you also need to include the plugin `@babel/plugin-proposal-class-properties`:
```

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -115,11 +115,36 @@ class MyElement extends LitElement {
...
+};
+
+plugins = [
+  ["@babel/plugin-proposal-class-properties"],
+];
+```
+
+For information about using class fields with **decorators**, see [Avoiding issues with class fields and decorators](/docs/components/decorators/#avoiding-issues-with-class-fields).
+
+**You cannot use *native* class fields** or their emulated output in Babel or TypeScript with properties intended to be reactive. Instead, properties should be initialized in the element constructor:
```
aomarks, line 140: I also suggest moving this section to the top, so that the 3 scenarios are (in order): JavaScript, TypeScript, Babel

```suggestion
In **JavaScript** you must not use class fields when declaring reactive properties. Instead, properties must be initialized in the element constructor:
```

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -73,25 +73,34 @@ In the future when decorators become a native web platform feature, this may no
 
 To use decorators with [TypeScript](https://www.typescriptlang.org/docs/handbook/decorators.html), enable the `experimentalDecorators` compiler option.
 
+You should also ensure that the `useDefineForClassFields` setting is `false`. Note, this should only be required when the `target` is set to `es2020` or greater, but it's recommended to explicitly ensure this setting is `false`.
```
aomarks, line 76: I believe it's only for `esnext`, not `es2020` (https://github.com/microsoft/TypeScript/issues/34787)

aomarks (changes requested):
Nice! I just have a few suggestions for the "Avoiding issues" section.

packages/lit-dev-content/site/docs/components/decorators.md
sorvell: Fixed

packages/lit-dev-content/site/docs/components/properties.md
sorvell: Done.

aomarks (approved):
Looks great!

aomarks (changes requested):
Oh, we can also now update the `/msg/class-field-shadowing` short link to point to this new section!

https://github.com/lit/lit.dev/blob/a8b7ea747236df24321d05bb7d8556a52fa69040/packages/lit-dev-server/src/redirects.ts#L14

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -73,25 +73,34 @@ In the future when decorators become a native web platform feature, this may no
...
 "experimentalDecorators": true,
+"useDefineForClassFields": false,
 ```
 
 Enabling `emitDecoratorMetadata` is not required and not recommended.
 
 ### Using decorators with Babel  { #decorators-babel }
 
-If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins:
+If you're compiling JavaScript with [Babel](https://babeljs.io/docs/en/), you can enable decorators by adding the following plugins and assumptions:
```
arthurevans, line 87: Using "assumptions" here is super jargony. You'd have to be pretty well-versed in Babel to know that there are config options called  "Compiler assumptions," and the average user  is likely to infer the generic meaning for assumptions here.

Suggest either just go with "plugins" and assume (heh) that the configuration steps that follow are part of adding the plugins, or say, "plugins and settings"

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -73,25 +73,34 @@ In the future when decorators become a native web platform feature, this may no
...
 
-*   [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/en/babel-plugin-proposal-decorators).
+*   [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/en/babel-plugin-proposal-decorators)
 *   [`@babel/plugin-proposal-class-properties`](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties)
 
-To enable the plugins, add code like this to your Babel configuration:
+Note, the `@babel/plugin-proposal-class-properties` may not be required with the latest versions of Babel.
+
+To enable these settings, add code like this to your Babel configuration:
```
arthurevans, line 94: "these settings" doesn't really refer back to anything.

```suggestion
To set up the plugins, add code like this to your Babel configuration:
```

arthurevans (approved):
LGTM, a couple of wording suggestions.

packages/lit-dev-content/site/docs/components/decorators.md
sorvell: Changed to "plugins and settings"

aomarks (approved):

Issue #517, opened 2021-09-24
Set up CSP policy for lit.dev


aomarks:
A good idea generally, but will be especially important once we are handling GitHub auth tokens.

PR #519, opened 2021-09-24
Run a separate dev server for the playground sandbox

aomarks:
Previously, during local development, we ran the Playground sandbox on the same origin as the main site for simplicity. On production, though, we run a separate `playground.lit.dev` server for the sandbox. We do this in production for security, because we want to run untrusted code in an isolated context.

Now during local dev we launch two `@web/dev-servers`: one for the main site, and one for the playground sandbox, mirroring what we do on production.

The ports are also now static: `5415` for the main dev server, and `5416` for the playground dev server. Previously the dev server opened on the first port that was available starting from `8000`. We need static ports because Eleventy now needs to know what the Playground sandbox URL will be, since it is baked into the HTML it generates. I picked these numbers mostly arbitrarily.

I've also done the equivalent for the "local production" configuration (`npm start`), and updated the docs for Docker. So there's now a port scheme that is hopefully easy to remember:

- Local dev server: port `5415 / 5416`
- Local production server: port `6415 / 6516`
- Local production server using docker: port `7415 / 7416`

The main motivation for this was upcoming work to set up a CSP policy (https://github.com/lit/lit.dev/issues/517). We need a different CSP policy on the main vs playground sandbox servers, because the main server _must not_ run untrusted code, while the playground sandbox _must be able to_.


packages/lit-dev-tools-esm/src/dev-server.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return;
+  },
+};
+
+/**
+ * Since we're using the `watch` setting, @web/dev-server injects a script into
+ * any .html file it serves that listens for changes and reloads the page.
+ *
+ * However, when serving a .html file from a Playground project, we want this
```
AndrewJakubowicz, line 45: typo at end: "we **don't** want this"

packages/lit-dev-tools-esm/src/dev-server.ts
```diff
@@ -0,0 +1,97 @@
...
+    return;
+  },
+};
+
+startDevServer({
+  config: {
+    port: MAIN_PORT,
+    rootDir: pathlib.join(CONTENT_PKG, '_dev'),
+    plugins: [
+      dontResolveBareModulesInPlaygroundFiles,
```
AndrewJakubowicz, line 74: Clarifying. (What follows is a very generic statement of how I interpret this): We want to send down something that hasn't had the imports manipulated such that it works in the playground. This is because the playground handles its own module resolution?

packages/lit-dev-tools-esm/src/dev-server.ts
```diff
@@ -0,0 +1,97 @@
...
+ * See https://github.com/modernweb-dev/web/issues/761 for a feature request to
+ * make this easier.
+ */
+const removeWatchScriptFromPlaygroundFiles: DevServerPlugin = {
+  name: 'remove-watch-script-from-playground-files',
+  transform(ctx) {
+    if (ctx.url.startsWith('/samples/')) {
+      return {
+        body: (ctx.body as string).replace(
+          /<!-- injected by web-dev-server.*<\/script>/gs,
```
AndrewJakubowicz, line 60: This is cool. Manually looked for this and it only shows up once locally :D

AndrewJakubowicz (approved):
NICE! LGTM!

package.json
```diff
@@ -16,8 +16,8 @@
     "test:links:redirects": "node packages/lit-dev-tools-esm/lib/check-redirects.js",
     "test:links:internal": "run-p -r start check-links:internal",
     "test:links:external": "run-p -r start check-links:external",
-    "check-links:internal": "wait-on tcp:8080 && blc http://localhost:8080 --recursive --exclude-external --ordered",
-    "check-links:external": "wait-on tcp:8080 && blc http://localhost:8080 --exclude 'http://localhost:8000/dev/' --exclude 'https://www.npmjs.com/*' --get --recursive --host-requests 1",
+    "check-links:internal": "wait-on tcp:6415 && blc http://localhost:6415 --recursive --exclude-external --ordered",
+    "check-links:external": "wait-on tcp:6415 && blc http://localhost:6415 --exclude 'http://localhost:8000/dev/' --exclude 'https://www.npmjs.com/*' --get --recursive --host-requests 1",
```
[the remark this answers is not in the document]
aomarks: Oh nice, I was actually wondering what that was :P

packages/lit-dev-tools-esm/src/dev-server.ts
aomarks: Oops, fixed.

packages/lit-dev-tools-esm/src/dev-server.ts
aomarks: Yep, that's correct. Also we *display* the file as raw text in the editor, so we also don't want to see resolved modules there. My comment actually only mentioned the display issue, but you're right we also want to control module resolution ourselves. Expanded comment.

PR #521, opened 2021-09-27
Replace current state in history instead of pushing new when saving tutorial

Matsuuu:
The history get's flooded quite fast when typing in the tutorial, and it would be beneficial to instead replacethe current history state instead of pushing a new one every time.

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-tutorial.ts
+++ b/packages/lit-dev-content/src/components/litdev-tutorial.ts
@@ -260,7 +260,7 @@ export class LitDevTutorial extends LitElement {
   };
 
   private _writeUrl() {
-    window.history.pushState(
+    window.history.replaceState(
       null,
       '',
       addModsParameterToUrlIfNeeded(this._info.url)


packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -260,7 +260,7 @@ export class LitDevTutorial extends LitElement {
   };
 
   private _writeUrl() {
-    window.history.pushState(
+    window.history.replaceState(
```
aomarks, line 263: Thanks for the PR!

I don't think this line is the problem. We want this method to use `pushState`, because we want to add a history entry when you move to the next or previous tutorial step, which is what this method does.

I think the problem is actually in playground itself here: https://github.com/google/playground-elements/blob/05538da0ee6fe13b3345d89b805dd06b2b8f30b3/src/playground-preview.ts#L198

When an iframe reloads, the parent window's history is affected. So the history states are actually from the playground preview reloading. I think the trick to prevent that would be to remove the iframe from the DOM before we change its `src`, and then attach it after.

Matsuuu: Ah that makes sense. I had some trouble getting the project to build so I thought I'd raise this PR to bring this problem to attention. 

```
When an iframe reloads, the parent window's history is affected. So the history states are actually from the playground preview reloading. I think the trick to prevent that would be to remove the iframe from the DOM before we change its src, and then attach it after.
```
If that's the case, I'll look into it after I've got the project set up

Matsuuu: I'll open up a separate PR for that as I get around to doing it. Closing this one as it has unnecessary commits

Issue #522, opened 2021-09-27
Show TOC for more resolutions


aomarks:
E.g. right now a 16" MBP on large scaling, which is equivalent to 1344x800, doesn't show the TOC, but it should fit there with a little tweaking.

We can adjust the width where the narrow media query cuts in / the min width of the main content / the padding around the columns for narrower widths.

PR #524, opened 2021-09-28
Squeeze RHS TOC into narrower horizontal resolutions

aomarks:
The RHS TOC will now display when the horizontal resolution is >1332px, where previously it required >1440px. This way we'll cover 1344x840 (the next highest scaling setting on a recent 16" MBP), and also 1366x768 (a generally common resolution).

(The exact 1332px number is a little arbitrary looking because we use a formula that makes the transitions look smooth, so this came from halving the width that we allow the TOC to squeeze to).

Fixes https://github.com/lit/lit.dev/issues/522

cc @castastrophe

### DIFF
--- a/packages/lit-dev-content/site/docs/components/overview.md
+++ b/packages/lit-dev-content/site/docs/components/overview.md
@@ -23,5 +23,3 @@ Creating a Lit component involves a number of concepts:
 Here's a sample component:
 
 {% playground-example "docs/components/overview/simple-greeting" "simple-greeting.ts" %}
-
-</div>
--- a/packages/lit-dev-content/site/css/docs.css
+++ b/packages/lit-dev-content/site/css/docs.css
@@ -542,7 +542,7 @@ td {
 }
 
 #rhsToc {
-  padding: calc(var(--docs-margin-top) + 0.3em) 1.5em 0 1em;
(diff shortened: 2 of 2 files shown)


AndrewJakubowicz (approved):
Fantastic!

Did some manual testing on the deployed site. When using the responsive dev tooling and typing in px widths everything works as expected.

However if I manually change the width of the window the RHS menu remains at a smaller width - 1319px (and I think it disappears at ~1316px). This creates a horizontal scroll bar.

Screenshot:

Screen Shot 2021-09-27 at 6 32 05 PM
 

Reproduced at: https://pr524-41243f8---lit-dev-5ftespv5na-uc.a.run.app/docs/components/styles/

Used Mac/Windows & Chrome. Opened dev tooling so that the width/height appears in the top right and manually shrunk the window. Was also able to repro this smaller width.

This may also not be a blocking issue. Great work!

packages/lit-dev-content/site/docs/components/overview.md
```diff
@@ -23,5 +23,3 @@ Creating a Lit component involves a number of concepts:
 Here's a sample component:
 
 {% playground-example "docs/components/overview/simple-greeting" "simple-greeting.ts" %}
-
-</div>
```
AndrewJakubowicz, line 27: Nice find!

AndrewJakubowicz (approved):

Issue #523, opened 2021-09-25
Error in documentation (Typescript) for inheriting-styles-from-a-superclass


davie-robertson:
https://lit.dev/docs/components/styles/#inheriting-styles-from-a-superclass

`super.styles,`

gives the following error

`'super' property access is permitted only in a constructor, member function, or member accessor of a derived class.`

AndrewJakubowicz:
Thanks for the issue, transferred to the lit.dev documentation site repository.

The preview is currently broken:

https://user-images.githubusercontent.com/15080861/134965647-cb8b9b6f-6168-45d7-9245-3be538495d6f.mp4

davie-robertson:
thanks

PR #525, opened 2021-09-28
Temporary fix to ensure styles example works

AndrewJakubowicz:
Issue: https://github.com/lit/lit.dev/issues/523

### Context

Typescript transpiles this example into invalid JavaScript!

See this [reproduction in the TS playground.](https://www.typescriptlang.org/play?ts=4.2.3#code/FAYwNghgzlAEAKEBOBTAdgF1gb2LWUGEGAliARgJ5gpQBcFSJaA5gNoC6sAvLJwNzAAvsFHhocAMIALEmAAmsFAA8M6eXESpMOPBWJkK1Wg0JNWnHnwB0tqAFcADiiTXCxqB0EigA)

Initially tried to upgrade playground-elements to 4.4.3 which is a possible fix. I ran into some issues. Thus this is a stopgap fix in the meantime, and it still makes sense in the lit.dev context.

Manually verify at this link: https://pr525-4ca1858---lit-dev-5ftespv5na-uc.a.run.app/docs/components/styles/#inheriting-styles-from-a-superclass

### Screenshot test

Added a golden for this playground preview - this should remain the same in the future when we move it back to `super` syntax:

inheritingStylesPlaygroundPreview-linux


aomarks (approved):
Nice!

PR #526, opened 2021-09-28
Implement a Content Security Policy (CSP) for lit.dev (report only for now)

aomarks:
Adds a new `contentSecurityPolicyMiddleware` which we run in both the prod and dev servers.

For now it is using the `Content-Security-Policy-Report-Only` heading. This means violations will be shown in the console, but they won't actually throw any exceptions. On the production site, violations will additionally get reported to the Google CSP collector, which has a dashboard we can use to see if violations are occuring.

The policy isn't as strict as we would like. In particular, we have to allow `unsafe-eval` because of a library we are using inside the Playground web worker. More details in the TODO. I plan to send some followup PRs that will incrementally lock down this policy.

### DIFF
--- a/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
@@ -1,0 +1,115 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import type Koa from 'koa';
+
+/**
+ * Options for contentSecurityPolicyMiddleware.
+ */
+export interface ContentSecurityPolicyMiddlewareOptions {
+  /**
+   * Enables some additional directives that are only required when running in
+   * local dev mode.
+   */
+  devMode?: boolean;
+
+  /**
+   * If true, CSP violations will be reported to the Google CSP Collector.
+   */
+  reportViolations?: boolean;
+}
+
+/**
+ * Creates a Koa middleware that sets the lit.dev Content Security Policy (CSP)
+ * headers.
+ *
+ * Some good resources about CSP:
+ * https://web.dev/strict-csp/
+ * https://www.w3.org/TR/CSP3/
+ * https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
+ * https://speakerdeck.com/lweichselbaum/csp-a-successful-mess-between-hardening-and-mitigation
+ */
+export const contentSecurityPolicyMiddleware = (
+  opts: ContentSecurityPolicyMiddlewareOptions = {}
+): Koa.Middleware => {
+  const cspHeaderValue = [
+    // TODO(aomarks) We should also enable trusted types, but that will require
+    // a policy in playground-elements for creating the worker, and a policy
+    // es-module-lexer for doing an eval (see next comment for more on that).
+
+    // TODO(aomarks) unsafe-eval is needed for an eval that is made by
+    // es-module-lexer to perform JavaScript string unescaping
+    // (https://github.com/guybedford/es-module-lexer/blob/91964da6b086dc5029091eeef481180a814ce24a/src/lexer.js#L32).
+    // There are a number of ways we could make this stricter: [1] use the
+    // asm.js build which doesn't use eval (but it's significantly slower), [2]
+    // use a separate CSP for the worker (except Chrome doesn't support that
+    // yet, though we could simulate it with an iframe), [3] get trusted types
+    // into the WASM build, [4] modify or fork the WASM build to implement
+    // string unescaping without eval (needs benchmarking).
+    //
+    // In dev mode, data: scripts are required because @web/dev-server uses them
+    // for automatic reloads.
+    `script-src 'self' 'unsafe-eval' https://www.googletagmanager.com/gtag/js ${
+      opts.devMode ? ` data:` : ''
+    }`,
+
+    // unpkg.com is needed to allow the Playground worker to fetch dependencies.
+    // TODO(aomarks) After https://crbug.com/1253267 is fixed we can serve a
+    // separate CSP policy just for the worker script (Firefox and Safari
+    // already support this).
+    //
+    // In dev mode, ws: connections are required because @web/dev-server uses
+    // them for automatic reloads.
+    `connect-src 'self' https://unpkg.com/${opts.devMode ? ` ws:` : ''}`,
+
+    // TODO(aomarks) These frame-src directives are only needed for embedding
+    // Playground previews. We can know the exact origin that is being used for
+    // Playground previews, so we could restrict this further by passing that in
+    // as a parameter to the middleware.
+    //
+    // In dev mode, http: is needed for http://localhost Playground iframes.
+    `frame-src https:${opts.devMode ? ' http:' : ''}`,
+
+    // We need 'unsafe-inline' because CodeMirror uses inline styles See
+    // https://discuss.codemirror.net/t/inline-styles-and-content-security-policy/1311/2
+    // TODO(aomarks) Is it possible to know all of the possible inline styles
+    // CodeMirror will use, and provide them here using 'unsafe-hashes'? They
+    // look quite dynamic though.
+    `style-src 'self' 'unsafe-inline'`,
+
+    // TODO(aomarks) We need fonts.gstatic.com for fetching Open Sans (Manrope
+    // is already hosted locally). We should probably just host Open Sans
+    // ourselves, because there is no cross-origin caching benefit (see
+    // https://developers.google.com/web/updates/2020/10/http-cache-partitioning),
+    // and it would be one fewer connection to make.
+    `font-src 'self' https://fonts.gstatic.com/`,
+
+    // TODO(aomarks) We use some data: URLs for SVGs in docs.css. There's
+    // probably a simpler way.
+    // The ytimg.com domain is needed for embedded YouTube videos.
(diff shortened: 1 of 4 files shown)


packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+    // into the WASM build, [4] modify or fork the WASM build to implement
+    // string unescaping without eval (needs benchmarking).
+    //
+    // In dev mode, data: scripts are required because @web/dev-server uses them
+    // for automatic reloads.
+    `script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.googletagmanager.com/gtag/js ${
+      opts.devMode ? ` data:` : ''
+    }`,
+
+    // unpkg.com is needed to allow the Playground worker to fetch dependencies.
```
AndrewJakubowicz, line 64: Does this mean that we don't support any other cdns (skypack, jsdelivr, etc.)?

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    // TODO(aomarks) This could be 'none' if we didn't use <svg><use> elements,
+    // because Firefox does not follow the img-src directive for them, so there
+    // is no other directive to use. See
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=1303364#c4 and
+    // https://github.com/w3c/webappsec-csp/issues/199.
+    `default-src 'self'`,
+
+    ...(opts.reportViolations
+      ? [`report-uri https://csp.withgoogle.com/csp/lit-dev`]
```
AndrewJakubowicz, line 108: Is this the domain where we the csp reports go? - Also clever use of spread.

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+      : []),
+  ].join('; ');
+
+  return async (ctx, next) => {
+    await next();
+    if (ctx.response.type === 'text/html') {
```
AndrewJakubowicz, line 114: Clarification: We only add this header to `text/html` responses?

This was also really helpful to manually test. Was able to see the `content-security-policy-report-only` header being applied on the test domains which is very cool.

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Options for contentSecurityPolicyMiddleware.
+ */
+export interface ContentSecurityPolicyMiddlewareOptions {
+  /**
+   * Enables some additional directives that are only required when running in
+   * local dev mode.
+   */
+  devMode?: boolean;
```
AndrewJakubowicz, line 17: I see! So directives are like a very specific allowlist. In devMode we allow `http` for example.

AndrewJakubowicz (commented):
This is great!

Going to read the other changes and come back to this since I want to understand a little more. Will review again tomorrow morning after reading the other PRs. Thank you for such rich documentation!

Naive initial question - is CSP possible to test? I.e. creating a situation that CSP should block?

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
aomarks: We always use unpkg in the Playground web worker, that's hard-coded. If you write a bare module in a Playground, it will be retrieved behind-the-scenes from unpkg.

However you can still directly import any arbitrary URL from a Playground editor -- because anything running in a Playground preview is unaffected by this CSP, because it runs on the totally separate playground.lit.dev origin.

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
aomarks: Yep! https://web.dev/strict-csp/ is a great article.

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
aomarks: Yep, this is a service that Google runs for aggregating CSP violations on Google services. It gives us some insight into whether and which CSP violations are happening, which could indicate bugs or attempted attacks.

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
aomarks: Yeah, it only needs to be set on top-level HTML responses, since CSP only applies there. (Actually it *should* also apply to worker scripts, which will be helpful for us in the future, but Chrome currently implements this incorrectly [but it is being worked on]). Setting it e.g. on an image would have no effect.

aomarks (commented):
> Going to read the other changes and come back to this since I want to understand a little more. Will review again tomorrow morning after reading the other PRs. Thank you for such rich documentation!

For sure, no rush, thanks!

aomarks:
> This is great!
> 
> Going to read the other changes and come back to this since I want to understand a little more. Will review again tomorrow morning after reading the other PRs. Thank you for such rich documentation!
> 
> Is CSP possible to test? I.e. creating a situation that CSP should block?

Once the CSP is actually turned on (removing the `-Report-Only` prefix), then the existing integration tests we have will be testing that the policy is allowing what we *want* to happen, since it won't work otherwise.

Interesting idea about testing the inverse -- confirming that the CSP is actually being enforced. I think we could use puppeteer from an integration test to e.g. insert an inline script, and we should see that fail/throw.

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
jimaek: Is there anything that we (jsDelivr) can do to get added as well? We have support for ESM and a detailed API that even provides correct endpoints for each project https://github.com/jsdelivr/data.jsdelivr.com

aomarks: Hi @jimaek!

This CSP directive won't stop somebody using the lit.dev playground to import from jsDelivr if they chose to.  In general there is nothing you can do on the playground today that you won't be able to do after this policy is enforced. Here's an example that works [today](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy5qcyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnaHR0cHM6Ly9lc20ucnVuL2xpdCc7XG5cbmV4cG9ydCBjbGFzcyBTaW1wbGVHcmVldGluZyBleHRlbmRzIExpdEVsZW1lbnQge1xuICBzdGF0aWMgc3R5bGVzID0gY3NzYHAgeyBjb2xvcjogYmx1ZSB9YDtcblxuICBzdGF0aWMgcHJvcGVydGllcyA9IHtcbiAgICBuYW1lOiB7dHlwZTogU3RyaW5nfSxcbiAgfTtcblxuICBjb25zdHJ1Y3RvcigpIHtcbiAgICBzdXBlcigpO1xuICAgIHRoaXMubmFtZSA9ICdTb21lYm9keSc7XG4gIH1cblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgPHA-SGVsbG8sICR7dGhpcy5uYW1lfSE8L3A-YDtcbiAgfVxufVxuY3VzdG9tRWxlbWVudHMuZGVmaW5lKCdzaW1wbGUtZ3JlZXRpbmcnLCBTaW1wbGVHcmVldGluZyk7XG4ifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhRE9DVFlQRSBodG1sPlxuPGhlYWQ-XG4gIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2ltcGxlLWdyZWV0aW5nLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPHNpbXBsZS1ncmVldGluZyBuYW1lPVwiV29ybGRcIj48L3NpbXBsZS1ncmVldGluZz5cbjwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgICBcImxpdFwiOiBcIl4yLjAuMFwiLFxuICAgIFwiQGxpdC9yZWFjdGl2ZS1lbGVtZW50XCI6IFwiXjEuMC4wXCIsXG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl4zLjAuMFwiLFxuICAgIFwibGl0LWh0bWxcIjogXCJeMi4wLjBcIlxuICB9XG59IiwiaGlkZGVuIjp0cnVlfV0), and it will work after this CSP policy is enforced too.

In [playground-elements](https://github.com/google/playground-elements), when you write a bare module specifier, we do our own bare module resolution, and fetch from unpkg behind the scenes. So this CSP directive is just to allow playground-elements to do that.

I do think it would be great to allow configuring which CDN is used, though. I previously filed https://github.com/google/playground-elements/issues/187 to track that. Please feel free to comment there!

AndrewJakubowicz (approved):
AMAZING! - My only concern is the integration test timeout but that's not in this change.

Issue #201, opened 2021-03-03
Consider self-hosting fonts


aomarks:
There's no longer any caching benefit to using fonts.google.com, because cache partitioning means cached assets are never shared between origins (https://developers.google.com/web/updates/2020/10/http-cache-partitioning). So, it may make sense to just host the woff files directly, so that no external connection is needed.

PR #529, opened 2021-09-29
Self host fonts

aomarks:
- Adds a tool that downloads fonts from Google Fonts. This includes font woff2 files, and the CSS files that configure them. Previously, we already self-hosted and inlined our font CSS to eliminate an extra hop (so that fonts start downloading ASAP), but this was done manually. Automating this makes it easy to bring in updates or add new fonts by running `npm run build:fonts`.

- The font files are checked into GitHub instead of being downloaded on every build so that we can control the exact version we have. If a buggy version gets pushed, or Google Fonts is down, we can just rollback. Kinda like a package lock.

- Note that we used to dynamically fetch Manrope from their GitHub repo. We did this because we were seeing some artifacting with the Google Fonts version. It doesn't look like a new version has been pushed to Google Fonts, but there was an anti-aliasing parameter that we changed which may have been the actual culprit. Once this preview is live, I'll ask the person who experienced artifacting before if they still see it now.

Reasons for doing this:
1. There's no caching advantage to using an external CDN for fonts because of https://developers.goole.com/web/updates/2020/10/http-cache-partitioning.
 2. It avoids an extra connection and allows fonts to be multiplexed over the same HTTP/2 connection as other assets.
 3. We can make our own smaller versions of the `@font-face` CSS rules. E.g. we know that we only care about the latin character set (for now). We can also then inline these into our HTML, so that fonts start downloading as soon as possible.
 4. It allows a simpler CSP policy.
 5. It's one less thing that can go down. Google Fonts probably has great uptime, but you can't beat having no external service dependency at all.

Fixes https://github.com/lit/lit.dev/issues/201


README.md
```diff
@@ -128,3 +128,15 @@ generates a zip archive `golden-results` which can be extracted into `tests/`.
 ```sh
 unzip golden-results.zip -d tests/
 ```
+
+### Fonts
+
+Fonts are served locally and copied into this Git repo from Google Fonts.
+
+To update existing font woff2 and `@font-face` CSS rules from Google Fonts, run
+`npm run build:fonts` from the `lit-dev-content` package.
```
AndrewJakubowicz, line 137: Could this be rolled into `npm run build` within `lit-dev-content`?

packages/lit-dev-tools-esm/src/update-fonts.ts
```diff
@@ -0,0 +1,192 @@
...
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// This script downloads the fonts used by lit.dev from Google Fonts so that
+// they can be hosted locally. We do this because:
+//
+// 1. There's no caching advantage to using an external CDN for fonts because of
+//    https://developers.goole.com/web/updates/2020/10/http-cache-partitioning.
```
AndrewJakubowicz, line 11: Typo: `goole` -> `google`.

packages/lit-dev-tools-esm/src/update-fonts.ts
```diff
@@ -0,0 +1,192 @@
...
+    latinOnly: true,
+    license: `
+Copyright 2019 The Manrope Project Authors (https://github.com/sharanda/manrope)
+SPDX-License-Identifier: OFL-1.1`,
+  },
+];
+
+const THIS_FILE = fileURLToPath(import.meta.url);
+const THIS_DIR = pathlib.dirname(THIS_FILE);
+const LIT_DEV_SITE_DIR = pathlib.relative(
```
AndrewJakubowicz, line 68: Nice!

packages/lit-dev-server/src/server.ts
```diff
@@ -75,7 +75,12 @@ app.use(
       // TODO(aomarks) Oddly can't access the request URL path from this API.
       // This `path` is the path on disk. Works for now, though.
       if (path.includes('/fonts/')) {
-        res.setHeader('Cache-Control', 'max-age=3600');
+        res.setHeader(
+          'Cache-Control',
+          // These are the cache headers used by fonts.googleapis.com. It can be
+          // aggressively cached because the font filenames are unique.
+          'max-age=86400, stale-while-revalidate=604800'
```
AndrewJakubowicz, line 82: TIL `stale-while-revalidate`

AndrewJakubowicz (approved):
Ran script locally and it works great!

packages/lit-dev-tools-esm/src/update-fonts.ts
aomarks: Oops thanks, fixed.

README.md
aomarks: It could be, but I actually was thinking it's preferable to isolate font updates as their own changes. Similar to the benefits of a package lock (although to be fair, fonts are waaay less likely to break something than a dependency, but it is still possible for a font to have a bug).

I'm thinking of adding a GitHub action soon which will automatically bring in the latest package lock, API docs, and fonts -- and automatically send us PRs when they change.

AndrewJakubowicz:
The screenshot comparison is failing. Maybe a slight change due to the font changes. LMK if you have any questions about getting the fresh screenshots from Github Actions.

aomarks:
> The screenshot comparison is failing. Maybe a slight change due to the font changes. LMK if you have any questions about getting the fresh screenshots from Github Actions.

Yep! Actually I'm still investigating the changes, because I confirmed with @sorvell that he is seeing some weird artifacting on his display with the Google Fonts version of Manrope. I'll be filing a bug with Google Fonts to see if we can get it fixed. So I'll probably wait on this PR for a bit.

AndrewJakubowicz:
Out of interest is the artifact visible in the screenshot from Github Actions (attached the updated screenshot for the failing test).
From eyeballing the `@` seems to be quite narrow.


helloWorldPlaygroundProject-linux
 w

packages/lit-dev-server/src/server.ts
justinfagnani: oh, it's the best! Might not be needed here since with unique file names the revalidate step should never return anything interesting, right? I guess the only downside is a unneeded revalidate request?

packages/lit-dev-tools-esm/src/update-fonts.ts
```diff
@@ -0,0 +1,192 @@
...
+    litdevFaceRules.push(litdevRule);
+  }
+
+  if (litdevFaceRules.length == 0) {
+    throw new Error(`Did not find any faces for font ${font.name}`);
+  }
+
+  // Fetch and write the font file.
+  for (const [gfFontUrl, litdevFontPath] of fontsToFetch) {
+    const fontResp = await fetch(gfFontUrl);
```
justinfagnani, line 172: can you put the loop body in an async function so the fonts can download in parallel?

justinfagnani (approved):

e111077:
In particular we are still having issues with locally served Manrope, but the big issue is that we're making the user download 200kb of fonts. Perhaps we can do something similar to what google fonts does with specific text, because we only need a small subset of extrabold and semibold

https://www.amitmerchant.com/minimizing-google-fonts-by-loading-it-for-certain-characters/

PR #530, opened 2021-09-29
Add dedicated CSP for the Playground worker

aomarks:
This will give Firefox and Safari a stricter CSP policy for the Playground worker today.

After https://crbug.com/1253267 (in progress) is fixed, the same will also apply to Chrome. At that point, we will also be able to remove the `'unsafe-eval'` and `https://unpkg.com/` sources from our main CSP, since all browsers will have an isolated CSP for the Playground worker.

### DIFF
--- a/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
@@ -34,6 +34,11 @@ export interface ContentSecurityPolicyMiddlewareOptions {
 }
 
 /**
+ * A Google service which aggregates CSP violations.
+ */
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (approved):
LGTM! Fantastic documentation!

Is the `CSP_REPORT_URI` something that needs to be setup server side, or is it an external service?

Also Github actions possibly only need to be retriggred. Error appears to be a timeout during npm install?

aomarks:
> Is the `CSP_REPORT_URI` something that needs to be setup server side, or is it an external service?

From what I understand, there is no additional setup we need to do.

> Also Github actions possibly only need to be retriggered. Error appears to be a timeout during npm install?

Wow yet another kind of failure today? We're having a flaky day.

Issue #533, opened 2021-09-30
Playground Gist integration for short URLs


aomarks:
Allow the user to signin with GitHub and save their Playgrounds as GitHub gists.

PR #534, opened 2021-09-30
Start on a fake GitHub server

aomarks:
Starts an implementation of a fake GitHub server which we'll use during dev and for automated tests.

Initially this just implements the OAuth authentication prompt. Our version looks like this:

And the real version looks like this:

See https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#1-request-a-users-github-identity for information on the flow.

Roughly how this will work is:
1. The user clicks share and opts to sign into GitHub
2. A popup opens showing the screen above.
3. If they click approve, they are redirected to `https://lit.dev/playground/signin/?code=<temporary code>`
4. The `/playground/signin/` page sends a `postMessage` back to the original Playground screen with the `code` pulled out of the URL parameter.
5. The popup closes
6. The original Playground screen sends a request to a new API running on our server, which exchanges the code plus a secret key for our OAuth app for a *token*, and sends that token back in the response.
7. The Playground now has a token that it can use to write gists on behalf of the signed in GitHub user.

First part of https://github.com/lit/lit.dev/issues/533


packages/lit-dev-content/site/playground/signin.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="robots" content="noindex" />
```
AndrewJakubowicz, line 5: Nit: no closing required `/>`. Can be `>`

packages/lit-dev-content/site/playground/signin.html
```diff
@@ -0,0 +1,13 @@
...
+  </head>
+  <body>
+      <!-- In the future, this page will communicate with its parent Playground
+           page (having been opened as a popup window) to pass back an
```
AndrewJakubowicz, line 9: Non blocking, just for my understanding: as an iframe?

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Koa middleware that simulates various GitHub APIs and authentication screens
+ * used by the lit.dev Playground.
+ */
+export const fakeGitHubMiddleware = (
+  options: FakeGitHubMiddlewareOptions
+): Koa.Middleware => {
+  let fake = new FakeGitHub(options);
+  return async (ctx, next) => {
+    if (ctx.path === '/reset') {
```
AndrewJakubowicz, line 47: Is this a github endpoint or a testing convenience? Clarifying this will be under `/playground/signin/reset`?

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
```diff
@@ -0,0 +1,133 @@
...
+  private readonly _userToCode = new Map<string, string>();
+
+  constructor(options: FakeGitHubMiddlewareOptions) {
+    this._options = options;
+  }
+
+  /**
+   * Generate a random user ID for this session and persist it in a cookie.
+   */
+  private _getOrSetUserId(ctx: Koa.Context) {
```
AndrewJakubowicz, line 70: Nit: May help to have cookie in the name here. `_getOrSetUserIdFromCookie`.

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * Simulates https://github.com/login/oauth/authorize, the screen that prompts
+   * a user to authorize a GitHub OAuth App.
+   *
+   * Documentation:
+   * https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#1-request-a-users-github-identity
+   */
+  authorize(ctx: Koa.Context) {
```
AndrewJakubowicz, line 97: Since this is a fake, is there any benefit to add extra verification/assertions such as ensuring that the `ctx` has a `scope`/`state` set for example?
Testing against a more constrained endpoint to restrict what we do.
non actionable comment.

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
```diff
@@ -0,0 +1,133 @@
...
+      ctx.status = 400;
+      ctx.body = 'error: missing or incorrect client_id url parameter';
+      return;
+    }
+
+    const userId = this._getOrSetUserId(ctx);
+    const code = randomString();
+    this._userToCode.set(userId, code);
+
+    const authorizeUrl = `${this._options.redirectUrl}?code=${code}`;
```
AndrewJakubowicz, line 108: Do we want to also support `state` for additional security?

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
```diff
@@ -0,0 +1,133 @@
...
+    ctx.type = 'text/html';
+    ctx.body = `
+      <h2>Fake GitHub authorization prompt</h2>
+
+      <button onclick="authorize()">Authorize</button>
+      <button onclick="cancel()">Cancel</button>
+
+      <script>
+        function authorize() {
+          document.cookie = 'authorized=1;path=/';
```
AndrewJakubowicz, line 124: TIL. This is where the authorized cookie is assigned. What is the `path=/`

AndrewJakubowicz (commented):
LGTM!

Quick question:
 - Is the playground.html renaming possible because we were already turning it into a `/playground` url somewhere else?
 -

AndrewJakubowicz (approved):

packages/lit-dev-content/site/playground/signin.html
aomarks: It will be a popup window. Next PR will demonstrate!

packages/lit-dev-content/site/playground/signin.html
aomarks: Done.

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
aomarks: Just a testing convenience. Added a note to the `reset` method.

It will be under `http://localhost:5417/reset` -- all of these endpoints are on a separate server, simulating the GitHub server.

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
aomarks: Done.

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
aomarks: `path` constrains the URLs on which a particular cookie will be sent to the server. `path=/` means all paths. By default it is the current path (which would be `/login/oauth` in this case).

The reason I set the path to `/` instead of leaving it as the default is so that when we call `ctx.cookies.set('authorized', null)` above, we don't have to know what the path was, since the default for that call is `/`, and it won't be deleted if the paths don't match. (So we could also do `ctx.cookies.set('authorized', null, {path:'/login/oauth'})`, but the other approach is simpler).

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
aomarks: Good question. I've been trying to think of whether we can use this for meaningful extra security, and I'm not sure yet.

Let's say our server generated a random string to be put in the `state` parameter, and then checked that `state` parameter when the response eventually comes back. This would seem to guard against attackers sending victims to an authorization URL outside of our login flow. But what would prevent an attacker from calling our API that generates random `states` themselves, and just including that?

One related key thing to know is that it shouldn't be possible to intercept a code by changing the `redirect_uri` parameter to something we don't control, because GitHub always checks that parameter against a URL hard coded into our App settings (it has to match it, or be a subdirectory of it: https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#redirect-urls).

aomarks:
> LGTM!
> 
> Quick question:
> 
> * Is the playground.html renaming possible because we were already turning it into a `/playground` url somewhere else?

When the Koa static middleware sees request path `/foo/bar/`, it checks on disk for both `<roo>/foo/bar.html` and `<root>/foo/bar/index.html`. So in this case `playground.html` and `playground/index.html` are equivalent in terms of their URL paths, and by creating the subdirectory we can now also serve `/playground/signin`

packages/lit-dev-server/src/middleware/fake-github-middleware.ts
aomarks: Seems reasonable. Added a check for `scope=gist`. Nothing for `state` for now, see other comment.

Issue #475, opened 2021-09-03
Docs: Inheriting styles from a superclass doesn't show an es5 compatible option


AndrewJakubowicz:
### Context

In the section [Inheriting styles from a superclass](https://lit.dev/docs/components/styles/#inheriting-styles-from-a-superclass) we show an example of using the super classes style with `super.styles` syntax.

```typescript
// Minimal example
class SuperClass {
  static get styles() {
    return ['style'];
  }
}

class ChildClass extends SuperClass {
  static get styles() {
    // On ES5, error here: `Only public and protected methods of the base class are accessible via the 'super' keyword.`
    return [...super.styles, 'additional style'];
  }
}
```

[...]

Issue #414, opened 2021-07-19
Document octal escape issue


justinfagnani:
JavaScript deprecated octal escape sequences in strict mode, but CSS's escape syntax matches JS's deprecated octal escape syntax, so CSS inside of JavaScript strings runs the risk of using illegal syntax in JS:

```ts
static styles = css`
  .foo::before {
    content: '\2022';
  }
`;
```

Here, the `\2022` needs to be written as `\\2022` because the string above is in JavaScript. `\\2022` is not an octal escape, but has an escaped `\` character in it.

Normally a user who writes `\2022` would get a syntax error in JavaScript or their favorite IDE or compiler, but complicating things is that octal escape sequences don't actually cause a syntax error in tagged template literals because the raw character sequence is preserved to give the tag access to the raw string. Instead the "cooked" string is passed to the tag as `undefined`, which breaks lit-html. This is a rare case in a hot code path and we don't want to have to check for `undefined` on strings that should be defined for all valid templates. We will add one for dev mode, but se still need documentation on this corner of valid template syntax.

https://github.com/lit/lit/issues/1093
https://github.com/lit/lit/issues/2013
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_octal

PR #535, opened 2021-09-30
Two small fixes for the styles doc

arthurevans:
- Added a note about super.styles (#475). Although #525 makes the sample work correctly, there's no mention of `super.styles` (which a reasonable person might substitute for `SuperClass.styles`) or of the specific TS issue. Not sure whether we want any more detail here (pointer to the TS bug; mention that `super.styles` should work with Babel or if you're not targeting es5... which I think is true but haven't tested).

- Added a section on octal escape sequences in `css` literals (#414).  Not sure here whether we have a preference for the double escape (`\\2022`) or JS escape (`\u2022`). I chose the latter in the example, but perhaps it doesn't matter much.


packages/lit-dev-content/site/docs/components/styles.md
```diff
@@ -74,6 +75,12 @@ Using an array of tagged template literals, a component can inherit the styles f
 
 {% playground-ide "docs/components/style/superstyles" %}
 
+<div class="alert alert-info">
+
+**Using super.styles**. You can also use `super.styles` to reference the superclass's styles property, but TypeScript doesn't convert this correctly when compiling for ES5. Explicitly referencing the superclass, as shown in the example, avoids this issue.
```
aomarks, line 80: Maybe clarify that this does work for ES2015 and above, though. Also I think we should show the `super.styles` approach in the examples. (@AndrewJakubowicz found that it works fine in TypeScript 4.4, we just need to update the Playground's TypeScript version).

arthurevans: Oh, cool. Yeah, I agree that showing the `super.styles` approach as the default is preferable. Is there a bug for updating the Playground typescript version?

aomarks: https://github.com/google/playground-elements/pull/217 to release with the new version. After this I'll send a PR to update lit.dev. Should be able to get that in before this PR merges.

aomarks: https://github.com/lit/lit.dev/pull/537

aomarks: Huh, bumping the TypeScript version fixed the type error, so there's no longer a squiggly when doing `super.styles`, but it didn't fix the output, which is still invalid note the `(void 0).styles`:

```ts
var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
    var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
    else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
    return c > 3 && r && Object.defineProperty(target, key, r), r;
};
import { css } from './node_modules/lit@2.0.0/index.js';
import { customElement } from './node_modules/lit@2.0.0/decorators.js';
import { SuperElement } from './super-element.js';
let MyElement = class MyElement extends SuperElement {
};
MyElement.styles = [
    (void 0).styles,
    css `div {
      color: red;
    }`
];
MyElement = __decorate([
    customElement('my-element')
], MyElement);
export { MyElement };
```

I wonder if we need to tweak the compiler options (https://github.com/google/playground-elements/blob/05538da0ee6fe13b3345d89b805dd06b2b8f30b3/src/typescript-worker/language-service-context.ts#L9)

justinfagnani: I still see an error in 4.4 and 4.5 when targeting ES5, but the ES2015 output does look correct. Oddly, I'm not seeing the `(void 0)` emit that you are @aomarks.

https://www.typescriptlang.org/play?target=1&ts=4.4.3#code/MYGwhgzhAECC0G8BQ1oQC5nQS2G9AniAKYwC80A5GJQNxIC+SSokMAQtMQB7rEB2AExjxkqDFlz4ipaBQgBXAA7EATgDoMMmAGoqAIzqMgA

I guess showing `super.styles` by default is ok... I just worry that code like that might not be able to be down-leveled by Babel in the case of an app that needs to target ES5 and is using 3rd party dependencies.

packages/lit-dev-content/site/docs/components/styles.md
```diff
@@ -105,6 +112,22 @@ class MyElement extends LitElement {
...
 ```
 
+### Using unicode escapes in styles
+
+CSS's unicode escape sequence is a backslash followed by four or six hex digits: for example, `\2022` for a bullet character. This similar to the format of JavaScript's deprecated _octal_ escape sequences, so using these sequences in a `css` tagged template literal causes an error.
+
+There are two work-arounds for adding a unicode escape to your styles:
+
+*   Add a second backslash (for example, `\\2022`).
+*   Use the JavaScript escape sequence, starting with `\u` (for example, `\u2022`).
```
justinfagnani, line 122: The third is to just include the unicode character in the JS source.

justinfagnani (approved):

packages/lit-dev-content/site/docs/components/styles.md
aomarks: @justinfagnani Interesting. It only happens when there is a decorator on the subclass:

https://www.typescriptlang.org/play?emitDecoratorMetadata=false&target=7&jsx=0#code/MYGwhgzhAECC0G8BQ1oQC5nQS2G9AniAKYwC80A5GJQNxIC+SSAAsAK4YD2AtgKIkexAHboAFAEokoSDABC0YgA90IgCYx4yVBiy58RUtAoR2AB2IAnAHQZDMANRUARnUZA

justinfagnani (approved):
🎉

packages/lit-dev-content/site/docs/components/styles.md
aomarks: Did we get a resolution for this?

aomarks: Oh I see, the latest update seems fine (though maybe we should note it's also broken for ES2015+ with a class decorator, not just ES5?) I think we should file a typescript bug?

arthurevans: I think under those circumstances we should just recommend against using it in TypeScript. If someone uses it and subsequently decides to add an IE11 build or to adopt decorators, their code will break and that seems bad.

arthurevans: ... especially since we invariably show TypeScript with decorators, so the only circumstance where it works is off the well-lit path.

AndrewJakubowicz: Made a [repro](https://www.typescriptlang.org/play?target=99#code/KYDwDg9gTgLgBAbwL4G4BQaD0Aqba7ZwAqAFgJYDOcUwYUEAJgK4DGwVAhnAEZMDmcAO5kYJOBRgcYZFuJgBPADbs4HAHYM4DYC2hToFAHT5CpYHDJqWiphTIQ1cCADNVWnXpjQ4LDrZWi5gAGFExgwFCGEkrsQXBePOaWulA0LDCK8nDAALYiMMCaHFRBABQAbhBkmgAMAJRBxgSYaM5MVtIO2SAcOWDKpelQAFzx8uEucADKYREAworFFHWINDBMUI5DKEgY1kvTs1ALBwj4clIycjFUALxwANoA5NHKTwC66Lto+xRUM+EoFMFMoAOrQADWVFABQ0-yOJz+iHOEkusleKnuD0MONCgKiIPYnzgmEwcDm0DS8Fy+UKPkYwFGQQBEWBMXBUChBJucCxOMMACVgM5lOlDHxgDBSgB9bgAGjgACIMRRFQrpRw6p8gmhvgABUC9frAH6LJEAIXoEOAjhhNoY8MBiKoZzgF2k6MJd0e-LxEW5ygoxNJcAAklZKTpqXkYAVNLptEzLRBrWoA5ifTiKlVanV00GUDrdkA). Can look for an open issue in TypeScript or open a new one.

AndrewJakubowicz: Another maybe interesting note, if you check `useDefineForClassFields` then the (void 0) bug no longer occurs.

PR #536, opened 2021-09-30
Start proving out GitHub authentication integration

aomarks:
- Adds a temporary unstyled "Share with GitHub" button to the Playground page behind the `?mods=gists` URL parameter. In dev, this is hooked up to the local fake GitHub server. On prod and in staging URLs, it won't do anything at all for now.

    

- Clicking the button performs the first half of the GitHub authentication process described at https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#1-request-a-users-github-identity. Just shows the code or error that was received to demonstrate that it's working, since the second half (token exchange) isn't implemented yet.

    

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-github-share-button.ts
+++ b/packages/lit-dev-content/src/components/litdev-github-share-button.ts
@@ -40,6 +40,7 @@ export class LitDevGitHubShareButton extends LitElement {
     if (!this.clientId || !this.authorizeUrl) {
       throw new Error('Missing required properties');
     }
+    // TODO(aomarks) Prevent multiple signups from being queued.
     const result = await signInToGithub({
       clientId: this.clientId,
       authorizeUrl: this.authorizeUrl,
--- a/packages/lit-dev-content/src/components/litdev-github-share-button.ts
+++ b/packages/lit-dev-content/src/components/litdev-github-share-button.ts
@@ -1,0 +1,59 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {LitElement, html} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
+import {signInToGithub} from '../github/github-signin.js';
+
+/**
+ * A button that shares a Playground project as a GitHub gist. If the user isn't
+ * signed into GitHub yet, they are prompted to do so.
+ *
+ * // TODO(aomarks) Implement actually sharing. This is just an early WIP.
+ * // TODO(aomarks) Show a scrim and some indication about what is happening
+ * //               while the popup is open.
+ * // TODO(aomarks) Style this button
+ */
+@customElement('litdev-github-share-button')
+export class LitDevGitHubShareButton extends LitElement {
+  /**
+   * GitHub OAuth App client ID. Generated when a GitHub OAuth App is first
+   * created.
+   */
+  @property()
+  clientId?: string;
+
+  /**
+   * URL where users will be redirected to authorize with GitHub.
+   */
+  @property()
+  authorizeUrl?: string;
+
+  override render() {
+    return html`<button @click=${this._onClick}>Share with GitHub</button>`;
+  }
+
+  private async _onClick() {
+    if (!this.clientId || !this.authorizeUrl) {
+      return;
+    }
+    const result = await signInToGithub({
+      clientId: this.clientId,
+      authorizeUrl: this.authorizeUrl,
+    });
+    if (result.error) {
+      alert(`error: ${result.error}`);
+    } else {
+      alert(`code: ${result.code}`);
+    }
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'litdev-github-share-button': LitDevGitHubShareButton;
+  }
+}
--- a/packages/lit-dev-content/src/components/litdev-github-share-button.ts
+++ b/packages/lit-dev-content/src/components/litdev-github-share-button.ts
@@ -38,7 +38,7 @@ export class LitDevGitHubShareButton extends LitElement {
 
   private async _onClick() {
     if (!this.clientId || !this.authorizeUrl) {
-      return;
+      throw new Error('Missing required properties');
     }
     const result = await signInToGithub({
       clientId: this.clientId,
--- a/packages/lit-dev-content/src/github/github-signin-receiver-page.ts
+++ b/packages/lit-dev-content/src/github/github-signin-receiver-page.ts
@@ -1,0 +1,37 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import type {GitHubSigninReceiverMessage} from './github-types.js';
+
+// This is the inlined script for the /playground/signin/ page.
+//
+// This page is assumed to be running in a popup window created by the main
+// Playground page. After the user clicks "Accept" or "Cancel" on the GitHub
+// authorization page, GitHub will redirect to this page with a "code" or
+// "error" URL parameter. This page will then message the original parent
+// Playground window with that code or error, and the popup window will close.
+
+const opener = window.opener as Window | null;
+if (opener) {
+  const url = new URL(document.location.href);
+  const params = url.searchParams;
+  const code = params.get('code');
+  const error = params.get('error');
+  let message: GitHubSigninReceiverMessage;
+  if (error) {
+    message = {error};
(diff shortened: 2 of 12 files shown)


packages/lit-dev-content/src/components/litdev-github-share-button.ts
```diff
@@ -0,0 +1,59 @@
...
+import {LitElement, html} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
+import {signInToGithub} from '../github/github-signin.js';
+
+/**
+ * A button that shares a Playground project as a GitHub gist. If the user isn't
+ * signed into GitHub yet, they are prompted to do so.
+ *
+ * // TODO(aomarks) Implement actually sharing. This is just an early WIP.
+ * // TODO(aomarks) Show a scrim and some indication about what is happening
```
AndrewJakubowicz, line 16: Clarification: Not familiar with the term scrim? Is this an indication on the parent page that it's awaiting something? Or is this to prevent someone clicking on the button over and over without addressing the widow.

Also we may need to expose a way for someone to "log out". Currently the sign up is very sticky, so you wouldn't be able to change accounts.

packages/lit-dev-content/src/components/litdev-github-share-button.ts
```diff
@@ -0,0 +1,59 @@
...
+  @property()
+  authorizeUrl?: string;
+
+  override render() {
+    return html`<button @click=${this._onClick}>Share with GitHub</button>`;
+  }
+
+  private async _onClick() {
+    if (!this.clientId || !this.authorizeUrl) {
+      return;
```
AndrewJakubowicz, line 41: This case feels like it's a setup/dev error. Should this be a louder error?

packages/lit-dev-content/src/github/github-signin-receiver-page.ts
```diff
@@ -0,0 +1,37 @@
...
+  const error = params.get('error');
+  let message: GitHubSigninReceiverMessage;
+  if (error) {
+    message = {error};
+  } else if (!code) {
+    message = {error: 'No code'};
+  } else {
+    message = {code};
+  }
+  opener.postMessage(message);
```
AndrewJakubowicz, line 31: Is there also a way to notify the `opener` if the page is closed? Since this is a mock it may not be relevant.

packages/lit-dev-content/src/github/github-signin.ts
```diff
@@ -0,0 +1,65 @@
...
+  options: GitHubSigninOptions
+): Promise<GitHubSigninReceiverMessage> => {
+  const url = new URL(options.authorizeUrl);
+  url.searchParams.set('client_id', options.clientId);
+  url.searchParams.set('scope', 'gist');
+  url.searchParams.set(
+    'redirect_uri',
+    new URL('/playground/signin/', document.location.href).href
+  );
+  // Width and height found by empirically.
```
AndrewJakubowicz, line 40: Are you using `empirically` as "by observation" here? I love it. TIL

packages/lit-dev-content/src/github/github-signin.ts
```diff
@@ -0,0 +1,65 @@
...
+    throw new Error('Could not open window. Are popups disabled?');
+  }
+  const result = await new Promise<GitHubSigninReceiverMessage>((resolve) => {
+    const listener = (event: MessageEvent<GitHubSigninReceiverMessage>) => {
+      if (event.source === popup) {
+        resolve(event.data);
+        window.removeEventListener('message', listener);
+      }
+    };
+    window.addEventListener('message', listener);
```
AndrewJakubowicz, line 61: Nit: `options.once`? It also probably doesn't matter - but I don't think this will clean up if the user closes the popup without one of the buttons.

packages/lit-dev-content/src/github/github-types.d.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export type GitHubSigninReceiverMessage =
+  | {code: string; error?: undefined}
+  | {code?: undefined; error: string};
```
AndrewJakubowicz, line 9: Would you ever have both `code` and `error`?

```typescript
export type GitHubSigninReceiverMessage = {code: string} | {error: string};
```

AndrewJakubowicz (approved):
Nice! Manually tested on chrome and safari. Some non blocking comments. Nice use of feature flag!

packages/lit-dev-content/src/github/github-types.d.ts
aomarks: Nope, can't have both. But turns out it works better to include the `error?: undefined` cases in the type because otherwise when we check for the `error` property, TypeScript will say "Hey, this is a union where one of the things doesn't have an error property, that's not safe!".

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3lAXPA3sqHpjAJYoDmAvvAD4EgwxIwkblXUDcAUGQDN4ACgCEUAHRMWMAJQFe8RKjRIIICRCSVhkoiFl9qvXqEiwEydFgBGORQX3tOlADTxprAPx4ArilABChBgWjoHQiRQH3h-QODgd082eFIKGj5BERspZlZ5fAcrVXVNbWEc-UNeaiA

(Also TIL you can comment on multiple lines!)

packages/lit-dev-content/src/github/github-signin.ts
aomarks: We can't actually use `once`, because that would remove the event listener as soon as *any* `message` event was fired on window. We can't assume that our popup is the only thing sending post messages, so that's why we have to check `event.source === popup`. Added a comment about this.

packages/lit-dev-content/src/components/litdev-github-share-button.ts
aomarks: > Clarification: Not familiar with the term scrim? Is this an indication on the parent page that it's awaiting something? Or is this to prevent someone clicking on the button over and over without addressing the widow.

A scrim is an overlay effect to help direct the user's attention to e.g. a dialog or in this case a popup, usually by darkening or blurring the underlying content. See the darkening effect when you click on a button here: https://material-components.github.io/material-web/demos/dialog/

> Also we may need to expose a way for someone to "log out". Currently the sign up is very sticky, so you wouldn't be able to change accounts.

True, for this demo you can simulate it by going to `https://localhost:5417/reset` but in the real UI we should have a logout button.

aomarks (commented):
Address feedback and made a bunch of changes. PTAL!

packages/lit-dev-content/src/github/github-signin.ts
AndrewJakubowicz: Ah of course! Makes sense!

packages/lit-dev-content/src/github/github-signin-receiver-page.ts
AndrewJakubowicz: Nice! Also thanks for clarifying how to popup will evolve!

AndrewJakubowicz (approved):
Very nice! Only nit I have left is that you can open multiple popups and then queue up the results. Can be follow up!

packages/lit-dev-content/src/github/github-signin.ts
```diff
@@ -49,17 +50,72 @@ export const signInToGithub = async (
...
+ * GitHub authentication code or an error. Also returns an abort function which
+ * can be used to clean up this event listener if we need to end early.
+ */
+const receiveCodeFromPopup = (
+  popup: Window
+): {
+  promise: Promise<GitHubSigninReceiverMessage>;
+  abort: () => void;
+} => {
+  const abortController = new AbortController();
```
AndrewJakubowicz, line 86: TIL!

packages/lit-dev-content/src/github/github-signin.ts
```diff
@@ -49,17 +50,72 @@ export const signInToGithub = async (
...
+      'message',
+      (event) => {
+        // Note we must check source because our popup might not be the only
+        // source of "message" events. This is also why we can't set "once".
+        if (event.source === popup) {
+          resolve(event.data);
+          abort();
+        }
+      },
+      {signal: abortController.signal}
```
AndrewJakubowicz, line 99: wow TIL again!

aomarks: > Very nice! Only nit I have left is that you can open multiple popups and then queue up the results. Can be follow up!

Ah yes, good point. Added a TODO!

Issue #531, opened 2021-09-30
Improve CSP handling for google analytics script


aomarks:
We currently have a hard-coded sha256 hash to allow the inline google analytics script to run.

If we ever updated that script, or updated our Google Analytics ID, this could easily get out of sync.

We should automate generation of this hash just like we do for other inline scripts. The reason we can't do this with the exact same method is that the Google Analytics ID needs to be baked into the source.

It would also be a good idea to register another Google Analytics account to use for dev and testing, because the main reason I missed this in my initial set of CSP PRs was that the Google Analytics script is only present at all in the main production build. (The reason for registering another account would be to avoid polluting our main account with dev/testing page loads).

PR #540, opened 2021-10-01
More CSP fixes

aomarks:
- Handle 304 cases. The reason we were getting a lot of CSP errors since https://github.com/lit/lit.dev/pull/532 was that we weren't accounting for 304 Not Modified responses. In those cases, no `Content-Type` header is set, because the browser will already know it from the most recent 200 response. Our middleware incorrectly assumed it could check `type` after awaiting the downstream Koa static middleware. We now instead set the policy based on the path of the request (anything ending in `/`), since that will work for both 200 and 304 responses.

   There's an interesting question about whether we *should* set a CSP at all for a 304 response. There's some discussion about that [here](https://github.com/w3c/webappsec-csp/issues/161) and [here](https://bugs.chromium.org/p/chromium/issues/detail?id=174301). Chrome *does* update the CSP if it's included in a 304 response, otherwise it ignores it in favor of the CSP from the last 200 response (same with all headers, except for the ones enumerated [here](https://chromium.googlesource.com/chromium/src/net/+/refs/heads/main/http/http_response_headers.cc#81)).

    The advantage of including it seems to be that if we update the CSP, but the content (and hence ETag) of a file has *not* changed, then the browser would otherwise continue using the stale CSP. OTOH, if you are using a nonce based script allowlist (we use hashes instead currently), then this wouldn't really work unless you did something clever to make the ETag aware of some CSP version.

- Fix policy for Google Analytics. We covered the origin needed for the initial script, but then *that* script goes on to require a connection to a different origin.

- Hook up a new Google Analytics test ID that I just created, so that we will catch CSP reports related to Google Analytics during dev and in PR builds going forward.

Part of https://github.com/lit/lit.dev/issues/531

### DIFF
--- a/cloudbuild-pr.yaml
+++ b/cloudbuild-pr.yaml
@@ -21,6 +21,8 @@ steps:
       - --cache-ttl=168h # 1 week
       # Bake in this revision's corresponding playground sandbox url
       - --build-arg=PLAYGROUND_SANDBOX=https://pr$_PR_NUMBER-$SHORT_SHA---lit-dev-playground-5ftespv5na-uc.a.run.app/
+      # Testing Google Analytics ID
+      - --build-arg=GOOGLE_ANALYTICS_ID=G-PPMSZR9W18
 
   # Create a new Cloud Run revision for the main site.
   #
--- a/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
@@ -80,7 +80,7 @@ export const contentSecurityPolicyMiddleware = (
       // TODO(aomarks) Remove unsafe-eval when https://crbug.com/1253267 is fixed.
       // See comment below about playgroundWorkerCsp.
       `'unsafe-eval'`,
-      `https://www.googletagmanager.com/gtag/js`,
+      `https://www.googletagmanager.com/`,
       GOOGLE_ANALYTICS_INLINE_SCRIPT_HASH,
       ...(opts.inlineScriptHashes?.map((hash) => `'${hash}'`) ?? []),
       // In dev mode, data: scripts are required because @web/dev-server uses them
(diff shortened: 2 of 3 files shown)


cloudbuild-pr.yaml
```diff
@@ -21,6 +21,8 @@ steps:
       - --cache-ttl=168h # 1 week
       # Bake in this revision's corresponding playground sandbox url
       - --build-arg=PLAYGROUND_SANDBOX=https://pr$_PR_NUMBER-$SHORT_SHA---lit-dev-playground-5ftespv5na-uc.a.run.app/
+      # Testing Google Analytics ID
+      - --build-arg=GOOGLE_ANALYTICS_ID=G-PPMSZR9W18
```
AndrewJakubowicz, line 25: So I understand correctly, this new test based google analytics will make sure if there is a csp issue we'll see it loudly in dev 🤞 ?

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -176,11 +183,19 @@ export const contentSecurityPolicyMiddleware = (
...
+    // because for a 304 Not Modified response, the Content-Type header will not
+    // be set.
+    //
+    // Also note that we don't necessarily need to set a CSP on 304 responses at
+    // all, because the browser will use the one from the previous 200 response
+    // if missing (as it does for all headers). However, by including a CSP on
+    // 304 responses, we cover the case where the CSP has changed, but the
+    // file's content (and hence ETag) has not. Note this approach would not
+    // work if we were using nonces instead of hashes.
+    if (ctx.path.endsWith('/')) {
```
AndrewJakubowicz, line 197: When manually testing and navigating to something like: `/playground`, I notice a 301 redirect to `/playground/`. Non blocking but curious about the guarantee of a trailing slash?

For example I can still navigate to: https://pr540-ed90bcb---lit-dev-5ftespv5na-uc.a.run.app/playground/index.html

cloudbuild-pr.yaml
aomarks: Yep!

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
aomarks: Yep we redirect all paths like `/foo` to `/foo/`. But nice catch that you can access the `.html` filenames too! We should block those (https://github.com/lit/lit.dev/blob/48903852d4ae835cf4d9752b0740c7c8fe8c2811/packages/lit-dev-server/src/middleware/redirect-middleware.ts#L17)

AndrewJakubowicz (approved):

PR #543, opened 2021-10-02
Bump playground version to bring in performance improvement

aomarks:
Brings in https://github.com/lit/lit.dev

PR #549, opened 2021-10-07
GitHub gists token exchange middleware and fake

aomarks:
Adds `github-token-exchange-middleware` which implements step 2 of https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#2-users-are-redirected-back-to-your-site-by-github.

Adds a fake version of the `https://github.com/login/oauth/access_token` API needed for token exchange.

### DIFF
--- a/packages/lit-dev-tools-esm/src/dev-server.ts
+++ b/packages/lit-dev-tools-esm/src/dev-server.ts
@@ -74,6 +74,8 @@ const removeWatchScriptFromPlaygroundFiles: DevServerPlugin = {
 };
 
 const clientId = 'FAKE_CLIENT_ID';
+// Important: We should never put real GitHub app secrets here. This one is just
+// for local testing with the fake server.
 const clientSecret = 'FAKE_APP_SECRET';
 
 startDevServer({
--- a/packages/lit-dev-tools-esm/src/dev-server.ts
+++ b/packages/lit-dev-tools-esm/src/dev-server.ts
@@ -11,6 +11,7 @@ import {redirectMiddleware} from 'lit-dev-server/lib/middleware/redirect-middlew
 import {playgroundMiddleware} from 'lit-dev-server/lib/middleware/playground-middleware.js';
 import {contentSecurityPolicyMiddleware} from 'lit-dev-server/lib/middleware/content-security-policy-middleware.js';
 import {fakeGitHubMiddleware} from 'lit-dev-server/lib/middleware/fake-github-middleware.js';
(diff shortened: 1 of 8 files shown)


packages/lit-dev-tools-esm/src/dev-server.ts
```diff
@@ -72,6 +73,9 @@ const removeWatchScriptFromPlaygroundFiles: DevServerPlugin = {
   },
 };
 
+const clientId = 'FAKE_CLIENT_ID';
+const clientSecret = 'FAKE_APP_SECRET';
```
AndrewJakubowicz, line 77: Could we add a comment here as well that this should never be replaced with the real secret.

AndrewJakubowicz (approved):
This looks good to me! How was this tested?

Getting an error when I try the `mods=gists` feature flag manual test (on preview url) although seems expected. Locally it seems to work with fake backend, although getting a `"Error while transforming _dev/lion-combobox.js"`. Seems unrelated.

Is it worth adding some server tests around the middleware or in a follow up?

AndrewJakubowicz (commented):
Also I really like the public/private error logging in the server. Do we want to make the public errors even more generic - the same?

aomarks (commented):
> This looks good to me! How was this tested?

Just manually. I'll be be writing a playwright test once there is some more UI to test.

> Getting an error when I try the `mods=gists` feature flag manual test (on preview url) although seems expected. Locally it seems to work with fake backend, although getting a `"Error while transforming _dev/lion-combobox.js"`. Seems unrelated.

Yeah unrelated to this PR, it's from https://github.com/lit/lit.dev/pull/543 where I bumped all the package locks. It's due to https://github.com/guybedford/es-module-lexer/issues/92 which I filed this morning. This is a dependency of @web/dev-server (and playground, incidentally), and its latest version fails on a (valid but somewhat unusual) comment-inside-import syntax used by lion-combobox. The solution is just to install the previous version of es-module-lexer, which I have included in the latest commit.

> Is it worth adding some server tests around the middleware or in a follow up?

I'm not sure we'll need more than the integration tests, since that should exercise things well, but we can probably better evaluate that once we have the integration tests and can think about what it might miss.

> Also I really like the public/private error logging in the server. Do we want to make the public errors even more generic - the same?

I think we should include useful information in the public logs so that we have some idea of where the failure is happening if we notice a problem or get a report. I just didn't want to include the raw output from the GitHub API.

Issue #550, opened 2021-10-07
Remove unsafe-eval from CSP when Chrome supports worker CSP


aomarks:
Once https://crbug.com/1253267 is fixed (estimated Chrome M97) and that Chrome release rolls out widely, we can remove `unsafe-eval` from our main HTML CSP, and keep it enabled only for the worker script. This should substantially improve the policy security, since only the Playground worker will be able to execute WASM or eval JS, but it won't be able to do hardly anything else (such as calling the new GitHub token exchange API).

PR #551, opened 2021-10-07
Enforce CSP

aomarks:
Switches our Content Security Policy from report-only mode to enforced mode.

According to our internal dashboard, it looks like CSP violation numbers dropped very sharply on October 1, which is the day https://github.com/lit/lit.dev/pull/540 landed. There do seem to be a few reports coming in as recently as October 5, but if so it is a very small number. Could be due to caching? Browser extensions injecting scripts/images etc. will also cause ongoing CSP violations, that's expected behavior.

Fixes https://github.com/lit/lit.dev/issues/517

Filed https://github.com/lit/lit.dev/issues/550 to track the most important improvement, which we can't do until https://bugs.chromium.org/p/chromium/issues/detail?id=1253267 is fixed.

### DIFF
--- a/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
@@ -201,9 +201,7 @@ export const contentSecurityPolicyMiddleware = (
     } else {
       policy = strictFallbackCsp;
     }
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz:
Why did the second build fail to deploy a preview site?

PR #554, opened 2021-10-08
Fix broken mobile nav

aomarks:
Turns out there was a CSP error that I missed that only affected mobile view. An inline `onclick` handler on the mobile nav open button.

This is a quick hack to roll forward that turns the onclick handler into an inlined script, which gets automatically hashed for CSP. Better solution would probably be to make a custom element that contains both the button and the drawer, and hook things up with a lit template.

### DIFF
--- a/packages/lit-dev-content/src/global/mobile-drawer.ts
+++ b/packages/lit-dev-content/src/global/mobile-drawer.ts
@@ -1,0 +1,16 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// TODO(aomarks) This is pretty hacky. There is currently a loose coupling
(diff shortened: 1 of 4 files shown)


packages/lit-dev-content/src/global/mobile-drawer.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+// TODO(aomarks) This is pretty hacky. There is currently a loose coupling
+// between the mobile menu button and the mobile menu itself (one is defined in
+// default.html, the other in header.html). They could be one custom element
+// instead.
```
AndrewJakubowicz, line 10: Is it worth mentioning this is for CSP? Otherwise it looks great!

AndrewJakubowicz (approved):
Nice find! Tested on my phone,

PR #556, opened 2021-10-10
Refactor icon buttons into common component

aomarks:
- Creates a new `<litdev-icon-button>` component.
- Updates the new Playground share button and the existing Tutorial solve & reset buttons to use it.
- Creates a new `<litdev-playground-download-button>` that uses the new shared button, and moves the tarball functionality out of the main `playground.ts` file into this component.
- Renames `<litdev-github-share-button>` to `<litdev-playground-share-button>`. I'm intending to make this component support both styles of sharing.
- Adds a screenshot test for the Tutorial page, to confirm the styles are the same after the refactor.


AndrewJakubowicz:
Tested functionality and it's good! Found a minor visual issue with the padding on Safari:

https://user-images.githubusercontent.com/15080861/136869001-91c44510-d813-4525-80d3-e10d05c269ff.mov

Also not sure if can tab select these buttons (although not sure if they can be tab selected in production currently).

aomarks:
> There is still a minor change in button width/height between Safari prod & this change. Can be seen when flicking between playgrounds. It's not as critical as before.

Looks like the default user agent stylesheet adds a small margin to `<button>` in Safari, but not Chrome. I think it's consistent now.

AndrewJakubowicz (approved):
LGTM! Great subtle fixes to support Safari!

PR #560, opened 2021-10-13
Miscellaneous repo cleanup

aomarks:
- Move tests into new `lit-dev-tests` package because:
  - Seems a bit more logical. Everything else lives in a package.
  - It allows us to skip installing test-related dependencies on Docker build.
- Re-organize GitHub action workflows. Most stuff was crammed into the one build action before.
- Run `playground-plugin` unit tests in CI (under new `test:unit` grouping).
- Combines golden screenshot generation + artifact upload into the main `integration-tests` action, but only run it if the tests fail.
- Use playwright's built-in `webServer` feature instead of `run-p` to launch the server used for integration testing.
- Simplify Prettier setup and check formatting in CI.
- Rename `lit-dev-tools` to `lit-dev-tools-cjs` to distinguish it from `lit-dev-tools-esm`.
- Only run the `manrope` font download step if it hasn't already run (it was needlessly running every time we installed anything or bootstrapped through lerna)
- Factor out a common `tsconfig.base.json`.
- Update dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ WORKDIR /usr/src/app
 # to execute the final Eleventy build step.
 
 # External dependencies
-COPY package*.json lerna.json ./
+COPY package*.json lerna.json tsconfig.base.json ./
 COPY packages/lit-dev-tools-cjs/package*.json ./packages/lit-dev-tools-cjs/
 COPY packages/lit-dev-tools-esm/package*.json ./packages/lit-dev-tools-esm/
 COPY packages/lit-dev-server/package*.json ./packages/lit-dev-server/


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+      - run: npm run bootstrap
+      - run: npm run build
+
+      - id: test
+        run: npm run test:integration
+
+      - if: ${{ always() && steps.test.outcome == 'failure' }}
+        run: npm run test:integration:update-golden-screenshots
+
+      - if: ${{ always() && steps.test.outcome == 'failure' }}
```
AndrewJakubowicz, line 27: Brilliant!

package.json
```diff
@@ -2,37 +2,29 @@
...
-    "test:links:internal": "run-p -r start check-links:internal",
-    "test:links:external": "run-p -r start check-links:external",
-    "check-links:internal": "wait-on tcp:6415 && blc http://localhost:6415 --recursive --exclude-external --ordered",
-    "check-links:external": "wait-on tcp:6415 && blc http://localhost:6415 --exclude 'http://localhost:8000/dev/' --exclude 'https://www.npmjs.com/*' --get --recursive --host-requests 1",
-    "test:integration": "run-p -r start test:integration:run",
-    "test:update-golden-screenshots": "run-p -r start test:integration:update-golden-screenshots",
-    "test:integration:install-deps": "npx playwright install-deps chromium",
-    "test:integration:run": "npx playwright test",
-    "test:integration:update-golden-screenshots": "npx playwright test --update-snapshots"
+    "test:unit": "lerna run test --scope lit-dev-tools-cjs --scope lit-dev-tools-esm",
```
AndrewJakubowicz, line 17: This was a bug - sorry! Good find and fix! Thank you!

packages/lit-dev-tests/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
AndrewJakubowicz, line 1: Q: Each package has a tsconfig. Are they all the same?

PR #566, opened 2021-10-16
Supporting updating gists with new revisions

aomarks:
After creating a gist, typing Ctrl+S or Cmd+S will now update the gist with a new revision, instead of creating a brand new gist every time, and there is a new "Update Gist" button in the share menu.

(I plan to iterate some more on the design for these buttons in a followup, since it doesn't look quite right to me yet).

Also discovered that renaming/deleting is currently broken, because of a change in the latest version of MWC that I didn't notice before. Fixed in https://github.com/google/playground-elements/pull/236 and integrated into this PR via a pre-release.


packages/lit-dev-tools-esm/src/fake-github-middleware.ts
```diff
@@ -394,4 +401,71 @@ class FakeGitHub {
...
+   * Simulates PATCH https://api.github.com/gists/<id>, the API for creating a
+   * new revision of an existing GitHub gist.
+   *
+   * Documentation:
+   * https://docs.github.com/en/rest/reference/gists#update-a-gist
+   */
+  async updateGist(ctx: Koa.Context) {
+    ctx.set('Access-Control-Allow-Origin', '*');
+    const accept = ctx.get('accept');
+    if (accept !== 'application/vnd.github.v3+json') {
```
AndrewJakubowicz, line 415: Nit: Just a mock so not required. But these initial validation functions are the same across each endpoint so could be separated.

packages/lit-dev-tools-esm/src/fake-github-middleware.ts
```diff
@@ -394,4 +401,71 @@ class FakeGitHub {
...
+      newFile.filename = newFilename;
+    }
+    const oldFiles = gist.files;
+    for (const [oldFilename, oldFile] of Object.entries(oldFiles)) {
+      // To delete a file, you set its content to an empty string. If omitted,
+      // the existing file is unchanged. See
+      // https://github.community/t/deleting-or-renaming-files-in-a-multi-file-gist-using-github-api/170967
+      if (newFiles[oldFilename] === undefined) {
+        newFiles[oldFilename] = oldFile;
+      } else if (!newFiles[oldFilename]?.content) {
```
AndrewJakubowicz, line 463: Worth making this more specific to ensure there is a `content` prop set and it's exactly `""`?

AndrewJakubowicz (approved):
LGTM! Since behind feature flag these aren't blocking but a couple thoughts:

I expected the "update gist" option to work by default, if you're on your own gist. To reproduce the precise behavior:

1. Create a gist. Currently it's possible to update gist with cmd+S.
2. Refresh page.
3. Ctrl+S no longer updates, now it opens prompt.

I don't have a good thought on how to resolve this next one. But if you are updating a gist and then want to copy the non-gist URL, it's possible to lose the prior gist url. For example you are working on a gist and have both "create" & "update". Then you maybe intentionally or accidentally click "copy url" in the "Long URL" section. (It seems possible to unintentionally not realize that the long url is a different url from the gist url).

After this action the "update gist" button disappears. You can back out of it though so maybe it's not critical.
Will it be possible to figure out from your github gists the playground url? Maybe the project is created with a README.md with the lit.dev playground url so that it's possible to easily get back to your playground project from Github.

^ These are all non actionable for this PR and just ideas.

Fantastic! Manually tested locally as well!

packages/lit-dev-tools-esm/src/fake-github-middleware.ts
aomarks: Done

packages/lit-dev-tools-esm/src/fake-github-middleware.ts
aomarks: Content can be empty string or undefined to trigger a delete. Updated comment to clarify.

PR #572, opened 2021-10-20
Preserve file order and hidden bit in gists

aomarks:
Gist files are always returned in alphabetical order, so previously we would lose the original project ordering. Also, hidden files like the default `package.json` would suddenly re-appear after sharing as a gist, because their `hidden` bit was lost.

To address this, we now create a special `__lit-dev-playground-metadata__.json` file in gists, which encodes a position and hidden bit for each file. There might be additional metadata we can throw in here too in the future.

### DIFF
--- a/packages/lit-dev-content/src/util/gist-conversion.ts
+++ b/packages/lit-dev-content/src/util/gist-conversion.ts
@@ -12,10 +12,10 @@ import type {SampleFile} from 'playground-elements/shared/worker-api.js';
  * primitives.
  */
 type GistPlaygroundMetadata = {
-  files: {
+  files?: {
     [filename: string]: {
-      /** File display order . Gist files are always returned ~alphabetically. */
-      position: number;
+      /** File display order. Gist files are always returned alphabetically. */
+      position?: number;
       /** Whether the file is hidden. */
       hidden?: boolean;
     };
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/src/util/gist-conversion.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Convert a GitHub gist to an array of Playground files.
+ */
+export const gistToPlayground = (gistFiles: GistFiles): SampleFile[] => {
+  const metadataFile = gistFiles[METADATA_FILENAME];
+  let metadata: GistPlaygroundMetadata = {files: {}};
+  if (metadataFile) {
+    try {
+      metadata = JSON.parse(metadataFile.content) as GistPlaygroundMetadata;
```
AndrewJakubowicz, line 62: I don't know if this is actionable, but we don't validate the structure of this metadata file. So if someone has hand written valid json that parses it will be cast and result in runtime errors.

But I think that's also probably fine.
This came to mind when I was wondering if the `.position` would always be a number.

AndrewJakubowicz (approved):
This is awesome. I really like the idea of a metadata file. Provides a really cool public API to crafting a "perfect" playground example. I like that with this there is a way to hide more files, or re-order them.

Exceptional

PR #573, opened 2021-10-20
Enable prod gist integration (still behind mods flag)

aomarks:
Not quite ready to launch, but having this live on the site behind the flag will let us share previews of it. In particular, we need this for a11y review, since it's going to be difficult to keep an up to date staging URL working (because it requires constantly updating the redirect URL in the app settings, and only one version can be live at a time).

### DIFF
--- a/packages/lit-dev-tools-cjs/src/lit-dev-environments.ts
+++ b/packages/lit-dev-tools-cjs/src/lit-dev-environments.ts
@@ -13,15 +13,15 @@ interface LitDevEnvironment {
   googleAnalyticsId: string;
   reportCspViolations: boolean;
   playgroundSandboxUrl: string;
-  githubMainUrl: string | undefined;
-  githubApiUrl: string | undefined;
-  githubAvatarUrl: string | undefined;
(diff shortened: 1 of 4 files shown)


packages/lit-dev-tools-cjs/src/lit-dev-environments.ts
```diff
@@ -178,12 +178,14 @@ const prod = environment({
...
-  githubAuthorizeRedirectUrl: undefined, // Not set up yet
-  githubClientId: undefined, // Not set up yet
-  githubClientSecret: undefined, // Not set up yet
+  githubMainUrl: 'https://github.com/',
+  githubApiUrl: 'https://api.github.com/',
+  githubAvatarUrl: 'https://avatars.githubusercontent.com/',
+  githubAuthorizeRedirectUrl: 'https://lit.dev/playground/signin/',
+  githubClientId: 'd5c8e81abdf8867459c1',
+  get githubClientSecret() {
+    return stringEnv('GITHUB_CLIENT_SECRET');
```
AndrewJakubowicz, line 187: Will the build break if these are misconfigured and not provided?

PR #583, opened 2021-11-02
Include use of TS declare syntax

KeithHenry:
TS can use class fields with decorators without turning off `useDefineForClassFields` at the cost of losing inline initialisation.

See https://github.com/lit/lit.dev/pull/479 and https://github.com/lit/lit/issues/1985 for discussion.


arthurevans:
Requesting a review from @sorvell who wrote the current section.

justinfagnani (commented):
@KeithHenry sorry this review took so long! The `declare` info was already added, but the `es2020` bit wasn't, if you care to resubmit.

PR #586, opened 2021-11-09
Hide content in playground example drawer when closed

aomarks:
Otherwise the examples are still visible to screen readers and are keyboard focusable.


AndrewJakubowicz (approved):
Manually tested on prod site and preview site. Fantastic improvement!

PR #589, opened 2021-11-30
Make API docs generator configurable

aomarks:
Part of running lit 1.x docs on the new site. Refactors the TypeDoc-based API docs generator to  work off a config, instead of being hard-coded for Lit 2.x. Will add Lit 1.x configs in a followup.


sorvell (approved):

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -0,0 +1,194 @@
...
+const litDir = pathlib.join(gitDir, 'packages', 'lit');
+const srcDir = pathlib.join(litDir, 'src');
+const outDir = pathlib.join(workDir, 'api-data');
+
+/**
+ * lit.dev API docs configuration for Lit 2.x
+ */
+export const lit2Config: ApiDocsConfig = {
+  repo: 'https://github.com/lit/lit',
+  sha: 'f8ee010bc515e4bb319e98408d38ef3d971cc08b',
```
justinfagnani, line 23: can you add a comment? Is this the SHA the API docs are generated from? If we want to generate docs for a specific release, can this be a sha-like and we use the release tag instead for readability?

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -0,0 +1,194 @@
...
+    },
+    {
+      slug: 'misc',
+      title: 'Misc',
+    },
+  ],
+
+  pageForSymbol(node): string {
+    const entrypoint = node.entrypointSources?.[0]?.fileName ?? '';
+    if (entrypoint.includes('/directives/')) {
```
justinfagnani, line 95: comment? Is this to put all directives on one page?

justinfagnani (approved):

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
aomarks: It's used for the `git checkout` command, and for generating GitHub `blob` URLs, so yeah it can be any commit reference. I renamed the field from `sha` to `commit`, and updated the comment in the interface declaration.

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
aomarks: Yep that's right -- isn't it kinda self-documenting though? The method is documented in the interface (because note there will be multiple of these configs soon).

PR #590, opened 2021-11-30
Support for multiple API docs repos

aomarks:
Misc changes to support analyzing multiple repos/versions when generating API docs.


packages/lit-dev-tools-cjs/src/api-docs/generate.ts
```diff
@@ -0,0 +1,111 @@
...
+  return false;
+};
+
+/**
+ * Clone the given Git repo URL at the given commit SHA into the given
+ * directory. If the directory already exists, do nothing.
+ */
+const cloneIfNeeded = async (repo: string, sha: string, dir: string) => {
+  if (await fileExists(dir)) {
+    console.log(`${dir} already exists, skipping git clone`);
```
justinfagnani, line 42: shouldn't you make sure to do the checkout in case the directory exists, but not at the right SHA?

justinfagnani (approved):

packages/lit-dev-tools-cjs/src/api-docs/generate.ts
aomarks: I want to support the use case where you are working on the code annotations directly from the cloned directory. If you make a change and then check that in, it would be bad to automatically blow that change away when you then go and re-generate the docs.

I added a check that the SHAs match, and throws an exception if it doesn't, telling you to either update the config or delete the directory to fix things.

PR #591, opened 2021-11-30
Check in API data instead of building it every time

aomarks:
Previously, we checked out, built, and analyzed the Lit every time we deployed. Now, we check the generated data files into the Git repo instead. There is also a new GitHub check that ensures the checked in files are up-to-date. The advantages of this are:

- We can see in PRs when the data files have changed. This lets us see the effects of changes from the annotations in the Lit source, changes to the API generator, or other changes/bugs.

- Deploying should be much faster, since we no longer need to check out, build, and analyze the Lit repo as part of deploying. This will help even more once we are also analyzing the Lit 1.x repos.


justinfagnani (approved):

.github/workflows/check-api-data-in-sync.yml
```diff
@@ -0,0 +1,24 @@
+# This action fails if the API docs JSON data files have changed but have not
+# been committed.
+
+# We check generated API data into the Git repo because otherwise it is very
+# difficult to tell when something has changed or broken in them. It also makes
+# deploying faster, since we don't need to clone and analyze repos every time.
```
AndrewJakubowicz, line 6: Does this mean if you are doing an unrelated change and the API docs change, you need to include the API change or be blocked from merging by Github Actions?

Worth adding a comment here for how to unblock yourself if this is failing?

AndrewJakubowicz (approved):
AMAZING! LGTM!

.github/workflows/check-api-data-in-sync.yml
aomarks: The API docs shouldn't change if you're doing something unrelated. You'd have to change the API generating code, or the config that determines which repo + commit we are analyzing.

PR #592, opened 2021-11-30
Update Lit 2 API docs to latest commit

aomarks:
Updates the Lit 2 API docs to the latest commit .

Also disables HTML pass-thru when rendering markdown from JSDoc, since the latest code has an unquoted <style> tag which was breaking rendering.

### DIFF
--- a/packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
+++ b/packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
@@ -20,7 +20,7 @@ const srcDir = pathlib.join(litDir, 'src');
  */
 export const lit2Config: ApiDocsConfig = {
   repo: 'lit/lit',
-  commit: 'f8ee010bc515e4bb319e98408d38ef3d971cc08b',
+  commit: '8c4c845b016bdbe93bf644d0160415d011166faa',
   gitDir,
   tsConfigPath: pathlib.join(litDir, 'tsconfig.json'),
   pagesOutPath: pathlib.resolve(workDir, 'pages.json'),
--- a/packages/lit-dev-api/api-data/lit-2/pages.json
+++ b/packages/lit-dev-api/api-data/lit-2/pages.json
@@ -6,7 +6,7 @@
       "v1": "api/lit-element/LitElement/"
     },
     "repo": "lit/lit",
-    "commit": "f8ee010bc515e4bb319e98408d38ef3d971cc08b",
+    "commit": "8c4c845b016bdbe93bf644d0160415d011166faa",
     "items": [
       {
         "name": "LitElement",
@@ -74,7 +74,7 @@
                 "sources": [
                   {
                     "fileName": "packages/reactive-element/src/reactive-element.ts",
-                    "line": 878,
+                    "line": 968,
                     "moduleSpecifier": "reactive-element/reactive-element.js"
                   }
                 ],
@@ -168,7 +168,7 @@
                 "sources": [
                   {
                     "fileName": "packages/reactive-element/src/reactive-element.ts",
-                    "line": 440,
+                    "line": 509,
                     "moduleSpecifier": "reactive-element/reactive-element.js"
                   }
                 ],
@@ -224,10 +224,14 @@
               {
                 "name": "addController",
                 "kindString": "Method",
+                "comment": {
+                  "shortText": "Registers a `ReactiveController` to participate in the element's reactive\nupdate cycle. The element automatically calls into any registered\ncontrollers during its lifecycle callbacks.",
+                  "text": "If the element is connected when `addController()` is called, the\ncontroller's `hostConnected()` callback will be immediately called."
+                },
                 "sources": [
                   {
                     "fileName": "packages/reactive-element/src/reactive-element.ts",
-                    "line": 771,
+                    "line": 860,
                     "moduleSpecifier": "reactive-element/reactive-element.js"
                   }
                 ],
@@ -282,10 +286,13 @@
               {
                 "name": "removeController",
                 "kindString": "Method",
+                "comment": {
+                  "shortText": "Removes a `ReactiveController` from the element."
+                },
                 "sources": [
                   {
                     "fileName": "packages/reactive-element/src/reactive-element.ts",
-                    "line": 785,
+                    "line": 875,
                     "moduleSpecifier": "reactive-element/reactive-element.js"
                   }
                 ],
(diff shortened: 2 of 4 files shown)


AndrewJakubowicz:
Looking over this. FYI, deploy timed out.

AndrewJakubowicz (approved):
Skimmed over the text in the json. I know it's extra work so not actionable, but it could be useful to extract the text changes separate to all the line and sha changes. Then we could review the content changes separately from all the sha's.

LGTM with green Github Actions!

aomarks:
> Skimmed over the text in the json. I know it's extra work so not actionable, but it could be useful to extract the text changes separate to all the line and sha changes. Then we could review the content changes separately from all the sha's.

Yeah agree. I think we could clean up this format by removing fields we don't actually need to render (like numeric IDs). May do in a followup.

aomarks:
> Looking over this. FYI, deploy timed out.

This was actually a new failure, because this PR brought in a new JSDoc comment that had an unquoted `<style>` tag inside of it, which then caused the HTML minifier and search-index-generator to crash.

We should never render raw HTML from JSDoc comments, so I've added a new custom filter that we now use that's the same as our normal markdown renderer but with HTML-pass-thru turned off. PTAL!

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -20,7 +20,7 @@ const srcDir = pathlib.join(litDir, 'src');
  */
 export const lit2Config: ApiDocsConfig = {
   repo: 'https://github.com/lit/lit',
-  commit: 'f8ee010bc515e4bb319e98408d38ef3d971cc08b',
+  commit: 'eb5c5d2b2159dcd8b2321fa9a221b8d56d127a11',
```
justinfagnani, line 23: should this really be the latest commit, or the current release?

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -20,7 +20,7 @@ const srcDir = pathlib.join(litDir, 'src');
  */
 export const lit2Config: ApiDocsConfig = {
   repo: 'https://github.com/lit/lit',
-  commit: 'f8ee010bc515e4bb319e98408d38ef3d971cc08b',
+  commit: '8c4c845b016bdbe93bf644d0160415d011166faa',
```
justinfagnani, line 23: This is why I might prefer to see release tags used instead of SHAs. I would have thought this would be commit for the lit@2.0.2 release: https://github.com/lit/lit/releases/tag/lit%402.0.2 so it took a bit to realize this was the for the lit-html@2.0.2 release. Of course release tags in the monorepo are going to have a somewhat similar question of package release was the most recent...

aomarks: Yeah, we rarely release the lit package, so using the lit release tag would mean we'll lag behind on most of the docs updates, which are mostly in the other packages.

aomarks: We could just pick any tag from the most recent set of releases each time. That would work, though the string is included in the "view source" links we generate, so you might be clicking on "view source" for a thing in one package, and see the name of another package in the URL, which could be slightly confusing (e.g. you click on LitElement but the URL has @lit/localize in it).

justinfagnani: In this case we really want the most recent release tag/SHA of `lit`, `lit-html`, `lit-element`, or `@lit/reactive-element`, right? I guess a comment saying that might make it pretty clear what any updater should do here, and we could keep it a SHA.

aomarks: Ok, PTAL. I used the latest lit-html tag and added a comment.

"view source" URLs now look like this: https://github.com/lit/lit/blob/lit-html@2.0.2/packages/reactive-element/src/reactive-element.ts#L801

We could also automatically convert the tag into a SHA for these URLs, if we don't want the slightly odd arbitrary package tag appearing instead of a SHA. WDYT?

AndrewJakubowicz (commented):
Been doing some manual tests. I don't know which are blocking, but an early one I identified:

The example code block for [addInitializer](https://pr592-1f13152---lit-dev-5ftespv5na-uc.a.run.app/docs/api/LitElement/#LitElement.addInitializer) is missing syntax highlighting. It does seem to have the correct class applied.


It seems a major portion of code blocks have lost syntax highlighting.

AndrewJakubowicz (changes requested):
Marking as requesting changes regarding the API syntax highlighting issue.

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -20,7 +20,10 @@ const srcDir = pathlib.join(litDir, 'src');
  */
 export const lit2Config: ApiDocsConfig = {
   repo: 'https://github.com/lit/lit',
-  commit: 'f8ee010bc515e4bb319e98408d38ef3d971cc08b',
+  // To update this property, go to https://github.com/lit/lit/releases and find
+  // the most recent release of either lit, lit-html, lit-element, or
+  // @lit/reactive-element, and use that tag.
```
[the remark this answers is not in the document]
aomarks: Most of our changes affect the `lit-html` and `lit-element` packages, and much less often the `lit` package. So if we update some documentation and release `lit-html`, then there will not be a new `lit` tag, only a new `lit-html` tag. So we have to pick a tag from whichever package was most recently published.

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
justinfagnani: That's in some ways a much nicer URL, but in others it's a little weird because you're looking at reactive element with the `lit-html@2.0.2` tag. 🤷‍♂️ 

My brain is going to more complicated schemes where the URL is based on the package and its latest tag, but I don't think that's actually a good idea.

justinfagnani: Yeah, that wouldn't be good because the re-exporting packages would point to older SHAs than newer releases of the packages they re-export. I suspect a SHA here would be less confusing because it's opaque, or we could come up with a date-or sequence-based immutable tag we use every time we release. SHA works now though and unblocks this PR :)

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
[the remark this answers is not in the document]
AndrewJakubowicz: Ahh and now the comment makes a whole lot of sense. I had the confusion of how lit-html was impacting lit or reactive-element. Thank you!

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
aomarks: Switched back to the SHA!

aomarks:
> The example code block for [addInitializer](https://pr592-1f13152---lit-dev-5ftespv5na-uc.a.run.app/docs/api/LitElement/#LitElement.addInitializer) is missing syntax highlighting. It does seem to have the correct class applied.
> 
> It seems a major portion of code blocks have lost syntax highlighting.

Should be fixed now! There seems to be a bug with markdownit where it doesn't like having multiple configuration instances of itself. New approach where we switch html on/off dynamically seems to work fine.

aomarks:
> Yeah agree. I think we could also clean up this format by removing fields we don't actually need to render (like numeric IDs). May do in a followup.

This was done -- PR has been rebased so the API JSON diff should be much easier to read now.

aomarks:
@AndrewJakubowicz PTAL!

justinfagnani (approved):

AndrewJakubowicz (approved):
LGTM!

PR #594, opened 2021-12-01
Serve old site articles from new site (undiscoverable)

aomarks:
**Note to reviewer**: I strongly suggest looking at each commit separately.

Adds all of the markdown content from lit-html.polymer-project.org and lit-element.polymer-project.org to lit.dev under `/docs/v1/`.

This content is not linked from anywhere, so it will not be discoverable until subsequent PRs set up a version switcher, and we do redirects from the old domains. Plan:

- [X] Serve old article content (this PR)
- [ ] Serve old API content
- [ ] Add a "these are old docs" banner
- [ ] Add a version switcher


packages/lit-dev-content/.eleventy.js
```diff
@@ -143,22 +143,34 @@ ${content}
...
-        }
-        if (a.fileSlug < b.fileSlug) {
-          return -1;
-        }
-        if (b.fileSlug < a.fileSlug) {
-          return 1;
-        }
-        return 0;
-      });
+      .filter((page) => !page.url.includes('v1'))
```
justinfagnani, line 172: Does that glob not support negation, like `.getFilteredByGlob('site/docs/!(v1)/**')`?

justinfagnani (approved):
Thanks for the nice commits!

packages/lit-dev-content/.eleventy.js
aomarks: Done (though had to do `.getFilteredByGlob(['site/docs/*', 'site/docs/!(v1)/**'])`)

PR #595, opened 2021-12-02
Serve old API from new site (undiscoverable)

aomarks:
**Note to reviewers**: I strongly suggest looking at each commit separately.

Renders the old lit-element-2.x and lit-html-1.x APIs within the new site, using the same style as the new docs.

The two libraries each have their own API section.

This content is not linked from anywhere, so it will not be discoverable until subsequent PRs set up a version switcher and we add redirects from the old domains. Plan:

- [X] Serve old API content (this PR)
- [ ] Add a "these are old docs" banner
- [ ] Add a version switcher


justinfagnani (approved):

sorvell (approved):

sorvell:
Noticed that we don't show property visibility (e.g. protected v. public) and filed: https://github.com/lit/lit.dev/issues/597

PR #599, opened 2021-12-02
Show banner and version selector for old docs

aomarks:
Adds a "You're viewing docs for an older version of Lit" banner to the top of the `/v1/` docs, and a Version selector in the top-left.

Version selector is hidden behind the `?mods=old-docs` mod for now.

Also fixes a CSS bug where the margin above the LHS nav was wrong when the content was too short (affected the first page of the old docs, and also the Community page of the new docs).

Currently when we switch versions, we always jump to the first page of the docs. In a follow up PR, I'll make it jump to the more specific matching page.

Plan:

- [X] Serve old API content
- [X] Add a "these are old docs" banner  (this PR)
- [X] Add a version switcher  (this PR)
- [ ] Make switching versions bring you to the matching page

### DIFF
--- a/packages/lit-dev-content/rollup.config.js
+++ b/packages/lit-dev-content/rollup.config.js
@@ -31,6 +31,7 @@ export default [
   {
     input: [
       'lib/components/copy-button.js',
+      'lib/components/litdev-banner.js',
       'lib/components/litdev-drawer.js',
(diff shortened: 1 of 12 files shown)


aomarks:
This is ready for review again, PTAL!

sorvell (approved):

PR #601, opened 2021-12-03
Go to the closest matching page when switching docs versions

aomarks:
**Note to reviewers**: I strongly suggest looking at each commit separately.

When switching between v1 and v2 docs, we'll now take you to the most similar page in the other version, instead of just to the landing page.

- Works in either direction
- Works from the "Click here for latest version" link
- Also works from the "Version" switcher
- Configured using front-matter in the markdown files
- For the API, configured from the API config files

I just used my own judgment to find the most similar pages. In a few cases the decision wasn't 100% obvious so feel free to suggest some better ones if you see anything odd.

Plan:

- [X] Serve old API content
- [X] Add a "these are old docs" banner
- [X] Add a version switcher
- [X] Make switching versions bring you to the matching page  (this PR)

### DIFF
--- a/packages/lit-dev-content/site/docs/index.md
+++ b/packages/lit-dev-content/site/docs/index.md
@@ -4,6 +4,8 @@ eleventyNavigation:
   key: What is Lit?
   parent: Introduction
   order: 1
+versionLinks:
+  v1:
 ---
 
 ![Lit Logo]({{ site.baseurl }}/images/logo.svg){.logo width="425" height="200"}
--- a/packages/lit-dev-content/site/docs/v1/lit-html/release-notes.md
+++ b/packages/lit-dev-content/site/docs/v1/lit-html/release-notes.md
@@ -4,6 +4,8 @@ eleventyNavigation:
   key: Release notes
   parent: lit-html
   order: 10
+versionLinks:
+  v2: releases/upgrade/
 ---
 
 ## lit-html 1.3.0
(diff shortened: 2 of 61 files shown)


packages/lit-dev-content/site/docs/v1/lit-html/release-notes.md
```diff
@@ -4,6 +4,8 @@ eleventyNavigation:
   key: Release notes
   parent: lit-html
   order: 10
+versionLinks:
+  v2: releases/upgrade/
```
justinfagnani, line 8: ok for now, but we need to finish out the release notes section on the new docs

packages/lit-dev-content/site/docs/index.md
```diff
@@ -4,6 +4,8 @@ eleventyNavigation:
   key: What is Lit?
   parent: Introduction
   order: 1
+versionLinks:
+  v1:
```
justinfagnani, line 8: intentional? if so, does `/` work?

justinfagnani (approved):
🎉

packages/lit-dev-content/site/docs/index.md
aomarks: We're doing effectively a naive `rootPath + '/' + versionLink` concat in the eleventy template here ,so using `/` here would make a double slash. I could also leave it off entirely, since it always defaults to the first page (which in both cases are "What is X"), but I figured this was kinda an explicit signal that I looked for the right matching path here.

packages/lit-dev-content/site/docs/v1/lit-html/release-notes.md
aomarks: +1

justinfagnani (approved):
Still LGTM

PR #603, opened 2021-12-04
Redirect old urls and add rel=canonical links

aomarks:
- Redirects `lit-html.polymer-project.org` and `lit-element.polymer-project.org` URLs to their new locations at `lit.dev`.
 
- Decided to redirect to the `v1` pages instead of the latest version because it seems less controversial/disruptive -- and we'll still have the chance to do more aggressive redirects to the latest content if we decide to do that sometime later.

- Adds `<link rel="canonical">` links to the old docs pages. This should help search results shift over to the latest version.

- This won't have any meaningful effect until we also update the DNS records to point at our new cloud project. I'll start this process on monday.

- I tested this locally by manually editing my `/etc/hosts` file. Seems to be working as expected. I have switched from `ctx.hostname` to `ctx.URL` here, so there is some small chance that Koa won't have set that correctly (I believe it needs to check `X-Forwarded-Host` to create the right URL). We'll be able to test that after deploying this by confirming the `www.lit.dev` redirect is still working.

Plan:

- [X] Serve old API content
- [X] Add a "these are old docs" banner
- [X] Add a version switcher
- [X] Make switching versions bring you to the matching page
- [X] Set up redirects by checking the `Host` header (this PR)
- [X] Set up `<link rel="canonical">` tags linking the old docs at their latest versions (this PR)


packages/lit-dev-content/site/_includes/default.html
```diff
@@ -16,7 +16,8 @@
     {% inlinejs "global/initialize-typescript-attribute.js" %}
     {% inlinejs "global/mobile-drawer.js" %}
 
-    {% if showOldDocsBanner %}
+    {% if selectedVersion !== latestVersion %}
+      <link rel="canonical" href="{{ versions[latestVersion].path }}/{{ versionLinks[latestVersion] }}">
```
justinfagnani, line 20: should there be another check that there is a versions[latestVersion] just in case?

justinfagnani (approved):

packages/lit-dev-content/site/_includes/default.html
aomarks: Hmm, not sure of an easy way to do that -- I dont see an "assert" or anything in nunjucks. I guess we could have a custom directive. But I sort of feel like there are dozens of similar things in our templates that are similarly brittle...

PR #605, opened 2021-12-05
Make API docs JSON format diffs more readable

aomarks:
- Removes the numeric `id` field. We don't use it in rendering. This was especially bad, because it's non-deterministic, so every build generated arbitrarily different JSON (and was causing our API-data-in-sync check to always fail).

- Moves GitHub URL generation into the eleventy template. This way, changing the SHA only updates one field in the JSON file, instead of every single reference.

- Removes comments being duplicated into multiple locations.

- Removes a number of other fields that we don't need in rendering.

- Results in a drop of 19K lines.

- **Note**: I did a diff of all generated HTML, to confirm that there are zero changes to the rendered output here (using https://meldmerge.org/ which can nicely diff folders).


packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
```diff
@@ -412,6 +412,64 @@ export class ApiDocsTransformer {
     }
   }
 
+  /**
+   * Remove fields that we don't need for rendering. This makes reading diffs
+   * much easier, since we check the generated JSON file in.
+   */
+  private prunePageData(node: unknown) {
+    if (node instanceof Array) {
```
AndrewJakubowicz, line 420: Non actionable question. This vs something like `Array.isArray`?

packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
```diff
@@ -412,6 +412,64 @@ export class ApiDocsTransformer {
...
+        (node as ExtendedDeclarationReflection).comment &&
+        (node as ExtendedDeclarationReflection).signatures?.[0]?.comment
+      ) {
+        delete (node as ExtendedDeclarationReflection).signatures?.[0]?.comment;
+      }
+      for (const [key, val] of Object.entries(node)) {
+        // Prune the child first, so that our "empty arrays and objects" check
+        // works more aggressively.
+        this.prunePageData(val);
+        if (
```
AndrewJakubowicz, line 437: Non actionable because it could be a larger change than I understand. But instead of justifying why we are opting things out, this could also be modeled as an allow-list, with justification for why something is left in. Not sure of the pro/cons but it would ensure that if typedoc added something, it wouldn't automatically be pulled in.

AndrewJakubowicz (approved):
LGTM! Did a small manual test but love that you included the html output diff comment.

Small non blocking comments.

packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
aomarks: The thing is that we don't build the data structure to begin with, we get it from TypeDoc and then augment it. So to do an allowlist approach, we'd need to re-build the data structure which I think would be more complicated.

packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
aomarks: I guess `instanceof Array` has problems when crossing iframe boundaries? But not really relevant here.

PR #611, opened 2021-12-08
Small playground a11y fixes

aomarks:
- Darker a11y suitable color for inactive copy button in gists panel
- Bump playground-elements version to pick up https://github.com/google/playground-elements/pull/244 which fixes some missing `aria-label` attributes


packages/lit-dev-content/site/css/docs-api.css
```diff
@@ -40,7 +40,7 @@
 }
 copy-button {
   height: 48px;
-  opacity: 0.5;
+  opacity: 0.75;
```
AndrewJakubowicz, line 43: We have two selectors for `copy-button`?

AndrewJakubowicz (approved):

AndrewJakubowicz:
Also did some manual testing - looked at playground & a page or two of documentation.

packages/lit-dev-content/site/css/docs-api.css
aomarks: Yeah, should probably refactor. Though there's an advantage to at least having the `height` set in the page stylesheets, because that way the sizing is correct even before the element upgrades...

PR #612, opened 2021-12-08
Make closing the gist flyout more reliable

aomarks:
- You can now close the flyout by clicking the "Share" button again. Previously this didn't work because clicking on it 1. caused the flyout to close itself (because it closes itself when it detects a click outside of itself), and then 2. the share button to open the flyout again (because it naively opened for every click).

- You can now close the flyout by clicking on the preview. Previously this didn't work because those clicks went to the iframe window, and never reached the host window. Now we temporarily set `pointer-events: none` when the flyout is open so that the main window gets those events instead.

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-playground-share-button.ts
+++ b/packages/lit-dev-content/src/components/litdev-playground-share-button.ts
@@ -93,6 +93,11 @@ export class LitDevPlaygroundShareButton extends LitElement {
   private _mostRecentSaveType: 'longurl' | 'gist' | undefined = undefined;
 
   /**
+   * When true, clicking on this button will not open the flyout.
+   */
+  private _ignoreClick = false;
+
+  /**
    * A function to allow this component to access the project upon save.
    */
   @property({attribute: false})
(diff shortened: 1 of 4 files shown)


packages/lit-dev-content/src/components/litdev-playground-share-button.ts
```diff
@@ -115,8 +115,13 @@ export class LitDevPlaygroundShareButton extends LitElement {
   }
 
   override update(changes: PropertyValues) {
-    if (changes.has('_open') && this._open) {
-      this._longUrl?.generateUrl();
+    if (changes.has('_open')) {
+      if (this._open) {
+        this._longUrl?.generateUrl();
+        this.dispatchEvent(new Event('opened'));
```
AndrewJakubowicz, line 126: Question: Worth registering event types into the TypeScript event map or making it more specific to the share button?

AndrewJakubowicz (approved):
LGTM and I manually tested in Safari and Chrome.

The iframe handling is genius.

packages/lit-dev-content/src/components/litdev-playground-share-button.ts
aomarks: There's no data in this event, and the event map doesn't inform any type checking or anything, so I guess it wouldn't achieve anything practical.

Also `'opened'` is such a generic name that it is likely to collide, and there's no way to declare an event tied to a specific element, right?

aomarks: That's true, maybe prefixing things with the CE name would be a good convention... haven't been doing that generally in this code base, but something to consider as a code improvement generally?

PR #613, opened 2021-12-08
Tweaks to gist metadata file

aomarks:
- Uses `zzzzzzzzzz` as the prefix for the metadata file instead of `__`, because that will sort last instead of first, allowing one of the user's files to become the implicit Gist title instead of it always being our weird metadata file.

- JSON indent the metadata file -- it might as well be readable.

### DIFF
--- a/packages/lit-dev-content/src/util/gist-conversion.ts
+++ b/packages/lit-dev-content/src/util/gist-conversion.ts
@@ -24,8 +24,15 @@ type GistPlaygroundMetadata = {
 
 /**
  * Filename for gist playground metadata.
+ *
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (approved):
Since the feature is behind a feature flag it probably doesn't matter that the prior metadata file shows up:

Tested by trying https://pr613-3add853---lit-dev-5ftespv5na-uc.a.run.app/playground/#gist=4b078a2303d92af5d6943efaf40eee03

It's a very minor nit.

aomarks:
> Since the feature is behind a feature flag it probably doesn't matter that the prior metadata file shows up:

Yeah, I think we just say existing gists are not really valid since we were in a kind of alpha before.

PR #585, opened 2021-11-04
New localization docs

aomarks:
Much more extensive localization docs.

Moves from a single page under "Related libraries" to its own section with multiple pages.

I took everything from https://github.com/lit/lit/blob/main/packages/localize/README.md, split it into multiple pages, and expanded it.

PR #615, opened 2021-12-09
Address localization comments from PR#585

aomarks:
This PR addresses all of the comments from @arthurevans in https://github.com/lit/lit.dev/pull/585.

### DIFF
--- a/packages/lit-dev-content/site/docs/localization/overview.md
+++ b/packages/lit-dev-content/site/docs/localization/overview.md
@@ -12,8 +12,8 @@ apps and components. Lit has first-party support for localization through the
 good choice over third-party localization libraries:
 
 - Native support for expressions and HTML markup inside localized templates. No
-  need for a new syntax and interpolation runtime for variable substitution —
-  just use the templates you already have.
+  need for a new syntax and interpolation runtime for variable substitution—just
+  use the templates you already have.
 
 - Automatic re-rendering of Lit components when the locale switches.
 
@@ -135,9 +135,11 @@ html`<button>${msg('Hello World')}</button>`;
 
 ### Strings with expressions
 
-If a string contains an expression and doesn't need to be tagged with `html`,
-then it must be tagged with `str` in order to be localizable. An error will be
-raised during extraction if you forget to include the `str` tag.
+Strings that contain an expression must be tagged with either `html` or `str` in
+order to be localizable. You should prefer `str` over `html` when your string
+doesn't contain any HTML markup, because it has slightly less performance
+overhead. An error will be raised when you run the `lit-localize` command if you
+forget the `html` or `str` tag on a string with an expression.
 
 Incorrect:
 <strike>
@@ -172,11 +174,15 @@ href="https://www.w3.org/International/articles/language-tags/index.en"
 target="_blank" rel="noopener">BCP 47 language tag standard</a>. Some examples
 of BCP 47 language tags are:
 
-- `en`: English
-- `es-419`: Spanish spoken in Latin America
-- `zh-Hans`: Chinese written in Simplified script
+- en: English
+- es-419: Spanish spoken in Latin America
+- zh-Hans: Chinese written in Simplified script
 
-Lit Localize defines a few important locale values:
+### Terms
+
+Lit Localize defines a few terms that refer to locale codes. These terms are
+used in this documentation, in the Lit Localize config file, and in the Lit
+Localize API:
 
 <dl class="params">
   <dt class="paramName">Source locale</dt>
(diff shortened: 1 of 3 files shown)


packages/lit-dev-content/site/docs/localization/overview.md
```diff
@@ -203,9 +209,8 @@ their own advantages.
 
 <div class="alert alert-info">
```
arthurevans, line 210: I know I recommended the run-in head, but looking at this section rendered on the web site, I realize that it looks kind of funny to have one sentence followed by a note. With both the initial version and my suggestion, it implies that the reader some kind of basis for deciding between the modes... But you haven't described them yet.

Often in this situation I use a bullet list with a brief summary of each option, then go into the full sections. Something like:

```
*   _runtime_ mode uses Lit Localize's APIs to load localized messages at runtime.
*   _transform_ mode eliminates the Lit Localize runtime code by building a separate JavaScript bundle for each locale. 
```

... Which could be followed either by the note (as currently formatted) or your original text, as a plain paragraph.

This just gives people a little preview of what they're going to find in the following sections.

packages/lit-dev-content/site/docs/localization/overview.md
```diff
@@ -444,7 +498,7 @@ Descriptions will be represented in XLIFF files using `<note>` elements.
 ## Message ids
```
arthurevans, line 498: I didn't notice this before, but would probably be better to use ID in text. (Except where referring to a literal property or in XLIFF markup.)  https://developers.google.com/style/word-list#id

Seems like it's more likely to read as  "eye-dee" like your driver's license rather than "id," the opposite of the superego.

packages/lit-dev-content/site/docs/localization/runtime-mode.md
```diff
@@ -37,44 +37,52 @@ of Lit Localize runtime mode that you can use as templates.
...
-[`output.localeCodesModule`](/docs/localization/cli-and-config/#general-settings)
-setting to generate a module which exports a `sourceLocale` string and
-`targetLocales` array that can be passed to `configureLocalization`, to ensure
-that your runtime code remains in sync with your config file.
-
-</div>
-
 `configureLocalization` returns an object with the following properties:
 
 ### getLocale
```
arthurevans, line 88: This formatting looks odd. Usually we'd use a colon to introduce a bullet list, but here it's followed immediately by a heading, which makes it look like some text is missing after `properties:`

I'd be inclined to either:

- Convert this to a definition list.
- Convert it to a bullet list.
- Enumerate the properties (either inline or in a bullet list) and add, "described in the following sections" or something.

packages/lit-dev-content/site/docs/localization/transform-mode.md
```diff
@@ -54,8 +54,8 @@ In your JavaScript or TypeScript project, optionally call
 `configureTransformLocalization`, passing an object with the following
 properties:
```
arthurevans, line 55: property?

packages/lit-dev-content/site/docs/localization/transform-mode.md
```diff
@@ -54,8 +54,8 @@ In your JavaScript or TypeScript project, optionally call
...
 
-- `sourceLocale: string`: Required locale code in which source templates in this
-  project are written.
+- `sourceLocale: string`: Locale in which source templates are written.
+  Specified as a locale code (for example: `"en"`).
 
 ```js
```
arthurevans, line 60: Code sample formatting seems odd, following directly after a bullet list. It's not clear whether it's supposed to be part of the description of `sourceLocale`, or a more general example.

If it's supposed to belong to `sourceLocale`, it should be indented 4 here so it lines up under the bullet text. If not, maybe add a sentence introducing it. Or even a fragment like, "For example: "

arthurevans: Also, as with the similar section in runtime mode, it's odd that the properties of the input object are shown as bullets, and the properties of the output object are shown as separate sections.  These should be consistent.

arthurevans (approved):
Looks good!

Added some minor suggestions that occurred on re-review.

packages/lit-dev-content/site/docs/localization/overview.md
aomarks: Done

packages/lit-dev-content/site/docs/localization/overview.md
aomarks: Done

packages/lit-dev-content/site/docs/localization/transform-mode.md
aomarks: Done

packages/lit-dev-content/site/docs/localization/runtime-mode.md
aomarks: Done (bullet list)

packages/lit-dev-content/site/docs/localization/transform-mode.md
aomarks: Done, done

PR #617, opened 2021-12-10
New blog design and atom feed

aomarks:
#### New blog design

- Front matter is now above the article instead of off to the side. I find my eye can more naturally find the title and author information now.

- Layout is now centered instead of left-aligned. Works better on wider screens, and is consistent with all our other pages.

- Theme is now light instead of dark. I'm thinking we should have theme-consistency across the site, instead of alternating it between pages. We should add a user-controlled light/dark theme toggle that affects the whole site.

- Adds a splash image which can be set per-blog post for some visual interest. Included a default one that's based on the colors/angles of the lit logo.

#### Other changes

- Added an Atom feed
- Added optional author GitHub links
- Centralized author data (images, links etc. had to be duplicated into each post previously)

The old design for reference:


packages/lit-dev-content/.eleventy.js
```diff
@@ -31,6 +31,7 @@ const {
 const {preCompress} = require('../lit-dev-tools-cjs/lib/pre-compress.js');
 const luxon = require('luxon');
 const crypto = require('crypto');
+const pluginRss = require('@11ty/eleventy-plugin-rss');
```
AndrewJakubowicz, line 34: Woooo!

arthurevans:
Besides RSS, what's the driver behind the new design? 

Rody proposed adding tags and syndicating posts (there's an old prototype at https://github.com/lit/lit.dev/pull/427)... Wondering how this would fit in.

aomarks:
> Besides RSS, what's the driver behind the new design?

Mainly improving readability. Specifically:

I find my eye can more naturally find the title and author information when it's above the content, instead of off to the side.

Centered layout works better on wider screens, and is consistent with our other pages.

Light theme, because I think we should have theme-consistency across the site, instead of alternating between pages (and we should later add a user-controlled light/dark theme toggle that affects the whole site.)

> Rody proposed adding tags and syndicating posts (there's an old prototype at #427)... Wondering how this would fit in.

Should fit in fine. We would just need to decide where those tags would go in the new design -- maybe just above or below the existing front-matter. (I do sort of feel that we have so little content on the blog right now that tags wouldn't really buy us too much now, but if we get more posts it seems great!).

packages/lit-dev-content/site/blog/feed.xml.njk
```diff
@@ -0,0 +1,36 @@
+---
+eleventyExcludeFromCollections: true
+---
+<?xml version="1.0" encoding="utf-8"?>
+<feed xmlns="http://www.w3.org/2005/Atom">
```
justinfagnani, line 5: I really don't know RSS, so... I trust you? :)

packages/lit-dev-content/site/css/blog-common.css
```diff
@@ -1,62 +1,29 @@
...
-  font-weight: 500;
-  line-height: 1.5em;
-  margin-top: 0;
-  font-size: 1.5em;
-}
-
-#blogContent {
-  padding: var(--blog-top-padding) 3em;
+article {
+  max-width: 800px;
```
justinfagnani, line 10: might be nice to have a site-wide CSS variable for this width

justinfagnani (approved):

packages/lit-dev-content/site/blog/feed.xml.njk
aomarks: Atom seems like the better format. For example, it supports multiple languages of the same article. Got the sense from reading around that it's more popular now.

It's also what Eleventy documents as its template, which 
I mostly copied here: https://www.11ty.dev/docs/plugins/rss/

Checked our output with https://validator.w3.org/feed/ too

packages/lit-dev-content/site/css/blog-common.css
aomarks: Yeah... it's not consistent right now. The home page can get quite a lot wider, but it uses multiple columns so that seems fine. Our docs main content page is slight off 800 px because I think I tuned it for some specific break points based on the other columns. Maybe could be unified, but don't think I'll get into that for this PR.

arthurevans (approved):
Like the new look and I agree that a consistent light/dark theme (and future light/dark switch) makes much more sense.

I only looked at the staged site. I have nothing useful to say about RSS or CSS or any other acronym ending in SS.

packages/lit-dev-content/site/css/blog-post.css
```diff
@@ -1,92 +1,98 @@
...
+  text-indent: 0;
 }
 
-#blogContent {
-  padding: var(--blog-top-padding) 3em;
-  font-family: "Open Sans", sans-serif;
-  font-size: 0.88889em;
+ul.authorLinks > li:not(:first-of-type)::before {
+  content: "·";
+  margin: 0 8px 0 4px;
```
AndrewJakubowicz, line 54: SUPER NIT: Interested why not center the dot? I.e. `margin: 0 8px;`?

aomarks: Hmm weird. On Linux it is centered currently, but you're right on macOS it is off-center.

AndrewJakubowicz (approved):

packages/lit-dev-content/site/css/blog-post.css
aomarks: Turns out it was a dev mode vs production mode difference! Very weird, I truly could not figure out why they were rendering differently though -- all relevant CSS seemed identical. Using flex has made it consistent.

abdonrd:
The new design looks great! 👏🏻

aomarks:
> The new design looks great! 👏🏻

Thanks!

aomarks:
> Why does the `<link>` element have a `data-title` instead of a `title`?

Err, I don't know. I actually based this on what `https://web.dev` does. TBH I don't know what the `title` attribute does in this case. Do you have more info?

simevidas:
Feed readers use the `title` attribute value as the title of the feed. If the `title` attribute is missing, the feed reader probably uses the title of the website itself (at least Feedly seems to do that). In your case, the title that you get in Feedly is “Lit Blog”, which is good, but there may be issues in other feed readers if the `title` attribute is missing.

Screenshot 2021-12-14 at 20 12 44

aomarks:
> Feed readers use the `title` attribute value as the title of the feed. If the `title` attribute is missing, the feed reader probably uses the title of the website itself (at least Feedly seems to do that). In your case, the title that you get in Feedly is “Lit Blog”, which is good, but there may be issues in other feed readers if the `title` attribute is missing.

Don't they use the `<title>` tag inside the Atom or RSS resource itself, rather than the `title` attribute of a link pointing to the feed? Our actual feed does have a `<title>` tag (see https://lit.dev/blog/atom.xml), and Feedly is displaying it correctly.

simevidas:
Ah, yes. Feed readers use the feed’s `<title>` value for sure. However, I found a use-case for the `title` attribute. RSS-related browser extensions may use it to display information about the website’s feed. I use the Feed Preview extension. Notice how it says “Untitled feed” on web.dev because the RSS `<link>` element doesn’t have a `title` attribute.

Screenshot 2021-12-14 at 21 12 23

aomarks:
> Feed Preview extension

Ah thanks for the info, which extension exactly so I can try it out?

aomarks:
> Ah thanks for the info, which extension exactly so I can try it out?

Found it (assume it was https://addons.mozilla.org/en-US/firefox/addon/feed-preview/)

simevidas:
Yep, that’s the one. 

I also tested the [RSS Feed Reader](https://chrome.google.com/webstore/detail/rss-feed-reader/pnjaodmkngahhkoihejjehlcdlnohgmp) extension in Chrome (the first result for “RSS”). When I try to add Lit Blog’s feed via that extension, notice how it shows the feed’s URL instead of the title. That’s another case where an extension relies on the `title` attribute.

Screenshot 2021-12-14 at 23 12 04

aomarks:
> Yep, that’s the one.
> 
> I also tested the [RSS Feed Reader](https://chrome.google.com/webstore/detail/rss-feed-reader/pnjaodmkngahhkoihejjehlcdlnohgmp) extension in Chrome (the first result for “RSS”). When I try to add Lit Blog’s feed via that extension, notice how it shows the feed’s URL instead of the title. That’s another case where an extension relies on the `title` attribute. Once you add the feed, you get the correct name, “Lit Blog”, in the UI, but in the Add Feed dialog at least, the extension tries to pull the title from the `title` attribute.

Fixing the `title` attribute was done in https://github.com/lit/lit.dev/pull/626 and is live now. Should be working!

PR #618, opened 2021-12-12
Fix invalid lit-html 1.x styles redirect

aomarks:


packages/lit-dev-tests/src/check-redirects.ts
```diff
@@ -91,18 +95,21 @@ const checkAllRedirects = async () => {
 
   let fail = false;
   const promises = [];
-  for (const [from, to] of pageRedirects) {
+  const allRedirectTargets = new Set([
```
AndrewJakubowicz, line 98: Great improvement to our tests!

AndrewJakubowicz (approved):
AMAZING!

Great fix and then paired with a test.

PR #620, opened 2021-12-13
Don't output feed.xml to feed.xml/index.html

aomarks:
This will also correct the content-type response header from HTML to XML


AndrewJakubowicz (approved):
Nice! Did not manually test but change looks good!

PR #621, opened 2021-12-13
Atom feed improvements for Feedly and others

aomarks:
- Adds `webfeeds:icon` and `webfeeds:logo` tags, which Feedly apparently depends on (even though Atom already specifies its own `<icon>` and `<logo>` tags...)

- Adds a new `icon.svg` to use instead of `flame.svg`, which is square and has a little padding. Feedly crops non-square images, so the logo was getting cut off.

- Fixes broken Twitter links.

- Fixes missing `rel=alternate` attribute.

- Added a CSP policy to allow the built-in Chrome XML viewer to display the Atom feed nicely.

- Renames the feed from `feed.xml` to `atom.xml`. The originally published feed which was invalid caused Feedly to do some caching that means that no articles show up, and the icon is missing, even after clicking its reload button and waiting a few hours. A different URL will be a different feed from Feedly's perspective, so we get a reset. `atom.xml` is probably a better name anyway, to distinguish it from a hypothetical `rss.xml` that we could have (and maybe we'll find a reason to have both for some reason). Included a 301 redirect in case somebody has already subscribed.

Proof:

### DIFF
--- a/packages/lit-dev-content/site/_includes/default.html
+++ b/packages/lit-dev-content/site/_includes/default.html
@@ -40,7 +40,7 @@
 
     <link rel="preconnect" href="https://fonts.gstatic.com">
 
-    <link href="/blog/feed.xml" rel="alternate" type="application/atom+xml" data-title="Lit Blog feed">
+    <link href="/blog/atom.xml" rel="alternate" type="application/atom+xml" data-title="Lit Blog feed">
     {% block head %}{% endblock %}
   </head>
   <body>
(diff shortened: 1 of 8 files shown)


AndrewJakubowicz (approved):
Ran it through: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fpr621-b975fa8---lit-dev-5ftespv5na-uc.a.run.app%2Fblog%2Fatom.xml

Asked about the `iframe` and `litdev-switchable-sample`. Should work well enough. And this is an improvement.

LGTM!

PR #623, opened 2021-12-14
Update to Lit 2.1 API documentation

AndrewJakubowicz:
Bump generated API docs so they include new Lit 2.1 features.

Example of a new section: https://pr623-b0bd433---lit-dev-5ftespv5na-uc.a.run.app/docs/api/decorators/#queryAssignedElements

Will merge after Lit 2.1 release.

### DIFF
--- a/packages/lit-dev-api/api-data/lit-2/pages.json
+++ b/packages/lit-dev-api/api-data/lit-2/pages.json
@@ -6,7 +6,7 @@
       "v1": "api/lit-element/LitElement/"
     },
     "repo": "lit/lit",
-    "commit": "8c4c845b016bdbe93bf644d0160415d011166faa",
+    "commit": "483943034a62bded13eca0c982ff7c93ac6639b6",
     "items": [
       {
         "name": "LitElement",
@@ -2437,6 +2437,9 @@
               {
                 "name": "willUpdate",
                 "kindString": "Method",
+                "flags": {
+                  "isProtected": true
+                },
                 "sources": [
                   {
                     "fileName": "packages/reactive-element/src/reactive-element.ts",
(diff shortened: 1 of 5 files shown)


aomarks (approved):
Looks fine, but I did notice that the `options.slot` and `options.selector` parameters don't get documented.

I guess TypeDoc must consider those invalid, since they don't match any actual parameter.

It's slightly confusing right now, because the `options` descriptions just tell you to look at MDN, but the `slot` and `selector` options aren't documented there.

Also, we should merge this just after the release, I assume?

AndrewJakubowicz:
There is a confusingly link [`QueryAssignedElementsOptions`](https://pr623-b0bd433---lit-dev-5ftespv5na-uc.a.run.app/docs/api/decorators/#QueryAssignedElementsOptions) at the top of the options but I don't find it intuitive to click. Maybe also making the `slot` and `selector` link to the options interface docs?

aomarks:
> There is a confusingly link [`QueryAssignedElementsOptions`](https://pr623-b0bd433---lit-dev-5ftespv5na-uc.a.run.app/docs/api/decorators/#QueryAssignedElementsOptions) at the top of the options but I don't find it intuitive to click. Maybe also making the `slot` and `selector` link to the options interface docs?

Oh, true! I think it was the phrasing of "See MDN parameters section for available options" that sounded authoritative -- like that was the only place I should look. Also the elements decorator says "Also accepts two more optional properties, slot and selector.", but the nodes one doesn't have a similar sentence.

AndrewJakubowicz:
This is a great comment. Will address on Lit core.

AndrewJakubowicz:
Thanks Al! Addressed your changes [on Lit](https://github.com/lit/lit/commit/1b67e03be7b969c543f639c49542bb7be3086d32) and updated the Lit 2 `sha` here to pull in the change.

aomarks (commented):
Should we be waiting to merge these PRs until we do the release? The docs won't match the released version otherwise, right?

AndrewJakubowicz:
Right! Blocked on Lit 2.1 release.

aomarks (approved):

Issue #503, opened 2021-09-15
Use @link for jsdoc, and make compatible with lit.dev API documentation


AndrewJakubowicz:
Tracking issue from https://github.com/lit/lit/pull/2161.

We should start using e.g. `{@link ReactiveController}`, since that will work in VSCode too. We will need to change or update this regexp to extract them for our lit.dev API docs: https://github.com/lit/lit.dev/blob/e2d7fd3625fb021c539348d7ad06121e871a8ebd/packages/lit-dev-tools/src/generate-api-docs-data.ts#L730

AndrewJakubowicz:
Bumping this onto my radar. `@link` is better for editors. VS code will allow clicking on `@link` and will jump to definition. Whilst `[[ ]]` renders directly in the editor without any possible user interaction.

AndrewJakubowicz:
Note, links in `@param` don't work.

AndrewJakubowicz:
Original issue is resolved.

> Note, links in `@param` don't work.

This may still be an issue, but should have a separate issue.

PR #629, opened 2021-12-21
Add @link and @linkcode support in generated API docs

AndrewJakubowicz:
Addresses issue https://github.com/lit/lit.dev/issues/503

With this change, we can generate API docs that support the `@link` and `@linkcode` jsdoc attributes.

Symbol resolution is unchanged, and I've added an additional regex to capture these two new attributes. Further I've also modified the backtick handling so it matches what the IDE renders on hover.

### Testing

 - Manual integration testing done against the following Lit core PR: https://github.com/lit/lit/pull/2370. See screenshots in that PR for why `@link` and `@linkcode` may be preferable.
 - Unit tested against supported syntactic structures.
 - Github Action checks that pull requests don't change the generated API docs without committing the change. This helped me ensure that this change is completely backwards compatible. (Thanks aomarks for that test!)

### Caveats

This PR doesn't exhaustively handle all `@link` syntactic structures mentioned on [jsdoc.app](https://jsdoc.app/tags-inline-link.html). For example we do not handle: `[link text]{@link namepathOrURL}` or `{@link namepathOrURL|link text}`. This was done intentionally as VsCode doesn't support these structures.

`[link text]{@link ApiDocsTransformer}` renders as:
Screen Shot 2021-12-22 at 12 04 09 PM

`{@link ApiDocsTransformer|label}` renders with the pipe in the label:
Screen Shot 2021-12-22 at 12 06 27 PM

### DIFF
--- a/packages/lit-dev-tools-cjs/src/api-docs/transformer.spec.ts
+++ b/packages/lit-dev-tools-cjs/src/api-docs/transformer.spec.ts
@@ -8,7 +8,7 @@ import {test} from 'uvu';
 import * as assert from 'uvu/assert';
 import {linkifySymbolsInCommentsBuilder} from './transformer.js';
 
-const symbolMapFixture1 = {
+const symbolMap = {
   $LitElement: [
     {
       page: 'LitElement',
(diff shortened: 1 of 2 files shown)


aomarks (approved):
Awesome!

PR #637, opened 2022-01-05
Update language around directives.

arthurevans:
- Update language in custom-directives.md to match new language in directives.md (i.e., a directive can be a simple function).
- Add note about "pay-as-you-go" nature of built-in directives.
-

### DIFF
--- a/packages/lit-dev-content/site/docs/templates/custom-directives.md
+++ b/packages/lit-dev-content/site/docs/templates/custom-directives.md
@@ -17,7 +17,7 @@ html`<div>
   </div>`
 ```
 
-While Lit ships with a number of [built-in directives](/docs/templates/directives/) like [`repeat()`](/docs/templates/directives/#repeat) and [`cache()`](/docs/templates/directives/#cache), users can author their own custom directives. There are two kinds of directive:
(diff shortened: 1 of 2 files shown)


graynorton (approved):
lgtm

packages/lit-dev-content/site/docs/templates/custom-directives.md
```diff
@@ -9,17 +9,37 @@ versionLinks:
...
+Directives are functions that can extend Lit by customizing how an expression renders. Using a directive in your template is as simple as calling a function:
 
 ```js
 html`<div>
      ${fancyDirective('some text')}
   </div>`
 ```
 
-However, instead of simply _returning_ a value to render, a directive gets special access to the underlying DOM associated with its expression. And a directive instance is persisted across multiple renders so it can maintain state. A directive can even update the DOM asynchronously, outside of the main update cycle.
+While Lit ships with a number of [built-in directives](/docs/templates/directives/) like [`repeat()`](/docs/templates/directives/#repeat) and [`cache()`](/docs/templates/directives/#cache), users can author their own custom directives. There are two kinds of directive:
```
aomarks, line 20:
```suggestion
While Lit ships with a number of [built-in directives](/docs/templates/directives/) like [`repeat()`](/docs/templates/directives/#repeat) and [`cache()`](/docs/templates/directives/#cache), users can author their own custom directives. There are two kinds of directives:
```

aomarks (approved):
Just the 1 missing `s` typo

PR #644, opened 2022-01-06
Prevent recursive processing of js/ directory in samples generator

aomarks:
The TS -> JS samples generator emits to the `js/` directory, but we were accidentally including that in the list of paths we watch, triggering an infinite loop of endless nested `js/` folders.


augustjk (approved):
LGTM!

PR #647, opened 2022-01-07
Make API docs analysis more robust

aomarks:
- Only bootstrap lit and its dependencies. We were wasting time installing e.g. test dependencies, plus this seemed to be causing some kind of race condition on some of our machines related to installing playwright. Now we shouldn't be installing playwright at all.

- Do a shallow clone of just the one commit we need for each repo, instead of fetching the full history. Much faster clone step.

- Automatically do a fresh clone/setup for each repo if we didn't manage to complete installation last time (previously we would just check if node_modules/ existed, but that could exist even if some subsequent setup step failed). We do this by writing a special file containing the commit SHA only if clone/setup succeeded. Otherwise we won't see this file, and will assume we need a fresh clone/setup.

- Automatically do a fresh clone/setup if the SHA changed. Previously the build would fail and the user had to delete the repo manually.

### DIFF
--- a/packages/lit-dev-tools-cjs/src/api-docs/generate.ts
+++ b/packages/lit-dev-tools-cjs/src/api-docs/generate.ts
@@ -36,73 +36,60 @@ const fileExists = async (path: string): Promise<boolean> => {
 };
 
 /**
- * Return the SHA of the given commit reference.
+ * Clone the given Git repo URL at the given commit into the given directory.
  */
-const shaOfReference = async (
-  gitDir: string,
-  reference: string
-): Promise<string> => {
-  const {stdout} = await execFileAsync(
+const clone = async ({repo, gitDir, commit}: ApiDocsConfig) => {
+  console.log(`cloning git repo ${repo} to ${gitDir} at ${commit}`);
+  // This is the leanest way to shallow fetch just one commit without fetching
+  // any other git history.
+  await execFileAsync('git', ['init', gitDir]);
+  await execFileAsync(
     'git',
-    ['show', '-s', '--format=%H', reference],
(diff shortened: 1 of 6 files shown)


packages/lit-dev-tools-cjs/src/api-docs/generate.ts
```diff
@@ -36,73 +36,60 @@ const fileExists = async (path: string): Promise<boolean> => {
...
-  await execFileAsync('npm', ['ci'], {cwd: dir});
-  for (const {cmd, args} of extraCommands ?? []) {
-    console.log(`running ${cmd} ${args.join(' ')} in ${dir}`);
-    await execFileAsync(cmd, args, {cwd: dir});
+const setup = async (config: ApiDocsConfig) => {
+  console.log(`running npm ci in ${config.gitDir}`);
+  await execFileAsync('npm', ['ci'], {cwd: config.gitDir});
+  for (const {cmd, args} of config.extraSetupCommands ?? []) {
+    console.log(`running ${cmd} ${args.join(' ')} in ${config.gitDir}`);
+    await execFileAsync(cmd, args, {cwd: config.gitDir});
```
AndrewJakubowicz, line 67: This is really nice!

AndrewJakubowicz (approved):
Code looks great. I also checked this in and ran a bunch of different build tests.

I tested a clean install (without any repos). Also tested deleting one, and running the build a second time. Saw the reusing message. Also built the site and manually checked.

Great improvement!

augustjk (approved):
Echoing @AndrewJakubowicz's results above. Came across no issues on installs/builds!

PR #648, opened 2022-01-07
Make JS sample generation more robust

aomarks:
Previously we used chokidar (a filesytem watcher) to find all of the files that needed to be symlinked from `samples/*` into `samples/js`. We did this both in the normal one-time mode, and in `--watch` mode.

In one-time mode, we waited for chokidar's "ready" event, and then assumed that all files would have been detected. Although this does seem to be what the documentation suggests the "ready" event means (https://github.com/paulmillr/chokidar#api), we were seeing cases on one of our machines where only a subset of files were getting detected. I suspect this is because on some machines, the "ready" event is not a reliable indicator that all files have been found.

We now instead do a more straightforward glob match for files using `fast-glob` when we're in the normal one-time mode, which should avoid any weirdness around chokidar. We still use chokidar when in `--watch` mode.

### DIFF
--- a/packages/lit-dev-tools-esm/src/generate-js-samples.ts
+++ b/packages/lit-dev-tools-esm/src/generate-js-samples.ts
@@ -23,6 +23,7 @@ import {
   BLANK_LINE_PLACEHOLDER_COMMENT_REGEXP,
 } from '@lit/ts-transformers';
 import prettier from 'prettier';
+import fastGlob from 'fast-glob';
 
 import type {ProjectManifest} from 'playground-elements/shared/worker-api';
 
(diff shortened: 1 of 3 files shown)


augustjk (approved):
I agree with your suspicion about the `ready` event from `chokidar`. This change works great for the one-time run!

packages/lit-dev-tools-esm/src/generate-js-samples.ts
```diff
@@ -229,10 +194,46 @@ const tsCompileOpts: InvokeTypeScriptOpts = {
   },
 };
 
+const nonTsFileGlobs = [
+  '**/*',
```
AndrewJakubowicz, line 198: Non blocking question. This was `'.'`. Are the two interchangeable?

AndrewJakubowicz (approved):

packages/lit-dev-tools-esm/src/generate-js-samples.ts
aomarks: I found that chokidar understood `.`, but `fast-glob` didn't. Both understand `**/*` though.

PR #653, opened 2022-01-14
Update sample config for localization

augustjk:
The existing sample was missing `output.outputDir` which is always a required field for runtime mode so the cli would error when using provided sample as is.


aomarks (approved):
Nice fix!

PR #655, opened 2022-01-19
Multiple Tutorials Infra

e111077:
| PR# | Parent | Child |
| -- | -- | -- |
| 2/4 | #660 | #661 |

## Added

* Added the ability to add multiple tutorials
* A data pipeline creates and consumes tutorial manifests
* No changes to how `/tutorial/` looks to users

## Particular items that need review

* [ ] Data pipeline architecture
  * [ ] packages/lit-dev-content/samples/tutorials/utils.cjs.ts
  * [ ] packages/lit-dev-content/samples/tutorials/intro-to-lit/11ty-data.cjs.ts
* [ ] Component implementation
  * [ ] packages/lit-dev-content/src/components/litdev-tutorial.ts
    * Manifest fetching is now async

## Extra

Use this link to see the diff without the shuffling of files around.

https://github.com/lit/lit.dev/pull/655/commits/bcc793c59c46d6ee82921de0afb78d0801179adb..18afc98e3b0d3fcb7541a85aade3f550ef3b5f8e


e111077:
@arthurevans I'd particularly like a look at the catalog page's categories & descriptions.

https://pr655-d4b465d---lit-dev-5ftespv5na-uc.a.run.app/tutorials/

arthurevans:
I like the descriptions you have here. On the categories—I think "Codelabs" is a very Google-specific term that's not particularly clear to folks who don't work for Google developer relations, and few people would deduce the "Tutorials/Codelabs" distinction from the names. I'm inclined to make the titles more transparent. For example:

- Learn Lit
- Build Projects

It's also possible that we may want to rearrange these a little bit over time. If we end up with a lot of feature tutorials, we might want to divide them into more basic and more advanced categories.

Along the same lines, we might want to start simple by not displaying beginning/advanced and the estimated time to complete on the cards. It seems to me that the first section should probably be mostly beginning stuff (even "advanced templating" isn't likely to be particularly advanced). The project-based section may vary more in complexity, though.

e111077:
@arthurevans Learn Lit & **Build With Lit**?

packages/lit-dev-content/samples/tutorials/intro-to-lit/11ty-data.cjs.ts
```diff
@@ -0,0 +1,88 @@
...
+import {TutorialData, TutorialManifest} from '../utils.cjs.js';
+import path from 'path';
+
+/**
+ * Generates a tutorial data object for this tutorial's catalog card. For more
+ * options, see the TutorialData type in /samples/tutorials/utils.cjs.ts
+ *
+ * @returns Tutorial Data Object that describes this tutorial for the catalog.
+ */
+export const getData = async (): Promise<TutorialData> => {
```
aomarks, line 33: Could this be a JSON file, e.g. "lit-tutorial-metadata.json", instead of a node script? Seems like it would be simpler and more amenable to easily authoring tutorials (e.g. through the playground UI) in the future.

Not quite seeing why we need promises and the path library here. (Nothing is actually async, and the location seems like it could be inferred by whatever script is crawling through these directories.)

e111077:
Alternatively Learn & Build

packages/lit-dev-content/samples/tutorials/utils.cjs.ts
```diff
@@ -0,0 +1,77 @@
+import fs from 'fs/promises';
```
aomarks, line 1: I think we should move all eleventy-related/node scripts related to tutorials into `packages/lit-dev-tools-cjs`. That's where we've putting eleventy plugins/script.

It seems like putting the eleventynode side of this inside of the `samples/tutorials` directory is complicating things a bit -- because so far, the `samples/` directory has only been for the actual samples content, rather than the eleventy infrastructure for it.

packages/lit-dev-content/site/tutorials/index.html
```diff
@@ -0,0 +1,126 @@
+---
+title: Tutorials
```
aomarks, line 2: Is there a way to access this from the preview?

aomarks (changes requested):
A start on the review, haven't looked over everything yet.

High level request: do you think we could split the tutorials index page work out into a separate PR? It seems like we could land updates to the tutorial viewer and file layout independently.

packages/lit-dev-content/samples/tutorials/intro-to-lit/11ty-data.cjs.ts
```diff
@@ -0,0 +1,88 @@
...
+ * 2. Do a call to `await loadTutorialData('[tutorial-directory-name]')` in
+ *    the order that you want it to appear in
+ *    `/site/tutorials/tutorials.11tydata.js`
+ */
+import {TutorialData, TutorialManifest} from '../utils.cjs.js';
+import path from 'path';
+
+/**
+ * Generates a tutorial data object for this tutorial's catalog card. For more
+ * options, see the TutorialData type in /samples/tutorials/utils.cjs.ts
```
AndrewJakubowicz, line 29: Nit: Should be able to directly link with:

```suggestion
 * options, see {@linkcode TutorialData}.
```

packages/lit-dev-content/samples/tutorials/intro-to-lit/11ty-data.cjs.ts
e111077: I initially wrote this stuff as JSON but went with node because JSON format is really gross to deal with when authoring. 
* it's annoying to write multi-line strings for description
* you have to hunt down trailing commas
* you have to escape your quotes
* and usng TS:
  1. tells you if you are missing any required fields with type checking
  2. makes sure you're using the right types
  3. makes it easier to refactor the JSON format

I agree though that the filename probably isnt the best. What do you think I should change it to?

I can pull out the location as well and do that in the 11ty data file that consumes this. Can also remove the async. Idea behind it was that I was thinking i could use `fs` somehow to automate steps or decrease the number of project.jsons i'd have to write.

It's currently unnecessary, so would you still want me to remove the async?

aomarks: The problem with a JS file that I'm thinking of is that we wouldn't be able to safely evaluate it on the client. The use case I'm imagining is creating a tutorial as a gist in the playground editor. To me it seems natural to put metadata for that tutorial in a file like "manifest.json" or something. We could then read all the data we need out of that on the client side when you point the tutorial viewer at that gist. If it's a .js file, we would have to `eval` it, which would be unsafe.

packages/lit-dev-content/samples/tutorials/utils.cjs.ts
e111077: Reasoning behind this was to centralize and decrease the number of locations you'd need to visit to author a new tutorial since Arthur was worried about complicating the authoring workflow. Currently to create one the locations you need to edit are:

1. `lit-dev-content/samples/tutorials/[dirname]`
2. `lit-dev-content/site/tutorials/tutorials.11tydata.js`
3. `lit-dev-content/site/tutorials/content/[dirname]`

My worries about moving it to `lit-dev-tools-cjs` are:

1. it adds another place the author must edit
2. These cjs files contain content themselves not tools as they are extensions of (`lit-dev-content/site/__data/*.js` which are also cjs files)

I agree with the TS causing more complexity. Instead I can make these `.js` files and add `//@ts-check` to still provide the TS typechecking for authorship. That should knock out the extra tsconfig.json and any changes I made to the build scripts

packages/lit-dev-content/site/tutorials/index.html
e111077: just append `/tutorials/` to the url. No way to navigate through the site currently

packages/lit-dev-content/samples/tutorials/intro-to-lit/11ty-data.cjs.ts
e111077: TIL

packages/lit-dev-content/samples/tutorials/utils.cjs.ts
aomarks: Sorry, I didn't mean that data you would need to edit to make a tutorial should go there, just any infrastructure scripts/eleventy plugins etc. I too like the idea that all tutorial content should be in the same directory.

packages/lit-dev-server/src/middleware/tutorials-middleware.ts
```diff
@@ -0,0 +1,40 @@
...
+     * where we are currently rendering the markdown to html so not
+     * /tutorials/content/*
+     */
+    if (
+      path.startsWith('/tutorials/') &&
+      !path.startsWith('/tutorials/content/') &&
+      path !== '/tutorials/'
+    ) {
+      // Get the contents of /tutorials/view/
+      const contents = await fs.readFile(
```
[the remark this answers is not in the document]
e111077: TIL I thought this would change the URL and it doesn't! Thanks for the review!

packages/lit-dev-content/samples/tutorials/utils.cjs.ts
aomarks: Clarifying some of my earlier comments...

I'm thinking we should mirror how the Playground examples list generation works today, which is this:

1. Each `samples/examples` directory has a `project.json` which has the title/description/section metadata in it. Example: https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/samples/examples/directive-async-replace/project.json

2. There is a script in `lit-dev-content/site/_data/playground_examples.js` which scans for all of the `project.json` files and returns the data for Eleventy: https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/site/_data/playground_examples.js

3. The Playground page reads the `playground_examples` data for templating: https://github.com/lit/lit.dev/blob/787b789b41f1fbb360a05c296a232b1963f940c9/packages/lit-dev-content/site/playground/index.html#L22

The tutorials generation is really similar to this, so I think we should mirror the same pattern. Specifically:

1. Each `samples/tutorials` directory has a `tutorial.json` which contains the list of steps and other metadata related to the index view.

2. Add a new `lit-dev-content/site/_data/tutorials.js` script.

3. Consume the `tutorials` variable in the template for the new tutorials page.

Benefits of this:

- Simpler infrastructure. We won't need the second `tsc` pipeline in the samples directory.

- I think a declarative JSON approach for tutorial metadata is probably than each tutorial having its own Node script:

  - If we later support tutorials in gists, then it wouldn't be possible to read the metadata out of a `.js` files, because we'd need to `eval` it, which would be unsafe. 

  - It feels more conventional/natural. It's similar to the existing Playground `project.json` files we have.

  - I hear you on the type checking concern, but:
    - We'll be adding tutorials quite slowly so the risk seems pretty low to begin with.
    - We can include a validity check in the step 2 script. [typescript-json-schema](https://github.com/YousefED/typescript-json-schema) could be a good way to do this, because it will generate a JSON schema automatically from the TypeScript interface. We could then automatically check validity from the schema. JSON schema definitions are also understood by VSCode using the special `"$schema" property.

aomarks (changes requested):

packages/lit-dev-content/package.json
```diff
@@ -10,18 +10,19 @@
     "postinstall": "ls site/fonts/manrope* > /dev/null || npm run fonts:manrope",
     "fonts:manrope": "rm -rf site/fonts temp && mkdir -p site/fonts && git clone https://github.com/sharanda/manrope.git temp/manrope && cd temp/manrope && git checkout 9ffbc349f4659065b62f780fe6e9d5a93518bd95 && cp fonts/web/*.woff2 ../../site/fonts/ && cd ../.. && rm -rf temp",
     "build": "npm run build:not-site && npm run build:site",
-    "build:not-site": "npm run build:ts && npm run build:rollup && npm run build:samples",
+    "build:not-site": "npm run build:samples && npm run build:ts && npm run build:rollup",
```
aomarks, line 13: Can the changes in this file be reverted now?

packages/lit-dev-content/samples/tsconfig.json
```diff
@@ -13,6 +13,8 @@
   },
   "include": ["./**/*.ts"],
   "exclude": [
+    // These must be commonjs as they are imported by 11ty data sources
```
aomarks, line 16: Can this be reverted now too?

packages/lit-dev-content/samples/tutorials/advanced-templating/11ty-data.cjs.ts
```diff
@@ -0,0 +1,32 @@
...
+    difficulty: "Advanced",
+    size: "large",
+    duration: 50,
+    location: path.basename(__dirname),
+    category: "Tutorial",
+    imgSrc: 'images/logo-whitebg-padded-1600x800.png',
+    imgAlt: 'This is the Lit icon',
+    description: `These templates are gonna be real advanced and I REALLY mean
+    it yo! Like give it a try!<br/><br/>
+    Now look at this a PAGEBREAK oooOOOOOh. I told you this templating is gonna
```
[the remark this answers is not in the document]
e111077: Moved over to using markdown, advanced templating should still work

packages/lit-dev-content/samples/tutorials/intro-to-lit/manifest.json
```diff
@@ -0,0 +1,33 @@
+{
```
aomarks, line 1: WDYT about `tutorial.json`? Just makes it a little more obvious what it's for, vs `manifest` which is a little generic.

packages/lit-dev-content/site/tutorial/index.html
```diff
@@ -62,4 +63,4 @@
   </playground-preview>
 </div>
 
-{% endblock %}
+{% endblock %}
```
aomarks, line 66: Nit, restore trailing newline.

packages/lit-dev-content/samples/tutorials/intro-to-lit/manifest.json
```diff
@@ -0,0 +1,33 @@
...
+      "title": "Template logic"
+    },
+    {
+      "title": "Styles"
+    },
+    {
+      "title": "Finishing touches"
+    }
+  ]
+}
```
aomarks, line 33: Nit: trailing newline.

Files with trailing newlines work better with unix tools, I think it's a good default.

(Note there's an "Insert trailing newline" vscode setting to always do this.)

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -4,32 +4,51 @@
...
 import './litdev-playground-change-guard.js';
 import './litdev-icon-button.js';
+import {Task, TaskStatus} from '@lit-labs/task';
 
-interface ExpandedTutorialStep extends TutorialStep {
+export interface TutorialStep {
+  title: string;
+}
+
+/* TODO(e111077): find a way to make a single source of truth for this.
```
aomarks, line 33: Can remove this comment now I think?

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -4,32 +4,51 @@
...
+interface ExpandedTutorialStep {
   idx: number;
   url: string;
   htmlSrc: string;
   projectSrcBefore: string;
   projectSrcAfter: string;
 }
 
+const samplesRoot =
+  getCodeLanguagePreference() === 'ts' ? '/samples' : '/samples/js';
```
aomarks, line 50: The language preference can change at any time when the user clicks the toggle from within the tutorial, so does this change break the ability to switch between languages without a reload?

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +93,217 @@ export class LitDevTutorial extends LitElement {
...
+  private _manifest: TutorialManifest = {steps: [{title: ''}]};
+
   /**
    * Whether the tutorial is currently in its "solved" state.
    */
   private _solved = false;
 
+  private get _projectLocation() {
+    // TODO(e111077): delete this once we go public with tutorials catalog
+    if (window.location.pathname === '/tutorial/') {
```
aomarks, line 109: WDYT about using a similar URL-testing approach as this for the `use-old-url` and `show-catalog-button` options too? That would mean we don't need to add the new `view.html` file, which is basically identical to `tutorial/index.html` except that it changes those attributes. We could instead have the middleware serve `tutorial/index.html` for the new URL scheme. They wouldn't even need to be reactive properties, since they can't change.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +93,217 @@ export class LitDevTutorial extends LitElement {
...
+      return 'intro-to-lit';
+    }
+    return window.location.pathname.replace('/tutorials/', '');
+  }
+
+  /**
+   * Playground project element generated from this.project
```
aomarks, line 116:
```suggestion
   * Playground project element discovered using the id from this.project
```

("generated" kind of made it sound like it was being created here)

packages/lit-dev-content/src/icons/back-arrow-icon.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {svg} from 'lit';
+
+// Source: https://material.io/resources/icons/?icon=arrow_back
+export const backArrowIcon = svg`
```
aomarks, line 10: I know this is based on the other icon modules here that I wrote (sorry!), but I think these should all actually be exporting `html` tagged templates, not `svg`, right? I think `svg` is only needed for creating elements inside the svg namespace. But here since we're returning an `<svg>` element, which is HTML namespace, so we want `html`?

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +93,217 @@ export class LitDevTutorial extends LitElement {
...
+   */
+  private _manifestTask = new Task(
+    this,
+    async ([projectLocation]) => {
+      const manifestRes = await fetch(
+        `${samplesRoot}/js/tutorials/${projectLocation}/manifest.json`
+      );
+
+      const manifest = (await manifestRes.json()) as TutorialManifest;
+      this._manifest = manifest;
```
aomarks, line 154: I think there could be a race condition here?

1. _projectLocation changes to A
2. task updates and starts fetching manifest A
3. _projectLocation changes to B
4. task updates and starts fetching manifest B
5. manifest B comes back, this._manifest is set to B
6. manifest A comes back, this._manifest is set to A <-- bad!

Maybe the solution is just to check `projectLocation === this._projectLocation` before re-assigning `this._manifest`?

It sort of feels like setting state from a task body might generally be an anti-pattern, because of cases like this, wdyt? (TBH this is the first time I've looked at the task API, so I don't know what the good usage patterns are yet)

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +93,217 @@ export class LitDevTutorial extends LitElement {
...
+    async ([idx]) => {
+      this._solved = false;
+      const active = this._info;
+
+      // Either use the preloaded html or fetch the new HTML (going back)
+      const html =
+        this._preloadedHtml?.idx === idx
+          ? await this._preloadedHtml.promise
+          : await this._fetchHtml(active.htmlSrc);
+      this._setProjectSrc(active.projectSrcBefore);
```
aomarks, line 176: Same comment as above. Seems a bit racey?

aomarks (changes requested):

packages/lit-dev-content/src/components/litdev-tutorial.ts
e111077: I really don't like url testing I personally would like to make every URL part user-inputted. I feel like it ties this component too close to this specific URL and section of this app and makes for potential refactors more painful as the URL dependencies are distributed across both frontend and backend. I'd rather just stick with this for now since we're going to remove it in a few weeks

aomarks: Sounds good.

> I personally would like to make every URL part user-inputted.

Can you clarify?

e111077:
Tried to prevent a race condition by essentially implementing my own `onComplete` in `willUpdate`.

See lit/lit#2454

packages/lit-dev-content/src/components/litdev-tutorial.ts
e111077: basically make the element more generic and make every reference to every URL base an `@property`. like `/samples/js/tutorials/tutorial.json` could be `${this.jsSamplesPath}/${this.tutorialsPath}/tutorial.json` and same thing for lots of things in `_idxToInfo`

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +86,252 @@ export class LitDevTutorial extends LitElement {
...
     return this;
   }
 
+  willUpdate(): void {
+    const manifestTaskValue = this._manifestTask.value;
+
+    /**
+     * When the manifest task has just finished resolving, read the URL
+     */
+    if (
```
aomarks, line 201: This section is kind of gnarly, but I don't have an obvious suggestion to clean it up! Seems like the complexity is coming in because of having to synchronize multiple tasks that can resolve independently along with some global state like the URL. Maybe the discussion in https://github.com/lit/lit/issues/2454 will be helpful for a future refactoring!

Anyway, seems like no action needed right now. LGTM!

aomarks (approved):
Nice!

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +86,252 @@ export class LitDevTutorial extends LitElement {
...
+    ) {
+      this._oldManifestValue = manifestTaskValue;
+      this._readUrl();
+    }
+
+    const htmlTaskValue = this._htmlTask.value;
+
+    /**
+     * When the html task has just finished resolving fetch the next step's html
+     */
```
AndrewJakubowicz, line 213: Nit: Comment format. [Prefer `//`](https://google.github.io/styleguide/tsguide.html#jsdoc-vs-comments)

AndrewJakubowicz (commented):
This is great!

Some minor nits (outside the golden path) form manual testing so can be punted from this PR.
https://pr655-606467e---lit-dev-5ftespv5na-uc.a.run.app/tutorials/intro-to-lit/ works great, however an invalid path isn't handled.

Error case 1, tutorial doesn't exist, i.e. https://pr655-606467e---lit-dev-5ftespv5na-uc.a.run.app/tutorials/not-exist/

Error case 2, invalid extension returns strange content security page. https://pr655-606467e---lit-dev-5ftespv5na-uc.a.run.app/tutorials/index.html

Both can maybe be handled by checking for tutorial existence.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -69,89 +86,252 @@ export class LitDevTutorial extends LitElement {
...
    * Whether the tutorial is currently in its "solved" state.
    */
   private _solved = false;
 
+  private get _projectLocation() {
+    // TODO(e111077): delete this once we go public with tutorials catalog
+    if (window.location.pathname === '/tutorial/') {
+      return 'intro-to-lit';
+    }
+    return window.location.pathname.replace('/tutorials/', '');
```
AndrewJakubowicz, line 106: Minor nit: Test and remove trailing `/`.

Because the server adds trailing slashes, this currently behaves like so:

```
// While on the page https://pr655-606467e---lit-dev-5ftespv5na-uc.a.run.app/tutorials/intro-to-lit/
// tested in console:
window.location.pathname.replace('/tutorials/', '')
// Returns: 'intro-to-lit/'
```

This results in double slashes in requests, i.e.
https://pr655-606467e---lit-dev-5ftespv5na-uc.a.run.app/samples/tutorials/intro-to-lit//tutorial.json

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -255,53 +438,53 @@ export class LitDevTutorial extends LitElement {
...
+  private _idxToInfo(idx: number): ExpandedTutorialStep {
+    const slug = `${idx}`.length == 1 ? `0${idx}` : `${idx}`;
+
+    const firstUrl = this.useOldUrl
+      ? '/tutorial/'
+      : `/tutorials/${this._projectLocation}`;
+
+    const nextUrl = this.useOldUrl
+      ? `/tutorial/#${slug}`
+      : `/tutorials/${this._projectLocation}#${slug}`;
```
AndrewJakubowicz, line 480: May need to add a trailing slash here between _projectLocation and slug. Although works either way so maybe doesn't matter.

AndrewJakubowicz (approved):
Nice!

e111077:
@AndrewJakubowicz I display an error for `1.` but am going to have to tackle `2.` in a later PR

packages/lit-dev-content/src/components/litdev-tutorial.ts
e111077: Will leave as so since every other url is written without the trailing `/`

aomarks: We should include it, because the canonical URL has a slash. If you click next, there won't be a slash, but then if you reload, it will appear (and there will be a 301 behind the scenes).

AndrewJakubowicz:
That sounds good. I think error 2 is unrelated to tutorials and more related to CSP.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -152,11 +165,17 @@ export class LitDevTutorial extends LitElement {
...
+      this._manifestLoadError = '';
       const manifestRes = await fetch(
         `${this._samplesRoot}/tutorials/${projectLocation}/tutorial.json`
       );
 
-      return (await manifestRes.json()) as TutorialManifest;
+      if (manifestRes.ok) {
+        return (await manifestRes.json()) as TutorialManifest;
+      } else {
+        this._manifestLoadError = `Could not fetch tutorial manifest. Invalid URL /${projectLocation}/`;
```
aomarks, line 176: Doesn't this introduce a similar race condition to before? If request A fails but is slow, and request B succeeds but is fast, then the error will get displayed even though it's no longer relevant.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -460,6 +492,15 @@ export class LitDevTutorial extends LitElement {
     return isNaN(idx) ? undefined : idx;
   }
 
+  /**
+   * Turns an index into a 2-length stringified number for the slug.
+   * @param idx index to slugify
+   * @returns A 2-length stringified number
+   */
+  private _idxToSlug(idx: number): string {
+    return `${idx}`.length == 1 ? `0${idx}` : `${idx}`;
```
aomarks, line 501: I think this is the same as:

```js
String(idx).padStart(2, "0");
```

e111077: darn i merged. Will spin up another PR right now to address

packages/lit-dev-content/src/components/litdev-tutorial.ts
e111077: In this case the UI makes navigating to these states unlikely. This will only happen if we forgot to deploy some of these files. Anyway, I'd prefer this failure to have high visibility WDYT?

e111077: tackled this in #667

PR #660, opened 2022-01-22
Implement tutorial server redirect middleware

e111077:
| PR# | Parent | Child |
| -- | -- | -- |
| 1/4 | none | #655 |

## Added

* A redirect from `/tutorials/*` to `/tutorials/view.html`

## Particular items that need review

* [ ] Tutorial middleware
  * [ ] packages/lit-dev-server/src/middleware/tutorials-middleware.ts

### DIFF
--- a/packages/lit-dev-server/src/middleware/tutorials-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/tutorials-middleware.ts
@@ -1,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import type Koa from 'koa';
+
+/**
+ * Koa Middleware to serve '/tutorials/view.html' from '/tutorials/*'.
+ *
+ * @param staticRoot Root of the static server
+ */
+export const tutorialsMiddleware = (): Koa.Middleware => async (ctx, next) => {
+  const path = ctx.request.path;
+
+  /**
+   * We want to intercept /tutorials/* but not /tutorials/ itself or the place
+   * where we are currently rendering the markdown to html so not
+   * /tutorials/content/*
+   */
+  if (
+    path.startsWith('/tutorials/') &&
+    !path.startsWith('/tutorials/content/') &&
+    path !== '/tutorials/'
+  ) {
+    // Serve /tutorials/view/
+    ctx.path = '/tutorials/view/index.html';
+  }
+  await next();
+};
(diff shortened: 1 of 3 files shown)


packages/lit-dev-server/src/middleware/tutorials-middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+   * We want to intercept /tutorials/* but not /tutorials/ itself or the place
+   * where we are currently rendering the markdown to html so not
+   * /tutorials/content/*
+   */
+  if (
+    path.startsWith('/tutorials/') &&
+    !path.startsWith('/tutorials/content/') &&
+    path !== '/tutorials/'
+  ) {
+    // Serve /tutorials/view/
```
aomarks, line 27: Might be a slightly unclear comment, since that's neither the URL nor the file. Probably don't need a comment IMO.

packages/lit-dev-server/src/middleware/tutorials-middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import type Koa from 'koa';
+
+/**
+ * Koa Middleware to serve '/tutorials/view.html' from '/tutorials/*'.
+ *
+ * @param staticRoot Root of the static server
```
aomarks, line 12: Remove comment since the parameter is gone.

aomarks (changes requested):

packages/lit-dev-server/src/middleware/tutorials-middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const tutorialsMiddleware = (): Koa.Middleware => async (ctx, next) => {
+  const path = ctx.request.path;
+
+  /**
+   * We want to intercept /tutorials/* but not /tutorials/ itself or the place
```
aomarks, line 18: Very small nit, but I would usually use `// ` comments for an inline comment multi-line comments like this, because I don't think the `/**` jsdoc style comment means anything here, because it's not associated with a symbol like a variable, method etc.

e111077: oh lol It was the only way I could get the auto-comment-next-line feature to work. Forgot to remove the second `*`

aomarks: Actually I meant multi-line `//` comments.

(This is totally just a style preference, not blocking, but since we started the conversation :))

The pattern I've been following is `/**` style jsdoc comments for annotating an API, and `//` (including multi-line) when annotating implementation.

This seems to match the TypeScript style guide: https://google.github.io/styleguide/tsguide.html#comments-when-calling-a-function. And also I think it also matches we do in most of our other libraries (e.g. see https://github.com/lit/lit/blob/7453e365000e6a289c139cf7e175a4742296333d/packages/lit-html/src/lit-html.ts#L174)

aomarks (approved):

PR #665, opened 2022-01-26
Disambiguate static HTMLSlotElement syntax and slot instance.

AndrewJakubowicz:
Fix as `HTMLSlot` doesn't mean anything. It's neither HTMLSlotElement
which could mislead the reader into thinking `assignedElements`
is static.

Documentation in [section above](https://lit.dev/docs/components/shadow-dom/#:~:text=slot.assignedNodes%20or%20slot.assignedElements) uses `slot` as the implicit reference to the
HTMLSlotElement instance.

Thus this change introduces consistent naming.

Thank you!


augustjk (approved):
LGTM!

PR #667, opened 2022-01-26
Implement litdev-tutorial suggestions from #665

e111077:
Implement missed suggestions from #655.

uhh idk also what's up with the diff. Some things are a bit off

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-tutorial.ts
+++ b/packages/lit-dev-content/src/components/litdev-tutorial.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-import {LitElement, html} from 'lit';
+import {LitElement, html, nothing} from 'lit';
 import {property, state} from 'lit/decorators.js';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
 import {when} from 'lit/directives/when.js';
@@ -86,21 +86,21 @@ export class LitDevTutorial extends LitElement {
   @state()
   private _preloadedHtml?: {
     idx: number;
-    promise: Promise<{html: string; isOk: boolean}>;
+    promise: Promise<string>;
   };
 
   /**
    * Tutorial manifest json file
    */
   private get _manifest(): TutorialManifest {
-    return this._manifestTask.value?.manifest ?? {steps: [{title: ''}]};
+    return this._manifestTask.value ?? {steps: [{title: ''}]};
   }
 
   /**
    * Message to be displayed if there was an error loading
    */
   private get _manifestLoadError() {
-    return this._manifestTask.value?.isOk === false
+    return this._manifestTask.status === TaskStatus.ERROR
       ? `Could not fetch tutorial manifest. Invalid URL /${this._projectLocation}/`
       : '';
   }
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -80,29 +80,40 @@ export class LitDevTutorial extends LitElement {
...
 
   /**
-   * Tutorial manifest json file
+   * Message to be displayed if there was an error loading
    */
-  private get _manifest(): TutorialManifest {
-    return this._manifestTask.value ?? {steps: [{title: ''}]};
+  private get _htmlLoadError() {
+    return this._htmlTask.value?.isOk === false
+      ? `Could not fetch step description. Invalid URL /${
```
AndrewJakubowicz, line 113: Nit: If a someone encounters this error, it may be useful to include a call to action, since it appears in the tutorial text area. Maybe a link to the tutorials home page? Non blocking.

For example, `https://lit.dev/tutorials/intro-to-lit/#9` is an invalid link that shows code but not step description.
Alternatively I'm uncertain about putting the error message in the content area and having it look similar to content.

e111077: It actually just shows up as `/intro-to-lit/#09` I didn't want to put an actual working link here. Besides I'm pretty sure we'd be more likely to run into this as authors rather than users. 

Probably fits better in a toast or a dialog, but I feel like that's a bunch of effort for MVP. Will write a TODO

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -86,21 +86,21 @@ export class LitDevTutorial extends LitElement {
...
+    return this._manifestTask.value ?? {steps: [{title: ''}]};
   }
 
   /**
    * Message to be displayed if there was an error loading
    */
   private get _manifestLoadError() {
-    return this._manifestTask.value?.isOk === false
+    return this._manifestTask.status === TaskStatus.ERROR
       ? `Could not fetch tutorial manifest. Invalid URL /${this._projectLocation}/`
```
aomarks, line 104: Could you move this error directly into the task `error` render function below? The indirection through the getter doesn't seem to be buying much. Same for `_htmlLoadError`.

packages/lit-dev-content/src/components/litdev-tutorial.ts
e111077: Added

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -322,11 +309,19 @@ export class LitDevTutorial extends LitElement {
           () => this._manifest.steps[this._idx].title
         )}
       </h1>
-      ${this._htmlLoadError || this._manifestLoadError
-        ? html`${this._htmlLoadError}<br />${this._manifestLoadError}`
-        : this._htmlTask.render({
-            complete: (description) => html`${unsafeHTML(description)}`,
-          })}
+      ${this._manifestTask.render({
+        // TODO(e111o77): Move these error messages out of content and into
```
aomarks, line 313:
```suggestion
        // TODO(e111077): Move these error messages out of content and into
```

:D

aomarks (approved):

packages/lit-dev-content/src/components/litdev-tutorial.ts
e111077: HI mY n4ME iS eLl1oTt

AndrewJakubowicz (approved):
LGTM! Awesome!

Issue #666, opened 2022-01-26
Implement 404 page


AndrewJakubowicz:
Currently invalid links take you to a blank 'Not Found' page. i.e: https://lit.dev/invalid/paths/

A 404 page provides an opportunity to guide the user out of the error.

### Completion criteria:

Design/implement a 404 page which guides the user back to the site.

Possible examples:
 - https://web.dev/invalid/test, https://developer.mozilla.org/en-US/invalid/path, both use a 404 page styled with their navigation. This makes it easy to click on a link to back out of the error.

AndrewJakubowicz:
Implemented in #668

PR #668, opened 2022-01-27
Add 404 not found page

AndrewJakubowicz:
Issue: https://github.com/lit/lit.dev/issues/666

Previously 404 status would return "Not found" text without any user action to return to lit.dev flow. This change adds a very lightweight 404 page and not found middleware.

Previous behavior: https://lit.dev/invalid/path
New behavior: https://pr668-bd7a32c---lit-dev-5ftespv5na-uc.a.run.app/invalid/path

### How

 - Installed https://github.com/koajs/send (a transitive dependency of koajs static) for streaming a file.
 - Added a lightweight notfound middleware as a final middleware that intercepts 404 status codes.
 - Light 404.html and 404.css (inspired by the Blog index styles).

### Testing

This was tested manually.

Screenshot:
Screen Shot 2022-01-27 at 9 34 03 AM

### DIFF
--- a/packages/lit-dev-content/site/404.html
+++ b/packages/lit-dev-content/site/404.html
@@ -1,0 +1,18 @@
+---
+title: Not Found
+---
+
+{% extends 'default.html' %}
+
+{% block head %}
+  {% inlinecss "404.css" %}
+{% endblock %}
+
+{% block content %}
+
+<article id="container-404">
+  <h1>404</h1>
+  <p>Sorry, we couldn't find that page.</p>
+</article>
+
+{% endblock %}
--- a/packages/lit-dev-server/src/middleware/notfound-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/notfound-middleware.ts
@@ -12,12 +12,12 @@ import type Koa from 'koa';
  * Creates a Koa middleware that serves a 404.html page on 404 status codes.
  */
 export const notFoundMiddleware =
-  (staticPath: string): Koa.Middleware =>
+  (staticRoot: string): Koa.Middleware =>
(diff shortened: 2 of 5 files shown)


packages/lit-dev-content/site/404.html
```diff
@@ -0,0 +1,18 @@
...
+
+{% extends 'default.html' %}
+
+{% block head %}
+  {% inlinecss "404.css" %}
+{% endblock %}
+
+{% block content %}
+
+<article id="container-404">
```
AndrewJakubowicz, line 13: TIL about [always including a hyphen in id](https://google.github.io/styleguide/htmlcssguide.html#:~:text=always%20include%20a%20hyphen)!

packages/lit-dev-server/src/middleware/notfound-middleware.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import send from 'koa-send';
+
+import type Koa from 'koa';
+
+/**
+ * Creates a Koa middleware that serves a 404.html page on 404 status codes.
+ */
+export const notFoundMiddleware =
+  (staticPath: string): Koa.Middleware =>
```
aomarks, line 15: `staticPath` -> `staticRoot` so it matches the other variable name?

aomarks (approved):
Awesome! Nice simple implementation.

packages/lit-dev-server/src/middleware/notfound-middleware.ts
AndrewJakubowicz: Great call. Done

augustjk (approved):
Nice!

PR #673, opened 2022-02-01
Add a page describing the bundles.

### LINKED ISSUE #674: Create a docs page for release bundles
Lit's release process now creates bundles and pushes them to lit/dist:

https://github.com/lit/lit/blob/d21922f4022e059b6bfbece60c25cd21e011e004/.github/workflows/release.yaml#L56

We need a page describing these bundles and how to use them.

bicknellr:
This page gives a short overview on where the bundles are, how to get them, and how to use them.

- [x] Add a section about CDNs and jsDelivr.
- [x] Wait for bundles to be pushed to the lit/dist repo.
- [x] Make the lit/dist repo public.
- [x] ~~Add a readme to the lit/dist repo or update the "About" panel with links to this section.~~
  - https://github.com/lit/dist/pull/1 will handle this, but I'll wait to merge it until after this PR.

Fixes #674, lit/lit#2272.


AndrewJakubowicz (commented):
This is awesome! Did a read through and it's super great!

Regarding the initial instructions to download via the "tags". Is that setup yet?

bicknellr:
Thanks for taking a look! Yeah, the release workflow in the main repo is currently set up to push to a new tagged commit in the dist repo with the bundles whenever `lit` is published to npm[^1], so GitHub's tags page will automatically list them and show the archive links. It will look like https://github.com/lit/lit/tags, but only versions aligning with the `lit` npm package and instead of e.g. `lit@2.1.2`, it'll just be `v2.1.2`.

[^1]: https://github.com/lit/lit/blob/3aa980ec975414e2ca9976023d20b3de4250eb14/.github/workflows/release.yaml#L83-L84

bicknellr:
Does the site get updated automatically on merge? If so, we should wait until the next release pushes bundles. (And, ideally, another one that doesn't push any also.)

aomarks:
> Does the site get updated automatically on merge? If so, we should wait until the next release pushes bundles. (And, ideally, another one that doesn't push any also.)

Automatic on push to main.

bicknellr:
Ok, thanks, I've added some checkboxes to account for that in the description.

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+The bundles can be found on GitHub in [the lit/dist
+repository](https://github.com/lit/dist). To download a particular version, go
+to [the repository's tags](https://github.com/lit/dist/tags), find the version
+you want, and click either the "zip" or "tar.gz" link to download an archive of
+the bundles in the chosen format.
+
+## Usage
+
+The archive has multiple folders, each containing a bundle and its source map.
+The bundles are standard JavaScript modules with no dependencies - any modern
```
sorvell, line 34: Suggest moving this to be the 2nd sentence in the intro, after `Lit is available...`. It's the most important info.

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+package](https://www.npmjs.com/package/lit) instead of these bundles.** The
+bundles intentionally combine most or all of Lit into a single file. This can
+prevent some build tools from removing the parts of Lit your project doesn't
+use, which means your users may download more code than they strictly need.
+
+</div>
+
+## Download
+
+The bundles can be found on GitHub in [the lit/dist
```
sorvell, line 25: Can we make the first thing here the actual link to the thing you should download (maybe one for min and one for al)?

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+
+For example, if you extracted the bundles into a directory called `lit-bundles`
+next to one of your source files, you could use the `core` bundle in that source
+file like this:
+
+```js
+import {LitElement, html} from './lit-bundles/core/lit-core.min.js';
+```
+
+### Source maps
```
sorvell, line 45: I don't think we need the detailed info here. Just that the bundles include source maps is enough info (e.g. you must also not edit this file, but we don't say that).

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+
+```js
+import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js';
+```
+
+See [jsDelivr's features
+page](https://www.jsdelivr.com/features#:~:text=GitHub%20CDN) for more
+information on how to format the URL.
+
+## Bundle variants
```
sorvell, line 68: Move this into the usage section and remove the detailed info about the files included. Think it's unnecessary.

packages/lit-dev-content/site/docs/libraries/bundles.md
sorvell: Combined organization suggestion based on other feedback:

##Download
 * start with 2 links to most current versions of min and all.
 * describe contents (from variants) and give warning about choosing and using one of them.
 * just say that source maps are included.
 
 ##Usage
  * just show the example usage for each variant and say they are each in their own folder.
  
  ##CDN

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+### Source maps
+
+Each bundle directory includes a source map. If you want to be able to view
+Lit's unminified sources while debugging, keep this file in the same directory
+as its associated bundle and don't rename it. If you update your project's
+bundle, you also need to update its source map.
+
+## CDN
+
+As an alternative to hosting your own copy of Lit's bundles, you can also import
```
sorvell, line 54:
```suggestion
You can also import
```

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+them from CDNs that serve assets directly from repositories hosted on GitHub.
```
sorvell, line 55:
```suggestion
the bundles from CDNs that serve assets directly from repositories hosted on GitHub.
```

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+
+**Don't use more than one bundle in your project.** The bundles each contain
+full copies of overlapping content. If you need to switch bundles, you should
+update every bundle reference in your project to point to the new bundle.
+
+</div>
+
+### `core/`
+
+- `lit-core.min.js` - the `core` bundle
```
sorvell, line 82: Remove file list for this and all. Just seems unnecessary and obvious.

sorvell (changes requested):

arthurevans:
I'm not sure libraries is the right place for this doc. We probably want to put something in the Getting Started section ("Using a CDN"? or "Prebuilt bundles" or something?)

Not sure what @justinfagnani @kevinpschaaf @sorvell think about where this goes, but I thought the main use case here was to let people get started without npm & a build step.

packages/lit-dev-content/site/docs/libraries/bundles.md
justinfagnani: I think we can remove mention of this all together. Our primary published sources have source maps and we don't document it there.

justinfagnani:
> I'm not sure libraries is the right place for this doc. We probably want to put something in the Getting Started section ("Using a CDN"? or "Prebuilt bundles" or something?)

@arthurevans @bicknellr  Can we chat on this today? I was thinking that *Tools and Workflows* could be changed slightly so that *Adding Lit* came earlier in the outline and was changed to encompass "Adding Lit with npm" and "Adding Lit from a CDN" - basically make it an "Installation" page as a first set in the roughly-chronological ordering of workflow steps.

The only question would then be what to do with the "Using Lit" parts of the "Adding Lit" page. And I think the content here would have to be a lot smaller to be a section on a page.

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+---
+
+Lit is available as pre-built, single-file bundles. These bundles are useful for
+situations where you can't use npm or just want to experiment without setting up
+a build process.
+
+<div class="alert alert-warning">
+
+**If you're using npm for client-side dependencies, you should use the [`lit`
+package](https://www.npmjs.com/package/lit) instead of these bundles.** The
```
justinfagnani, line 16: This is where it'd be good to have a canonical "Install from npm" page on lit.dev to link to.

justinfagnani, line 16: Maybe "instead of these bundles" -> "not these bundles" (to avoid seeming to imply that these bundles are a first option)

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+bundles intentionally combine most or all of Lit into a single file. This can
```
justinfagnani, line 17: It's not just build tools, it's also the browser - the browser doesn't load what's not imported. I would cut this to "This can cause your page to load more JavaScript than needed"

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+prevent some build tools from removing the parts of Lit your project doesn't
+use, which means your users may download more code than they strictly need.
+
+</div>
+
+## Download
```
justinfagnani, line 23: Before *Download* I would describe the two bundle type briefly in a list/table format.

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### Source maps
+
+Each bundle directory includes a source map. If you want to be able to view
+Lit's unminified sources while debugging, keep this file in the same directory
+as its associated bundle and don't rename it. If you update your project's
+bundle, you also need to update its source map.
+
+## CDN
```
justinfagnani, line 52: I would probably put CDN before Download - that might be the bigger use case

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+
+As an alternative to hosting your own copy of Lit's bundles, you can also import
+them from CDNs that serve assets directly from repositories hosted on GitHub.
+
+For example, you could import the latest `v2` release of [the `core`
+bundle](#core) from [jsDelivr](https://jsdelivr.com) like this:
+
+```js
+import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js';
```
justinfagnani, line 61: IMO, this should be one of the first bits of information in the doc

packages/lit-dev-content/site/docs/libraries/bundles.md
```diff
@@ -0,0 +1,103 @@
...
+[`html`](/docs/api/templates/#html), [`css`](/docs/api/styles/#css),
+[`render`](/docs/api/templates/#render), and
+[`LitElement`](/docs/api/LitElement/#LitElement), for example.
+
+### `all/`
+
+- `lit-all.min.js` - the `all` bundle
+- `lit-all.min.js.map` - the `all` bundle's source map
+
+The `all` bundle includes everything in the `core` bundle as well as [all
```
justinfagnani, line 96: I'd try to write this more concisely / declaratively:

Includes everything in "core" plus all other modules in the `lit` package. `html`, and `svg` from `lit/static-html.js` are renamed `staticHtml` and `staticSvg` respectively to avoid name collisions.

justinfagnani (changes requested):

arthurevans:
@justinfagnani looking at the Adding Lit page, it doesn't quite tie in with the rest of the Tools & workflows section. In particular, it doesn't mention bare module specifiers, which you need to deal with if you're installing from npm. 

I could maybe see moving Adding Lit and Starter Kits to the top of the section. Would probably need to add a forward reference to "Requirements" for the discussion on handling bare module specifiers. 

It would be a little bit buried here. If someone just wants to experiment with Lit using a CDN, they're not likely to look in a section on "Adding Lit to an existing project." But we could add a paragraph to Getting Started to point to this section, wherever it ends up.

Alternately, we could lead off the Tools section with "Installing Lit" and include npm, bundle download, & CDN instructions.

The "Adding Lit" page could get slimmed down a little--or we could use the new "Installing" page as a replacement for "Adding Lit."

packages/lit-dev-content/site/docs/libraries/bundles.md
bicknellr: I think the unversioned link to jsDeliver now covers this.

bicknellr:
I've condensed this and moved it into the "Getting Started" page, PTAL:
https://pr673-5701dd4---lit-dev-5ftespv5na-uc.a.run.app/docs/getting-started/#use-bundles

packages/lit-dev-content/site/docs/getting-started.md
```diff
@@ -45,6 +45,48 @@ import {LitElement, html} from 'lit';
 
 {% endswitchable-sample %}
 
+## Use bundles
+
+Lit is also available as pre-built, single-file bundles. The bundles are
+standard JavaScript modules with no dependencies - any modern browser should be
+able to import and run the bundles directly.
```
kevinpschaaf, line 52: Should we add something like this so it's _super_ clear?
```suggestion
able to import and run the bundles from within a `<script type="module">`.
```

I know we're intentionally not choosing to support legacy script use cases and I think this is good, but I'm also not sure how common distributing an _ESM_ bundle is for the library bundle use case.

sorvell (approved):

packages/lit-dev-content/site/docs/getting-started.md
```diff
@@ -45,6 +45,48 @@ import {LitElement, html} from 'lit';
...
+
+**If you're using npm for client-side dependencies, you should use [the `lit`
+package](#install-locally-from-npm), not these bundles.** The bundles
+intentionally combine most or all of Lit into a single file, which can cause
+your page to download more code than it needs.
+
+</div>
+
+To browse the bundles, go to <https://cdn.jsdelivr.net/gh/lit/dist/> and use the
+dropdown menu to choose a version. There are two types of bundle:
```
aomarks, line 64:
```suggestion
dropdown menu to choose a version. There are two types of bundles:
```

packages/lit-dev-content/site/docs/getting-started.md
bicknellr: sgtm

packages/lit-dev-content/site/docs/getting-started.md
```diff
@@ -45,6 +45,49 @@ import {LitElement, html} from 'lit';
 
 {% endswitchable-sample %}
 
+## Use bundles
+
+Lit is also available as pre-built, single-file bundles. The bundles are
```
thescientist13, line 50: Would it make sense to enumerate some of the uses of this vs the "classic" npm approach?  e.g. rapid prototyping, cloud IDEs, etc just to help make it clear(er) why this option exists, but why the other is probably more favored / idiomatic?

bicknellr:
Hmm, I dismissed @justinfagnani's review hoping to get an option to re-request, but I don't see it. PTAL though

packages/lit-dev-content/site/docs/getting-started.md
bicknellr: Thanks for taking a look! I've added some text that was suggested when I brought this up earlier today. We're trying to keep this section pretty short.

AndrewJakubowicz (approved):
Left a small comment. Tested the two bundles in the playground and I love it!

[Testing core bundle playground link](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy5qcyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L2doL2xpdC9kaXN0QDIvY29yZS9saXQtY29yZS5taW4uanMnO1xuXG5leHBvcnQgY2xhc3MgU2ltcGxlR3JlZXRpbmcgZXh0ZW5kcyBMaXRFbGVtZW50IHtcbiAgc3RhdGljIHN0eWxlcyA9IGNzc2BwIHsgY29sb3I6IHB1cnBsZSB9YDtcblxuICBzdGF0aWMgcHJvcGVydGllcyA9IHtcbiAgICBuYW1lOiB7dHlwZTogU3RyaW5nfSxcbiAgfTtcblxuICBjb25zdHJ1Y3RvcigpIHtcbiAgICBzdXBlcigpO1xuICAgIHRoaXMubmFtZSA9ICdTb21lYm9keSc7XG4gIH1cblxuICByZW5kZXIoKSB7XG4gICAgcmV0dXJuIGh0bWxgPHA-SGVsbG8gYnVuZGxlcyEgWUVTISBUaGFuayB5b3UgUnVzc2VsbCEhISE8L3A-YDtcbiAgfVxufVxuY3VzdG9tRWxlbWVudHMuZGVmaW5lKCdzaW1wbGUtZ3JlZXRpbmcnLCBTaW1wbGVHcmVldGluZyk7XG4ifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhRE9DVFlQRSBodG1sPlxuPGhlYWQ-XG4gIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2ltcGxlLWdyZWV0aW5nLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPHNpbXBsZS1ncmVldGluZyBuYW1lPVwiV29ybGRcIj48L3NpbXBsZS1ncmVldGluZz5cbjwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgfVxufSJ9XQ).

packages/lit-dev-content/site/docs/getting-started.md
```diff
@@ -45,6 +45,51 @@ import {LitElement, html} from 'lit';
...
+
+**If you're using npm for client-side dependencies, you should use [the `lit`
+package](#install-locally-from-npm), not these bundles.** The bundles
+intentionally combine most or all of Lit into a single file, which can cause
+your page to download more code than it needs.
+
+</div>
+
+To browse the bundles, go to <https://cdn.jsdelivr.net/gh/lit/dist/> and use the
+dropdown menu to choose a version. There are two types of bundles:
```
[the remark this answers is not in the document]
AndrewJakubowicz: Ooops this is my bad. I got confused because the list of versions is currently really short.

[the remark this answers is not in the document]
bicknellr: Thanks for pointing this out, I've added a bit of text here that should hopefully make this clearer.

PR #683, opened 2022-02-03
[tutorials] Add asides and fix anchors

e111077:
Markdown headers e.g. `## header 2` would generate a linkable anchor which would link to wrong things inside the tutorials section. We hide them here.

Additionally, we add support for asides in tutorial markdown (still getting finalized styles from ocupop but this is how they look now)

image


packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -145,6 +147,34 @@ customElements.define('my-element', MyElement);
 {% endswitchable-sample %}
 </pre>
 
+### Asides
+
+You can also insert an aside in your instructions by using the following format:
+
+```html
+<aside class="aside aside-positive">
```
aomarks, line 155: Nice use of `<aside>`. Would support converting all our existing `<p class="alert">` to this too.

aomarks, line 155: Doesn't seem like we need a class `aside` when the element itself is an `aside`? `<aside class="positive">` etc. seems simpler?

aomarks (approved):

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
e111077: I went on google codelabs and looked at what element they used haha

e111077: oh geez it should be alert. Will change though like this syntax more

aomarks (approved):

aomarks (approved):

PR #684, opened 2022-02-03
Increase tutorial-card image height

e111077:
increase image height to 5x unit so that the image ratio is 2:1 rather than 5:2


aomarks (approved):

AndrewJakubowicz (approved):
Al is fast!

PR #686, opened 2022-02-04
[tutorials] make "after" directory optional

e111077:
Tutorial manifest now has 

```ts
interface TutorialStep {
  title: string;
  hasAfter?: boolean;
}
```

if `hasAfter` is not defined, then the `solve` button will load the `before/` directory of the next step. If `hasAfter` is defined as `true` for that step, then the `solve` button will load the `after/` directory of that step.

Commits without tutorial refactoring:

https://github.com/lit/lit.dev/pull/686/files/8b816e8017d2fa80fe2301573ec1f80b5f05854f..HEAD


arthurevans:
NIce.

PR #688, opened 2022-02-04
Blog post for eleventy plugin

aomarks:
Blog post announcing the first release of `@lit-labs/eleventy-plugin-lit`


packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+  - al-marks
+---
+
+The Lit team is pleased to announce the first release of
+[`@lit-labs/eleventy-plugin-lit`](https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit),
+a new plugin for [Eleventy](https://www.11ty.dev/) that pre-renders your Lit
+components as static HTML during your Eleventy build, and lets you re-hydrate
+them after your JavaScript loads.
+
+<center>
```
AndrewJakubowicz, line 17: Nit: Is this deprecated? [[ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center)]

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+</center>
+
+<aside class="warning">
+
+As with all Lit Labs projects, this release should be considered experimental
+and incomplete. We'd love for you to try it out and leave feedback, but expect
+to find a few rough edges.
+
+Check out the <a
+href="https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit">roadmap</a>
```
AndrewJakubowicz, line 30: I clicked this and was a little confused about where the roadmap is.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
+---
+tags: blogPosts
+layout: blog-post.html
+title: "Eleventy + Lit"
+summary: "Announcing a new Lit Labs Eleventy plugin for static pre-rendering of Lit components"
```
justinfagnani, line 5: Let's double check that we want to call this "static pre-rendering". I can't necessarily think of anything better, but I wonder if this is meaningful to everyone?

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+date: 2022-02-07
+author:
+  - al-marks
+---
+
+The Lit team is pleased to announce the first release of
```
justinfagnani, line 11: maybe "first Lit Labs release" or "first experimental/preview release" to drive home the point in the lede before the alert.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+and styles, preserving all of the encapsulation boundaries of web components,
+even before a single line of JavaScript has downloaded!
+
+Later, when the JavaScript definition of `<demo-greeter>` has finally loaded,
+the pre-rendered component is upgraded to its fully interactive JavaScript
+implementation through a process called re-hydration.
+
+## Why should I use it?
+
+Pre-rendering components is an effective method for reducing the time it takes
```
justinfagnani, line 102: Maybe "can be an effective method"? We've definitely seen cases on large sites where deep SSR of components is _slower_ because of the additional HTML size.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+between a page starting to load, and the content becoming visible — a
+measurement known as the *time to first paint (FP)* or *time to first contentful
```
justinfagnani, line 104: Web Vitals is trying to reenforce the idea that LCP is a better metric. I think that's relevant here in that all the content is rendered early and you don't get a meaningless first paint w/o the component contents. Don't know if there's a way to weave that in, but you could just reference LCP and link to web.dev?

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+## How does it work?
+
+### Lit SSR
+
+The plugin uses `@lit-labs/ssr` under the hood, which works by creating an
+instance of each component in a lightweight browser-like environment within the
+Eleventy Node process, and evaluating the element's template using a special
+version of Lit's `render` function. This special `render` function behaves much
+like the `render` function that Lit uses in the browser, except that it emits
+stream of strings, instead of writing to the DOM.
```
AndrewJakubowicz, line 126: Nit: "**a** stream" or maybe "stream**s** of strings"? If it's one stream then I think it needs the added "a" before it.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+into your site is
+[documented](https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit#ponyfill)
+in the `@lit-labs/eleventy-plugin-lit` README.
+
+For more information about Declarative Shadow DOM, check out *[Declarative
+Shadow DOM (web.dev)](https://web.dev/declarative-shadow-dom/)*.
+
+### Re-hydration
+
+While static pre-rendering lets your components initial paint happen as quickly
```
kevinpschaaf, line 154:
```suggestion
While static pre-rendering lets your components' initial paint happen as quickly
```

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+upgraded to its JavaScript implementation, becoming responsive and interactive.
+
+Lit comes with automatic support for re-hydration via the special
+`lit/experimental-hydrate-support.js` module. As long as this module has been
+loaded, Lit components will detect when they have been pre-rendered, and adopt
+the shadow root that has already been attached to them.
+
+Re-hydration with Lit is very efficient, because Lit knows that it doesn't need
+to re-render the existing shadow root when it upgrades. If any data has changed
+between pre-rendering and upgrade, then only the parts of the shadow root that
```
justinfagnani, line 168: Actually, we currently require the same data on hydration in order to prime the previous value state of parts. This is something to discuss.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
AndrewJakubowicz: Oops it's in a pending PR: https://github.com/lit/lit/pull/2488

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+changed will be updated.
```
kevinpschaaf, line 169: This isn't entirely true today because we require the initial hydration to be done with the same data rendered on the server, after which the data can change and re-render (for now in the 11ty plugin this will always be true since the data can only come from static attributes).

This is a design choice, between hydration just priming the internal storage used for dirty-checking subsequent updates with initial data, vs. actually diffing what's in the DOM against the initial data and making changes at hydration time (we only do the former today). The open question of what we actually want to support is documented in https://github.com/lit/lit/issues/1434.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Yes good point. I think matching the precise terminology used in https://developers.google.com/web/updates/2019/02/rendering-on-the-web could be a good idea.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
kevinpschaaf: For the last sentence, I think you could just say:
```suggestion
to re-render the existing shadow root when it upgrades. If data changes after
hydration, then only the parts of the shadow root that changed will be updated.
```

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+as possible, the components won't yet be interactive, because their JavaScript
+implementations haven't loaded.
+
+*Re-hydration* is the process where a statically pre-rendered component is
+upgraded to its JavaScript implementation, becoming responsive and interactive.
+
+Lit comes with automatic support for re-hydration via the special
+`lit/experimental-hydrate-support.js` module. As long as this module has been
+loaded, Lit components will detect when they have been pre-rendered, and adopt
+the shadow root that has already been attached to them.
```
AndrewJakubowicz, line 164: Nit: "attached to them" somehow confused me. Is it an implementation detail? Does it still make sense if shortened to: "and adopt the pre-rendered shadow root."
No action required.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+
+Check out the
+[roadmap](https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit#roadmap)
+for `@lit-labs/eleventy-plugin-lit` for some of the main features and fixes you
+can expect to land over the coming months.
+
+If you try out `@lit-labs/eleventy-plugin-lit`, let us know if you have any
+suggestions or find any problems by starting a
+[discussion](https://github.com/lit/lit/discussions) or filing an
+[issue](https://github.com/lit/lit/issues).
```
AndrewJakubowicz, line 181: Doesn't need to be addressed, but we could also add some query params here so that we can quickly identify where issues have come from. I.e.

`https://github.com/lit/lit/issues/new?title=[11ty-ssr]:%20Describe%20your%20issue` Will pre-fill a title with [11ty-ssr].

AndrewJakubowicz (commented):
AMAZING!

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
augustjk: That's a great thought but I feel like linking to `/new` bypasses the template selection and skips the user potentially searching for an existing issue.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+DOM, but Firefox and Safari have not yet implemented it. A very small polyfill
+is available for browsers that don't yet have support. Integrating the polyfill
+into your site is
+[documented](https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit#ponyfill)
+in the `@lit-labs/eleventy-plugin-lit` README.
+
+For more information about Declarative Shadow DOM, check out *[Declarative
+Shadow DOM (web.dev)](https://web.dev/declarative-shadow-dom/)*.
+
+### Re-hydration
```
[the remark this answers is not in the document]
aomarks: I wondered about that too. https://developers.google.com/web/updates/2019/02/rendering-on-the-web always uses "re-hydration".

Maybe the logic is like, you started with a hydrated component, then you freeze-dried it when you rendered it to HTML, then you re-hydrate it to bring it back to life?

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Reading it more carefully, I think this is a summary of the way the terms are defined in this article:

- *Static rendering*: Build-time + mostly works without JS
- *Prerendering*: Build-time + mostly does not work without JS
- *Server Rendering*: Rendered on the server + mostly works without JS
- *Client-Side Rendering (CSR)*: All rendering and data fetches done on the client
- *Universal Rendering* or *SSR*: Server Rendering + Client-Side Rendering with rehydration

One key snippet:

```
However, it’s important to understand the difference between static rendering and
prerendering: static rendered pages are interactive without the need to execute
much client-side JS, whereas prerendering improves the First Paint or First
Contentful Paint of a Single Page Application that must be booted on the client
in order for pages to be truly interactive.
```

It seems like by these definitions, we could be doing either "static rendering" or "prerendering" depending on the nature of the components.

I'm also a little confused that the article seems to distinguish between "Server Rendering" and "Server-Side Rendering"!

justinfagnani: The distinction between "Server Rendering" and "SSR" is one I've heard and used before. Basically it's calling out that we've always done some amount of "server side rendering" like with CGI and PHP. "Server rendering" is supposed to be distinct from SSR which is now deep rendering of previously client-side components on the server.

Static rendering vs prerendering is _not_ a distinction I've seen elsewhere.

My thoughts about "static pre-rendering" are:
* Do people understand "static"? Probbaly since this is for a static site generator?
* Do people understand "pre-rendering" as opposed to just "rendering"? I'm not sure I even do. 🤷‍♂️ All of eleventy is "pre" rendering, no?

So maybe "static rendering"?

Also, separately, I almost always see "hydration" not "rehydration"

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,181 @@
...
+it takes to download, parse, and evaluate the JavaScript implementations of
+those components.
+
+For a deeper dive into the benefits of pre-rendering, check out *[Rendering on
+the Web
+(developers.google.com)](https://developers.google.com/web/updates/2019/02/rendering-on-the-web)*.
+
+## How does it work?
+
+### Lit SSR
```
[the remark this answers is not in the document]
aomarks: Done

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Oops yep thanks. Done!

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Yes I like your suggestion, done!

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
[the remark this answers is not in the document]
aomarks: Switched to "hydration", despite that article.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Yeah, I wanted to encourage searching for existing issues.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Done. Removed the metric discussion from this first paragraph all together (doesn't really need to be there), but added a link to the web.dev LCP article in the "see also" paragraph after it.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: > So maybe "static rendering"?

Done. Makes sense, I like the alignment with "static site generator" in this context. Removed all use of "pre-render" in favor of "static rendering" or "statically rendered".


> Also, separately, I almost always see "hydration" not "rehydration"

Done.

justinfagnani (approved):
Awesome!

kevinpschaaf (approved):
Looks great.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
AndrewJakubowicz: Yay! Nit: add `#roadmap` fragment.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
```diff
@@ -0,0 +1,176 @@
...
+### Declarative Shadow DOM
+
+When `@lit-labs/ssr` renders a component, it uses `<template shadowroot>` HTML
+elements to represent the component's shadow DOM and styles. When the browser
+HTML parser encounters a `<template shadowroot>`, it applies that template's
+content as the shadow root of its parent element. This way, all of the DOM and
+style encapsulation guarantees of web components are preserved.
+
+As of February 2022, Chrome and Edge have native support for Declarative Shadow
+DOM, but Firefox and Safari have not yet implemented it. A very small polyfill
```
AndrewJakubowicz, line 139: Doesn't need action, but "polyfill" is used here whilst the readme uses "ponyfill".

Maybe change all usages to be "ponyfill"? [Is this the definition for ponyfill?](https://github.com/sindresorhus/ponyfill).

🦄🦄🦄🦄🦄🦄🦄🦄🦄

AndrewJakubowicz (approved):
🦄🦄🦄🦄🦄Amazing!🦄🦄🦄🦄🦄

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
AndrewJakubowicz: Ah really great point about RSS.

packages/lit-dev-content/site/blog/2022-02-07-eleventy.md
aomarks: Yeah, the distinction is that a polyfill patches an API into existence, whereas a ponyfill serves exposes an equivalent API but through a separate explicit API. The declarative template shadow root is a ponyfill by this definition, because you have to call a `hydrateShadowRoots` function -- it's not enough to just import the script.

I think the distinction is not super useful here though, plus I think we might switch it to be a proper polyfill sometime soon.

Changed the terminology in the README at https://github.com/lit/lit/pull/2499, and the one spot here too.

PR #693, opened 2022-02-09
Fix `did` with `div`

AndrewJakubowicz:
A minor typo fix. I think html falls back on valid html in the case it doesn't understand the tag.

### DIFF
--- a/packages/lit-dev-content/site/playground/index.html
+++ b/packages/lit-dev-content/site/playground/index.html
@@ -69,7 +69,7 @@ title: Playground
      id="project">
   </playground-project>
 
-  <did id="tabsEditorAndPreview">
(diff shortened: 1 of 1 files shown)


augustjk (approved):
I see what you "did" there. 😉  Good find!

arthurevans:
In this case, it looks like it creates an `HTMLUnknownElement` instead of a div... But I guess it doesn't matter because it's styled using an ID selector...

Screen Shot 2022-02-08 at 6 22 39 PM

(One of the effects of using a hyphenated custom element name like `my-el` is that the browser knows to parse it as a (potential) custom element even if it hasn't been defined yet. So it ends up an `HTMLElement` full of future potential instead of a perpetual `HTMLUnknownElement`. )

PR #697, opened 2022-02-10
[tutorials] Update intro tutorial step 7

### LINKED ISSUE #691: [tutorials] Missing css import
On Step 7/8 of the tutorial the provided code is missing the `css` import. While this is pretty easy to figure out, it may cause some to stumble as there is no reference in the tutorial that it needs to be imported. Without it, the preview area just renders blank.

augustjk:
Fixes #691 

- Add js file to before to preserve css import
- Update after file to remove property only used in next step

### DIFF
--- a/packages/lit-dev-content/samples/tsconfig.json
+++ b/packages/lit-dev-content/samples/tsconfig.json
@@ -13,9 +13,6 @@
   },
   "include": ["./**/*.ts"],
   "exclude": [
-    // Intentionally contains unused imports. This one's .js output is
-    // hand-rolled becuase TypeScript will elide imports.
-    "tutorial/02-define/before/my-element.ts",
     // Position of static properties block is imperfect, and this example is
     // prominent on the homepage. See https://github.com/lit/lit/issues/2159.
     "examples/hello-world/simple-greeting.ts",
--- a/packages/lit-dev-content/samples/tutorials/intro-to-lit/06/before/todo-list.js
+++ b/packages/lit-dev-content/samples/tutorials/intro-to-lit/06/before/todo-list.js
@@ -1,34 +1,28 @@
 import {LitElement, html, css} from 'lit';
-import {customElement, property, query} from 'lit/decorators.js';
 
-type ToDoItem = {
-  text: string,
-  completed: boolean
-}
-
-@customElement('todo-list')
 export class ToDoList extends LitElement {
-  static styles = css`
-    .completed {
-      text-decoration-line: line-through;
-      color: #777;
-    }
-  `;
+  static properties = {
+    listItems: {attribute: false},
+  };
+  
+  // TODO: Add styles here
 
-  @property({attribute: false})
-  listItems = [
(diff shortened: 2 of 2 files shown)


packages/lit-dev-content/samples/tutorials/intro-to-lit/06/after/todo-list.ts
```diff
@@ -15,13 +15,11 @@ export class ToDoList extends LitElement {
     }
   `;
 
-  @property()
+  @property({attribute: false})
```
AndrewJakubowicz, line 18: Great catch. This made me realize there are a couple other inconsistencies between Step 6 and Step 7.

In Step 6 there's indentation that changes on the solve.
Step 6 uses `class` whilst Step 7 uses `export class`.

However they aren't blocking. Maybe we track in an issue and resolve separately.

packages/lit-dev-content/samples/tutorials/intro-to-lit/06/after/todo-list.ts
```diff
@@ -15,13 +15,11 @@ export class ToDoList extends LitElement {
...
   listItems = [
     { text: 'Make to-do list', completed: true },
     { text: 'Add some styles', completed: true }
   ];
-  @property()
```
AndrewJakubowicz, line 23: Great fix!

AndrewJakubowicz (approved):
Nice. I also manually did the tutorial steps modified and everything works great when following the tutorial text.

This is a strict improvement since it fixes the JS tutorial. I don't think my nits block this change.

e111077 (changes requested):
Please add an exemption here for that TS file

https://github.com/lit/lit.dev/blob/324ad70458663f8f7c4960a3f60cbe8d587d2a83/packages/lit-dev-content/samples/tsconfig.json#L18

Also that highlighted one, do you mind changing that to `tutorials/intro-to-lit/01/before/my-element.ts` while you're there?

e111077:
Hmm nevermind it seems like Al says it'll automatically know. Do you mind removing that line in the tsconfig then?

augustjk:
> Hmm nevermind it seems like Al says it'll automatically know. Do you mind removing that line in the tsconfig then?

Good find! It has been removed.

packages/lit-dev-content/samples/tutorials/intro-to-lit/06/after/todo-list.ts
arthurevans: Mea culpa here. It's really hard to maintain all the before/after versions (@e111077's recent changes should help a ton with that). Currently the `{attribute: false}` is also missing in the previous step where this property is added.

https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/samples/tutorials/intro-to-lit/05/before/todo-list.ts#L6

Looks like the everything _before_ step 7 omits the export statement.

e111077 (approved):

PR #704, opened 2022-02-15
Fix broken 404 pages with file extensions and 404 status

AndrewJakubowicz:
Currently a 404 url that doesn't end in a `/` causes a very strict CSP policy to be sent. This results in a very broken page being rendered on the browser. There was a second bug where a 404 page was always being set to a status of `200`.

Repro for CSP issue: https://lit.dev/test.ext
Repro of 200 status: https://lit.dev/repro/

Thank you to @aomarks for thinking of the general technique and proposing this solution.

Fix is to set a `404` status code after the the notfound middleware sends the 404 html page, and set a less strict CSP policy for 404 status codes.

### DIFF
--- a/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
+++ b/packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
@@ -234,8 +234,8 @@ export const contentSecurityPolicyMiddleware = (
     }
     ctx.set('Content-Security-Policy', policy);
     await next();
-    if (ctx['404-csp-intercept']) {
-      // If the 404 middleware has run, use a more forgiving CSP policy.
+    if (ctx.status === 404) {
+      // Ensure 404 page can render with less strict policy.
       ctx.set('Content-Security-Policy', entrypointsCsp);
     }
   };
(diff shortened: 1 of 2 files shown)


aomarks:
> Originally I tried to detect a `404` status code in the content security policy middleware, but it is always `200` because of the `notfound-middleware`. Instead could get this working by setting a private field, '404-csp-intercept', on the Koa context for the request.

Oh! Yeah it looks like the status is getting set to 200 in the case of a missing file. It should definitely remain a 404, not a 200. It seems weird that `send()` would change the status? https://github.com/lit/lit.dev/blob/main/packages/lit-dev-server/src/middleware/notfound-middleware.ts

image

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -233,6 +233,10 @@ export const contentSecurityPolicyMiddleware = (
       policy = strictFallbackCsp;
     }
     ctx.set('Content-Security-Policy', policy);
-    return next();
+    await next();
```
aomarks, line 236: Could move this above the `if`, and then it can be `if (ctx.path.endsWith('/') || ctx.status === 404) { ...`

AndrewJakubowicz:
Great catch. That is my bad. I noticed it but didn't question it.

This also makes this fix cleaner. Thank you!

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
AndrewJakubowicz: Beautiful!

aomarks (approved):
Great!

PR #709, opened 2022-02-18
integrate SSR and SSR tutorial catalog

### LINKED ISSUE #706: [infra] integrate 11ty SSR plugin
blocked by #705

e111077:
integrate SSR and SSR the tutorial catalog page (sans hydration)

Would specifically like feedback on whether adding `/src/components/ssr.js` is a good idea

fixes #706


e111077:
@aomarks can I be added to the GCP project? I don't have permissions to rerun the deploy which seems to have had a flake fail

aomarks:
> @aomarks can I be added to the GCP project? I don't have permissions to rerun the deploy which seems to have had a flake fail

Done. There may be an email you need to accept, not sure.

packages/lit-dev-content/src/global/dsd-polyfill.ts
```diff
@@ -0,0 +1,11 @@
+(async () => {
+  if (!HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')) {
+    const {hydrateShadowRoots} = await import(
```
aomarks, line 3: Since this is an async operation, it's not going to block the loading of component definitions (even though the _module_ is guaranteed to execute before component scripts do -- that's not sufficient here).

I believe that means there's a race condition where a component could upgrade before hydration support was installed, so it wouldn't hydrate.

This is why the example strategy at https://github.com/lit/lit/tree/main/packages/labs/eleventy-plugin-lit#example-bootup-strategy moves component modules into dynamic `import` scripts after this promise has resolved.

packages/lit-dev-content/site/_includes/default.html
```diff
@@ -15,6 +15,12 @@
     {% inlinejs "global/apply-mods.js" %}
     {% inlinejs "global/initialize-typescript-attribute.js" %}
     {% inlinejs "global/mobile-drawer.js" %}
+    {% if ssr %}
+      {% if not nohydrate %}
+        <script type="module" src="{{ site.baseurl }}/js/global/lit-hydrate-support.js"></script>
+      {% endif %}
+      <script type="module" src="{{ site.baseurl }}/js/global/dsd-polyfill.js"></script>
```
aomarks, line 22: I think this should be inlined, not external.

aomarks (changes requested):
Nice!

aomarks:
Thoughts on how we would evaluate whether this is beneficial or not? Deciding when something *should* be SSR'd is an interesting open question. I guess we should do some before/after lighthouse scores/first paint numbers/bytes over wire comparisons.

Just looking at the HTML, something that jumps out to me is how every card instance has a fairly large `<style>` block duplicated into it. It probably compresses well with gzip/brotli, at least.

packages/lit-dev-content/src/global/dsd-polyfill.ts
aomarks: (Minor correction, I said "hydration support" above, but I should have said "the declarative shadow dom polyfill". A similar timing issue exists though -- the polyfill must have installed before any components upgrade.)

I think we'd need to define a `components.js` or similar module for every page, instead of individual component script tags, so that we can defer component imports until the polyfill has run:

```ts
import {installDsdPolyfillIfNeeded} from '../dsd-polyfill.js';
(async() {
  await installDsdPolyfillIfNeeded();
  import('../components/foo.js');
  import('../components/bar.js');
})();
```

*Or*, another approach would be to always import the polyfill (maybe even inline it), and do the check + install synchronously. This is actually probably the simpler way to go, since it just requires that you put the polyfill `<script>` tag before any component definitions. The downside is that we always download the polyfill even on Chrome -- but it's so small that this probably doesn't matter (and maybe we can make it even smaller). I think I'm leaning towards this approach...

aomarks (changes requested):

packages/lit-dev-content/src/global/dsd-polyfill.ts
e111077: TBH I feel like the underlying issue here is that the _right way_<sup>(tm)</sup> is bad DX as you have to define a new entrypoint in the rollup config after you create your new `.ts` file. How about we make a subdir in our src directory called something like `entrypoints` or `/src/pages/entrypoints` that rollup just does an `fs` call and automatically create a new bundle for files in that directory.

That should enable us to do the correct method more easily.

packages/lit-dev-content/rollup.config.js
```diff
@@ -127,4 +129,33 @@ export default [
       }),
     ],
   },
+
+  /**
+   * Creates an iife for scripts that need to be inlined with inlinesyncjs and
```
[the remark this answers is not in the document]
e111077: it's the only output mode I can find on rollup that will bundle everything into a single file. In case somebody down the line wants to use the `inlinesyncjs` macro they don't have to worry about imports

packages/lit-dev-content/src/global/dsd-polyfill.ts
aomarks: I think the "always load the polyfill" approach is the simplest, especially for now, and maybe forever. The declarative shadow dom polyfill is really small. So if we just put its script tag first, nothing much else needs to change with our bundling strategy.

packages/lit-dev-content/rollup.config.js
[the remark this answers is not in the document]
e111077: actually it's not necessary anymore removing

e111077:
Before SSR (score 94):
image

After (score 97):
image

We can probably also run Elizabeth's `rollup-plugin-minify-html-literals` which also minifies css template literals.

e111077:
For reference:

Live page:
https://lit.dev/tutorials/?mods=tutorialCatalog

SSR without template minification:
https://pr709-e49a402---lit-dev-5ftespv5na-uc.a.run.app/tutorials/?mods=tutorialCatalog

SSR with template minification:
https://pr709-5c2feb0---lit-dev-5ftespv5na-uc.a.run.app/tutorials/?mods=tutorialCatalog

e111077:
After a bunch of reruns the SSR'd approach seems like a better experience on slow 3G because it's instantly `<a>` tag clickable. The numbers don't support LCP though. Not sure how we can numbers this correctly

e111077:
Quick ping on this one @aomarks

packages/lit-dev-content/src/global/dsd-polyfill.ts
```diff
@@ -0,0 +1,5 @@
+import {hydrateShadowRoots} from '@webcomponents/template-shadowroot/template-shadowroot.js';
+
+// Polyfill will short if the browser supports native DSD
```
aomarks, line 3: True, though it does it in a more costly way than necessary I think (https://github.com/webcomponents/template-shadowroot/blob/master/src/_implementation/feature_detect.ts could just be `HTMLTemplateElement.prototype.hasOwnProperty('shadowRoot')`). We should probably just update the polyfill, though, so no action needed.

aomarks (approved):
Cool! The new approach inlining the polyfill looks good.

Have you tested this with a hydrated component?

aomarks:
> The numbers don't support LCP thoug

Can you clarify? Do you mean LCP is worse with SSR on average?

e111077:
I've SSRd and hydrated some other stuff like the litdev-buttons. It has worked just fine!

ignore the LCP comment can't reproduce anymore, but basically LCP rating was worse on the SSR'd version but the actual experience was better on slow 3g  + 6x cpu slowdown

PR #710, opened 2022-02-18
Refactor playground page into a component

aomarks:
Previously, most of the logic for orchestrating the Playground page was in a regular script that waited for DOMContentLoaded, and then manually queried for the underlying Playground components and hooked them together.

Now... we still do most of that! Except, it's contained within a new `<litdev-playground-page>` component.

This is the first step in a refactoring to make the Playground easier to work on.

You **definitely** want to look at the commits separately, because the first one just does a rename. Almost all of the logic is the same here, and the second commit should show that much more easily than viewing the full diff at once.


AndrewJakubowicz (approved):
Exceptional 👏 

Locally I also tested:
 - ctrl + s with gists and without gists. ✅ 
 - Navigating to an example (and having examples scroll correctly)  ✅ 
 - Navigated from shared long url.  ✅ 
    - Correctly didn't display language toggle.  ✅ 
 - Download button.  ✅ 
 - Clicked on a handful of examples.  ✅ 

Lightly tested across Firefox and Safari.

AndrewJakubowicz (changes requested):
oops found something tiny in Firefox.

The file-list seems to have slightly changed positioning on Firefox resulting in a scroll bar:
Screen Shot 2022-02-17 at 5 54 19 PM

Edit: It reproduces in production. Nvm. Punting. Is it my default Firefox settings.

aomarks:
> oops found something tiny in Firefox.
> 
> The file-list seems to have slightly changed positioning on Firefox resulting in a scroll bar: Screen Shot 2022-02-17 at 5 54 19 PM
> 
> Edit: It reproduces in production. Nvm. Punting. Is it my default Firefox settings.

Oh dear. Want to file a quick issue with this screenshot?

PR #711, opened 2022-02-18
Fix blog author list

e111077:
In eleventy 1.0.0 frontmatter data of a layout is prepended rather than overridden


aomarks (approved):

AndrewJakubowicz (approved):
LGTM! Also manually glanced over blog list and individual posts.

PR #712, opened 2022-02-18
fix broken playground test

e111077:


aomarks (approved):
Oops! My bad, thanks for fixing this!

PR #716, opened 2022-02-23
Update playwright tests to work with esmodules and add retry

AndrewJakubowicz:
**Note for reviewer**: the majority of line changes are package-lock.json from upgrading playwright in the lit-dev-tests package.

### Context

After updating playwright it no longer implicitly compiles typescript when type module is set in package.json. This results in an error `Error: playwright.config.ts: Cannot import a typescript file from an esmodule.`.

See repro of error here: https://github.com/lit/lit.dev/runs/5308236729?check_suite_focus=true

### Fix

- Compile the playwright config file and all the tests and run the tests from the built JavaScript directory instead.
- Update playwright config such that generated snapshots are still checked into the src directory. Required some name updates.
- While modifying the config, I added three retries to snapshot tests to minimize flakes.

This results in letting us upgrade playwright. Adding retry makes it much less likely for flakes to block CI.

### Testing

Tested locally with `npm` scripts:

- `npm run test:integration`
- `npm run test:integration:update-golden-screenshots`

Also checked that CI integration tests are green and also ran.


augustjk (commented):
I did find this https://playwright.dev/docs/release-notes#improved-typescript-support but using it required changing any type/interface imports to `import type` and emits Node warnings about usage of `--experimental-loader` flag.

Running test against compiled JS does work but I wonder if keeping it TS is slightly better for quicker feedback when writing tests?

augustjk:
Upon investigation with @AndrewJakubowicz we found that playwright's translation of TS is different than what `tsc` is doing so it probably makes more sense to stick with doing the compilation ourselves as this PR currently has. Seems more stable to do that.

augustjk (approved):

AndrewJakubowicz:
Async @augustjk suggested leveraging type only imports where possible, and also shared a TypeScript 4.5 feature: [`type` Modifiers on Import Names](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names), which will be great to use in the future once we're on TS 4.5. Will update with feedback.

aomarks:
> Upon investigation with @AndrewJakubowicz we found that playwright's translation of TS is different than what `tsc` is doing so it probably makes more sense to stick with doing the compilation ourselves as this PR currently has. Seems more stable to do that.

That sounds good to me too.

aomarks (approved):

PR #727, opened 2022-03-23
[tutorials] enable code checking

### LINKED ISSUE #698: [tutorials] Add a "check code" feature
Sometimes the user doesn't want to see the answer but rather only check whether their current answer is correct. A "check" code button on a tutorial step could work here.

Possible implementation:

* Author creates a hidden zzz_code-check.ts (something to make it show up last in gists) file in the project
* Author loads script at the end of index.html & subscribes to postmessage on window
* Author marks that step as checkable in tutorial.json
* FE reads tutorial.json to see that step is checkable and show the check button
* FE listens for postmessage from playground waiting for success
* When user clicks "Check" button, FE posts a message to run the unit test
  * unit test runs
  * unit test posts success or failure to window
* FE displays whether it failed or passed

Authoring experience seems a bit iffy here. Might have to make postmessage subscribe code copy-pastable, but i'd rather be able to just import helper functions for the author, but copying yet-another-file might not be great for the author

e111077:
Enables code checking on tutorials.

See tutorials/contributing and the "advanced templating" tutorial for a demo

fixes #698

https://pr727-94a077a---lit-dev-5ftespv5na-uc.a.run.app/tutorials/advanced-templating/?mods=tutorialCatalog


packages/lit-dev-content/samples/base.json
```diff
@@ -3,6 +3,10 @@
     "package.json": {
       "content": "{\n  \"dependencies\": {\n    \"lit\": \"^2.0.0\",\n    \"@lit/reactive-element\": \"^1.0.0\",\n    \"lit-element\": \"^3.0.0\",\n    \"lit-html\": \"^2.0.0\"\n  }\n}",
       "hidden": true
+    },
+    "_check-code-helpers.js": {
```
AndrewJakubowicz, line 7: Should this only be included in tutorials with check-code enabled? I think currently this will appear in your shared gist.

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -30,13 +30,32 @@ Lit tutorials are a way to provide a guided, interactive learning experience to
...
 
 * `before/` holds the playground project for what is first presented to the user
 * `after/` holds the playground project for when the user clicks the `solve` button.
   * If `hasAfter` is `false` or `undefined` for this step's metadata in `tutorial.json`, then the `after/` directory is optional and the step will load the next step's `before/` directory
 </details>
 
+<details>
+  <summary>If <code>checkable: true</code> in tutorial.json add code checking</summary>
+
+* create a `_check-code.js` file in the `before/` directorie
```
AndrewJakubowicz, line 43: Nit: directory

e111077:
Author now has to extend `/samples/check-code-base.json` instead of `/samples/base.json` for steps with code checking.

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -30,13 +30,35 @@ Lit tutorials are a way to provide a guided, interactive learning experience to
...
 </details>
 
+<details>
+  <summary>If <code>checkable: true</code> in tutorial.json add code checking</summary>
+
+* create a `_check-code.js` file in the `before/` directory
+* In your `project.json`
+  * set your `"extends"` field to `"/samples/checkable-tutorial-base.json"`
+    * e.g. `"extends": "/samples/checkable-tutorial-base.json"`
+  * add the code-checking file it as a `hidden` file to the step's `project.json`.
```
AndrewJakubowicz, line 47:
```suggestion
  * add the code-checking file as a `hidden` file to the step's `project.json`.
```

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -183,6 +208,87 @@ The available asides are:
...
+
+Next, import this new file into your `index.html` file. And feel free to add the `<!-- playground-fold(-end) -->` comments to hide the import.
+
+
+<details open>
+  <summary>example: <code>index.html</code></summary>
+
+```html
+<head>
+  <!-- playground-fold --><script type="module" src="./_check-code.js"></script><!-- playground-fold-end -->
```
AndrewJakubowicz, line 246: Could also recommend using `playground-hide` if we want this to never appear and confuse the user. I think with fold it's possible to expand this import.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -59,6 +69,13 @@ export class LitDevTutorial extends LitElement {
   @property()
   project?: string;
 
+  /**
+   * ID of the <playground-preivew> in the host scope whose iframe we must
```
AndrewJakubowicz, line 73:
```suggestion
   * ID of the <playground-preview> in the host scope whose iframe we must
```

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -59,6 +69,13 @@ export class LitDevTutorial extends LitElement {
...
+   * access in order to communicate across to and from the playground.
```
AndrewJakubowicz, line 74:
```suggestion
   * access in order to communicate to and from the playground.
```

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -334,17 +440,49 @@ export class LitDevTutorial extends LitElement {
...
-    </div>`;
+  private _renderCheckIcon() {
+    if (this._checkIndeterminate) {
+      return checkCodeIcon;
+    } else if (this._isChecking) {
+      return loopIcon;
+    } else if (this._checkPassed) {
+      return checkPassedIcon;
+    } else {
+      return checkFailedIcon;
```
AndrewJakubowicz, line 484: Complexity reducing idea.

With three independent state variables, it creates 8 combinations of booleans to handle. I think the state can be simplified into a single `@state` with multiple enum options or intersection type strings, i.e. `'check-indeterminate' | 'is-checking' ...` modeling a state machine pattern with transitions between states. This will reduce the complexity of needing to change multiple variables for one state change. For example right now `isChecking` and `checkPassed` can both be true but I think it's redundant.

It seems like the states are:

 - `checkIndeterminate` for showing the `checkCodeIcon` and button being enabled.
 - `isChecking` to prevent multiple concurrent checks.
 - `check-success` for success.
 - `check-failure` for failure.

^ Each of the above states are exclusive so can be stored on one `@state`.
What do you think?

packages/lit-dev-content/site/css/tutorial.css
```diff
@@ -286,6 +286,20 @@ resize-bar[active] {
   display: none;
 }
 
+.checkButton.checking svg {
+  animation: spin 5s linear infinite;
+}
+
+@keyframes spin {
```
AndrewJakubowicz, line 293: The loading state is so fast that this is never shown. May almost be worth adding a slow-down to the checking state to handle this case.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -400,22 +579,53 @@ export class LitDevTutorial extends LitElement {
...
+    this._checkIndeterminate = false;
+    this._checkPassed = true;
+    this._isChecking = false;
+    this._clearCheckingTimeout();
+    this._requestSolvedCode = true;
     this._setProjectSrc(this._info.projectSrcAfter, true);
   }
 
+  private _clearCheckingTimeout() {
+    if (this._currentCheckTimeout) {
```
AndrewJakubowicz, line 591: If you want, can remove the check. Passing null [is silently handled by clearTimeout](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout#notes).

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -30,13 +30,35 @@ Lit tutorials are a way to provide a guided, interactive learning experience to
...
+* import the new file in `index.html`
+  * e.g.
+```html
+<head>
+  <!-- playground-fold --><script type="module" src="./_check-code.js"></script><!-- playground-fold-end -->
+
+  <script type="module" src="./my-element.js"></script>
+</head>
+```
+* Install the code checker from `'./check-code-heper.js'` and pass it an async callback that returns an object of type `{passed: boolean, message?: string}`
```
arthurevans, line 58:
```suggestion
* Install the code checker from `'./_check-code-helpers.js'` and pass it an async callback that returns an object of type `{passed: boolean, message?: string}`
```

e111077: OMG SOMEONE HEP ME

packages/lit-dev-content/site/css/tutorial.css
e111077: Like an artificial slowdown?

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -30,13 +30,35 @@ Lit tutorials are a way to provide a guided, interactive learning experience to
...
 
+<details>
+  <summary>If <code>checkable: true</code> in tutorial.json add code checking</summary>
+
+* create a `_check-code.js` file in the `before/` directory
+* In your `project.json`
+  * set your `"extends"` field to `"/samples/checkable-tutorial-base.json"`
+    * e.g. `"extends": "/samples/checkable-tutorial-base.json"`
+  * add the code-checking file it as a `hidden` file to the step's `project.json`.
+    * e.g. `"_check-code.js": { "hiden": true }`
```
arthurevans, line 48:
```suggestion
    * e.g. `"_check-code.js": { "hidden": true }`
```

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
e111077: TIL 💯

packages/lit-dev-content/site/css/tutorial.css
e111077: oh I misread. Yeah, i put it in and tested that speed just for this reason

AndrewJakubowicz: Yeah. One second is perceived as instant, and would give some time to see the spinner. Right now it flashes by - at least in the current example.

arthurevans:
This is super cool! My one piece of feedback was that the code check message was kind of unobtrusive, especially since its background blends into the left-hand background. I wonder if we could make it a little easier to see, maybe add a border or shadow that contrasts with both sides. (Currently I think it has a faint gray/black shadow, which helps on the white side but not on the dark side.)

I know this is how a snackbar is supposed to look, but it's kind of problematic having it in the center of a sharply divided screen. It also seems a little off because it's not a passing notice from the app so much as a direct response to pressing the Check button. So it feels like UI wise it should be anchored more directly to the button.

As always, I Am Not a UI Designer.

e111077:
@arthurevans unfortunately, bg color is not configurable in mwc-snackbar (omg pretty sure this was me). And seeing how it's gonna go the way of the dino soon, I don't see much good reason to cludge in more customizations into it.

packages/lit-dev-content/site/css/tutorial.css
e111077: I added a delay to the second step of this tutorial so you can demo it

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -20,16 +20,26 @@ import {resetIcon} from '../icons/reset-icon.js';
...
+import {loopIcon} from '../icons/loop-icon.js';
 
 import '@material/mwc-icon-button';
+import '@material/mwc-snackbar';
+import {Snackbar} from '@material/mwc-snackbar';
 import './litdev-example-controls.js';
 import './litdev-playground-change-guard.js';
 import './litdev-icon-button.js';
 import {Task, TaskStatus} from '@lit-labs/task';
+import {PostDoc} from 'postdoc-lib';
```
aomarks, line 35: It seems a little unfortunate that a bunch of the interesting code that's coming in with this PR is not getting reviewed here, because it was published to a separate postdoc library, and that library wasn't reviewed.

Maybe some background on the problem the library solves would be helpful here.

For the sake of argument, what wouldn't work about this simple approach?:

- Parent sends a "check" message to the iframe.
- iframe receives message, checks `event.source === parentWindow` to confirm the message came from the parent, does the check, and then posts a response back to `parentWindow`
- Parent receives message and checks `event.source === iframe.contentWindow` and does its thing with the response.

e111077: The issues are around startup time for the iframes and missing the initial handshake messages which complicate the initial message sending. You have to write the same handshake logic on both sides as both need to initiate a handshake and receive a handshake depending on startup.

Additionally, this has to be repeated on reload, but that might be slightly easier with the src changing every time in this case.

Additionally postdoc uses a messagechannel to not pollute the window after handshake is made. This will become more relevant when we implement the "multiple tutorial steps on the same page / scrollable view".

Also with all the repeated code, it makes sense to centralize this into a tiny library, but since the code needs to be both in `/src` and in `/samples/checkable-tutorial.json`, that means the code would have to be duplicated due to our build system, and until we have wireit, there's a lot of cludge around messing with the build system. Hence the simplest solution would be to use something off of NPM.

There are other npm libraries i import and massage into what we need if you prefer. I also don't mind bringing postdoc into the lit org.

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -183,6 +208,86 @@ The available asides are:
 * `info`
 * `labs`
 
+## Code Checking
+
+To enable code checking for a step, add the `checkable: true` flag to the step in `tutorial.json`.
+
+Next create a file which you will run your code checking. In this example we will call it `_check-code.js`.
```
aomarks, line 215:
```suggestion
Next create a file which will run your code checking. In this example we will call it `_check-code.js`.
```

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -106,6 +128,9 @@ interface TutorialManifest {
     // Set to true if there is an "after" directory for this step or if it is
     // the last step in the tutorial.
     hasAfter?: boolean;
+    // Whether or not the step is code checkable. see the `Code Checking`
+    // section below for more details.
+    checkabe?: boolean;
```
aomarks, line 133:
```suggestion
    checkable?: boolean;
```

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -183,6 +208,86 @@ The available asides are:
 * `info`
 * `labs`
 
+## Code Checking
+
+To enable code checking for a step, add the `checkable: true` flag to the step in `tutorial.json`.
+
+Next create a file which you will run your code checking. In this example we will call it `_check-code.js`.
+
+> ⚠️ **N.B.** you **MUST** use a `.js` file extension or else this repo will not be able to TS build because we use playground-elements to inject the hidden `_check-code-helpers.js` file.
```
aomarks, line 217: I guess I am a pleb, but I didn't know what N.B. was. Suggest just leading with "Note:"

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -183,6 +208,86 @@ The available asides are:
...
+
+Then add this code check file as a `hidden` file to the `project.json` of the before directory.
```
aomarks, line 219:
```suggestion
Then add this code check file as a `hidden` file to the `project.json` of the `before` directory.
```

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -183,6 +208,86 @@ The available asides are:
...
+  <!-- playground-hide --><script type="module" src="./_check-code.js"></script><!-- playground-hide-end -->
+  <script type="module" src="./my-element.js"></script>
+</head>
+<body>
+  <my-element name="User"></my-element>
+</body>
+```
+</details>
+
+In your `_check-code.js` file, import and call the `installCodeChecker` from the hidden `'./_check-code-helpers.js'` file which is injected by playground elements in `checkable-tutorial-base.json`.
```
aomarks, line 255:
```suggestion
In your `_check-code.js` file, import and call `installCodeChecker` from the hidden `'./_check-code-helpers.js'` file which is injected by playground elements in `checkable-tutorial-base.json`.
```

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -183,6 +208,86 @@ The available asides are:
...
+
+`installCodeChecker` will set up communications between the tutorial page and the playground and call an async callback when the user requests code checking.
+
+The return type of the callback should be:
+
+`{passed: boolean, message?: string}`
+
+Where `passed` is whether the code has passed the checks and `message` is the optional error message to display`.
```
aomarks, line 263:
```suggestion
Where `passed` is whether the code has passed the checks and `message` is the optional error message to display.
```

packages/lit-dev-content/samples/tutorials/advanced-templating/01/before/_check-code.js
```diff
@@ -0,0 +1,13 @@
+import {installCodeChecker} from './_check-code-helpers.js';
+
+installCodeChecker(async () => {
+  const passed = true;
+  const message = 'This test always passes just showing off the delayed check state.';
```
aomarks, line 5:
```suggestion
  const message = 'This test always passes, just showing off the delayed check state.';
```

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -20,16 +20,26 @@ import {resetIcon} from '../icons/reset-icon.js';
...
 import {forwardArrowIcon} from '../icons/forward-arrow-icon.js';
 import {backArrowIcon} from '../icons/back-arrow-icon.js';
+import {checkFailedIcon} from '../icons/check-failed-icon.js';
+import {checkPassedIcon} from '../icons/check-passed-icon.js';
+import {checkCodeIcon} from '../icons/check-code-icon.js';
+import {loopIcon} from '../icons/loop-icon.js';
 
 import '@material/mwc-icon-button';
+import '@material/mwc-snackbar';
+import {Snackbar} from '@material/mwc-snackbar';
```
aomarks, line 30:
```suggestion
import type {Snackbar} from '@material/mwc-snackbar';
```

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -20,16 +20,26 @@ import {resetIcon} from '../icons/reset-icon.js';
...
 import './litdev-example-controls.js';
 import './litdev-playground-change-guard.js';
 import './litdev-icon-button.js';
 import {Task, TaskStatus} from '@lit-labs/task';
+import {PostDoc} from 'postdoc-lib';
+import {PlaygroundPreview} from 'playground-elements/playground-preview';
```
aomarks, line 36:
```suggestion
import type {PlaygroundPreview} from 'playground-elements/playground-preview';
```

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -100,9 +137,55 @@ export class LitDevTutorial extends LitElement {
   }
 
   /**
-   * Whether the tutorial is currently in its "solved" state.
+   * Whether the user has requested the solved code by clicking the solve button
```
aomarks, line 140: Maybe clarify this jsdoc a little? Wondering if `_requestSolvedCode` is going to get reset after the snackbar displays.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -100,9 +137,55 @@ export class LitDevTutorial extends LitElement {
...
+  @state() private _requestSolvedCode = false;
+
+  /**
+   * Receives check status messages from the playground and updates the check
+   * UI accordingly.
+   *
+   * @param e Message event from the playground. Has status and possible
+   *     validation message
+   */
+  private onPlaygroundMessage = (
```
aomarks, line 152:
```suggestion
  private readonly onPlaygroundMessage = (
```

packages/lit-dev-content/samples/checkable-tutorial-base.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "/samples/base.json",
+  "files": {
+    "_check-code-helpers.js": {
```
aomarks, line 4: Suggest making this an actual `.ts` file instead of inlining. All files in a `project.json` are automatically converted to be inline by https://github.com/lit/lit.dev/blob/main/packages/lit-dev-tools-cjs/src/playground-inline.ts.

packages/lit-dev-content/src/icons/check-code-icon.ts
```diff
@@ -0,0 +1,27 @@
...
+>
+  <g>
+    <rect fill="none" height="24" width="24" />
+  </g>
+  <g>
+    <path
+      d="M14.05,17.58l-0.01,0.01l-2.4-2.4l1.06-1.06l1.35,1.35L16.54,13l1.06,1.06 l-3.54,3.54L14.05,17.58z M12,2C6.5,2,2,6.5,2,12s4.5,10,10,10s10-4.5,10-10S17.5,2,12,2z M7.34,6.28l1.41,1.41l1.41-1.41 l1.06,1.06L9.81,8.75l1.41,1.41l-1.06,1.06L8.75,9.81l-1.41,1.41l-1.06-1.06l1.41-1.41L6.28,7.34L7.34,6.28z M12,20 c-2.2,0-4.2-0.9-5.7-2.3L17.7,6.3C19.1,7.8,20,9.8,20,12C20,16.4,16.4,20,12,20z"
+      fill-rule="evenodd"
+    />
+  </g>
```
aomarks, line 26: - xmlns is unnecessary on `<svg>` elements. 
- `enable-background` doesn't seem to do anything here (I removed it on the preview for the check-code icon, and there was no change).
- There is an empty `<rect>`.
- There are unneccessary `<g>`s.

```suggestion
// Source: google fonts material icons
export const checkCodeIcon = html`<svg
  height="24px"
  viewBox="0 0 24 24"
  width="24px"
  fill="currentcolor"
>
    <path
      d="M14.05,17.58l-0.01,0.01l-2.4-2.4l1.06-1.06l1.35,1.35L16.54,13l1.06,1.06 l-3.54,3.54L14.05,17.58z M12,2C6.5,2,2,6.5,2,12s4.5,10,10,10s10-4.5,10-10S17.5,2,12,2z M7.34,6.28l1.41,1.41l1.41-1.41 l1.06,1.06L9.81,8.75l1.41,1.41l-1.06,1.06L8.75,9.81l-1.41,1.41l-1.06-1.06l1.41-1.41L6.28,7.34L7.34,6.28z M12,20 c-2.2,0-4.2-0.9-5.7-2.3L17.7,6.3C19.1,7.8,20,9.8,20,12C20,16.4,16.4,20,12,20z"
      fill-rule="evenodd"
    />
```

packages/lit-dev-content/src/icons/check-code-icon.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: google fonts material icons
```
aomarks, line 9: Let's link to the specific icon URL, like https://material.io/resources/icons/?icon=restart_alt.

packages/lit-dev-content/src/icons/check-failed-icon.ts
```diff
@@ -0,0 +1,22 @@
...
+// Source: google fonts material icons
+export const checkFailedIcon = html`<svg
+  xmlns="http://www.w3.org/2000/svg"
+  enable-background="new 0 0 24 24"
+  height="24px"
+  viewBox="0 0 24 24"
+  width="24px"
+  fill="currentcolor"
+>
+  <rect fill="none" height="24" width="24" />
```
aomarks, line 18:
```suggestion
export const checkFailedIcon = html`<svg
  height="24px"
  viewBox="0 0 24 24"
  width="24px"
  fill="currentcolor"
>
```

packages/lit-dev-content/src/icons/check-passed-icon.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Source: google fonts material icons
+export const checkPassedIcon = html`<svg
+  xmlns="http://www.w3.org/2000/svg"
+  height="24px"
+  viewBox="0 0 24 24"
+  width="24px"
+  fill="currentcolor"
+>
+  <path d="M0 0h24v24H0V0z" fill="none" />
```
aomarks, line 17:
```suggestion
export const checkPassedIcon = html`<svg
  height="24px"
  viewBox="0 0 24 24"
  width="24px"
  fill="currentcolor"
>
```

packages/lit-dev-content/src/icons/loop-icon.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Source: google fonts material icons
+export const loopIcon = html`<svg
+  xmlns="http://www.w3.org/2000/svg"
+  height="24px"
+  viewBox="0 0 24 24"
+  width="24px"
+  fill="currentcolor"
+>
+  <path d="M0 0h24v24H0V0z" fill="none" />
```
aomarks, line 17:
```suggestion
  height="24px"
  viewBox="0 0 24 24"
  width="24px"
  fill="currentcolor"
>
```

packages/lit-dev-content/src/icons/loop-icon.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: google fonts material icons
```
aomarks, line 9: Link to specific icon.

packages/lit-dev-content/src/icons/check-passed-icon.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: google fonts material icons
```
aomarks, line 9: Link to specific icon.

packages/lit-dev-content/src/icons/check-failed-icon.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: google fonts material icons
```
aomarks, line 9: Link to specific icon.

packages/lit-dev-content/src/icons/check-code-icon.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
aomarks, line 1: So far we've been naming these files with the name of the icon, rather than where they are used. An advantage of that is that when you are looking to see if we have some material icon, you can look in the `icons/` directory to find it. The same icon might be useful in two contexts on the site.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -334,17 +429,53 @@ export class LitDevTutorial extends LitElement {
...
+    return html`<div id="tutorialFooter">
+        <litdev-icon-button @click=${this._onClickSolve}>
+          ${solveIcon} Solve
+        </litdev-icon-button>
+
+        ${this._manifestTask.render({
+          complete: (manifest) =>
+            !manifest.steps[this._idx].checkable
+              ? nothing
+              : html` <litdev-icon-button
```
aomarks, line 447: Maybe factor this out into a `_renderCheckButton`? Indentation is a little severe.

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -373,6 +504,19 @@ export class LitDevTutorial extends LitElement {
       CODE_LANGUAGE_CHANGE,
       this._onCodeLanguagePreferenceChanged
     );
+
+    const previewElement = this._preview!;
+    // must be lowercase or else always undefined
+    const tagname = previewElement.tagName.toLowerCase();
```
aomarks, line 510: Can use `localName` instead of `tagName.toLowerCase()`. It's lowercase to begin with.

aomarks (changes requested):

aomarks:
Pretty cool! Seems like a very flexible and easy to use system for adding checking.

packages/lit-dev-content/samples/checkable-tutorial-base.json
e111077: TIL!

aomarks:
Was this rebased? The "changes since my last review" feature doesn't work :(

e111077:
Yeah there was a rebase. I didn't know it'd break that. Sorry will hold on next time

packages/lit-dev-content/src/icons/check-circle-crossed-icon.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: https://fonts.google.com/icons?selected=Material+Icons+Outlined:unpublished
```
aomarks, line 9: URL seems wrong

packages/lit-dev-content/src/icons/loop-icon.ts
aomarks: Link still missing

packages/lit-dev-content/src/icons/loop-icon.ts
```diff
@@ -0,0 +1,21 @@
...
+export const loopIcon = html`<svg
+  xmlns="http://www.w3.org/2000/svg"
```
aomarks, line 11: `xmlns` unnecessary

aomarks (changes requested):

packages/lit-dev-content/src/icons/check-circle-crossed-icon.ts
e111077: it works for me, what do you see? I get this:

image

aomarks: Sorry my mistake, you're right it works fine.

aomarks (approved):

PR #730, opened 2022-03-25
[tutorials] Working with Lists tutorial

### LINKED ISSUE #749: [tutorials] Working with lists tutorial

augustjk:
This tutorial adapts https://lit.dev/docs/templates/lists/ and https://lit.dev/playground/#sample=examples/directive-repeat into an interactive tutorial.

Resolves #749


packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,13 @@
+In this tutorial, we will be exploring different ways of rendering a list of
```
e111077, line 1: no we:

> This tutorial will be exploring...

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,13 @@
...
+items in Lit!
+
+Let's begin by looking a how we might render a list of names from an array.
```
e111077, line 4: Let's not use` Let's` who is the implied `we`? is it the Lit team, the user and us, is it the Lit library itself?

Suggestion:

> Begin my rendering a list of names from an array. The example already has set up a custom element with the `names` property that is an array of strings.

some help:

- commands: just give the command
  - lets begin -> begin
  - we can see -> "See" or "You may notice" or plainly "Notice"
    - prefer the assertive voice rather than "may" so the user can be assured they know that we know what we're teaching them
- we
  - tell the user who we is
  - we -> the example / the code

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,13 @@
...
+We've already set up a custom element with a `names` property that is an
+array of strings.
+
+We can utlize the `map` array prototype method to generate a list item
```
e111077, line 8: Fom here i'll just leave comments saying "we" or "lets" so you can keep track on where this stuff is

e111077, line 8: Perhaps before going into the first action for the user add a section on "prior knowledge: docs link docs link docs link"

like reactive properties. We expect you to know how reactive properties work and handle updates

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,13 @@
...
+for each item in the array within the unordered list (`<ul></ul>`).
+
+```ts
```
e111077, line 11: Tell the user directly to add this too the code and add the filename before the sample:
`### my-element.ts`

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,18 @@
+If we're not working with arrays, we can also create our own array of
```
e111077, line 1: Hmm this set not being an array doesn't seem to be the problem in this step that you're tackling. Lit can still render a Set because it's an iterable e.g. just render `this.items` and see.

The problem being solved here is that you need to run .map on an iterable but .map is only available on Array not the generic iterable. Thus you can use `for...of`.

I'd also consider changing the title of this step to address the crux of lesson like `Using for Loops`

e111077, line 1: we

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,18 @@
...
+templates to render, for example, by using a `for...of` loop.
+
+```ts
```
e111077, line 4: filename

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,18 @@
...
+const templates = [];
+for (const item of this.items) {
+  templates.push(html`<li>${item}</li>`);
+}
+```
+
+It can then be placed inside the template returned from the render method.
+```ts
```
e111077, line 12: filename

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,14 @@
+We've see how we can use a `for ... of` loop to populate an array of templates
```
e111077, line 1: We

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,14 @@
...
+to render. However, Lit also provides a built-in
+[`map`](https://lit.dev/docs/templates/directives/#map) directive for a
+more declarative way of writing templates.
+
+The previous example can be solved by first importing the `map` directive
+```ts
```
e111077, line 7: filename

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,14 @@
...
+import {map} from 'lit/directives/map.js';
+```
+
+then placing the expression below in the template returned by the render method.
+```ts
```
e111077, line 12: filename

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,14 @@
+We've see how we can use a `for ... of` loop to populate an array of templates
+to render. However, Lit also provides a built-in
+[`map`](https://lit.dev/docs/templates/directives/#map) directive for a
+more declarative way of writing templates.
+
+The previous example can be solved by first importing the `map` directive
```
e111077, line 6: make this a direction / command for the reader.

> The previous example can be solved by first importing the `map` directive. Import the `map` directive like so:

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,14 @@
...
+```ts
+import {map} from 'lit/directives/map.js';
+```
+
+then placing the expression below in the template returned by the render method.
```
e111077, line 11: Make it a command:

> Then use the map method in the rendered template:

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,14 @@
+We've see how we can use a `for ... of` loop to populate an array of templates
+to render. However, Lit also provides a built-in
+[`map`](https://lit.dev/docs/templates/directives/#map) directive for a
+more declarative way of writing templates.
```
e111077, line 4: I'd drill into this a bit more. Explain why the map directive exists it's declarative, provides something similar to the .map interface to generic iterables, and makes your code cleaner to read.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,18 @@
...
+return html`
+<ul>
+  ${templates}
+</ul>
+`
+```
```
e111077, line 18: perhaps lead into the next step a bit. something like "it works but is uggo code"

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
+The rendering methods we've seen so far make it so that Lit keeps track of the
```
e111077, line 1: we

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+DOM elements and will update them accordingly when the property items change.
+This is usually fine and efficient but sometimes we may want to control how
```
e111077, line 3: we on this line and the next

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+the each item is tied to the DOM element so that we'd rather have the DOM
+element move rather than updating.
+
+For instance, take a look at the example component we have here where each
```
e111077, line 7: we and on the next line

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+todo item is rendered as a list item with a checkbox. We also have buttons
+to sort these items in ascending and descending alphabetical order. Try
```
e111077, line 9: "Try checking one..." -> "Check one..."

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+checking one of the items and changing the sort order. You'll notice that
+the item text changes but not the position of the checked box. This is
+because Lit is updating the text rather than moving the entire DOM node.
```
e111077, line 12: I feel like this is missing some more exposition on the underlying reason, but I'd trust someone on the core team to provide better wording than I

cc: @justinfagnani @sorvell @kevinpschaaf

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+
+This is where the [`repeat`](https://lit.dev/docs/templates/directives/#repeat)
+directive comes in. We can replace the expression that's currently using
```
e111077, line 15: we

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+the `map` method to utilize `repeat` like this.
+```ts
+import {repeat} from 'lit/directives/repeat.js';
+```
+```ts
+${repeat(
+  this.items,
+  (item) => item.id,
```
e111077, line 23: explain how `ID` works. This is always a big question from users.

Also a useful comparison I've made before was that setting the ID function to return the index is functionally equivalent to `.map`

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,29 @@
...
+  (item) => html`<li><label><input type="checkbox">${item.task}</label></li>`
+)}
+```
+
+Check out the [documentation](https://lit.dev/docs/templates/lists/#when-to-use-map-or-repeat)
+on when to use `map` or `repeat` for more details!
```
e111077, line 29: Maybe add a "congrats you're done" step that tells them what more to read up on next like the templating docs or a topic that you think would logically follow lists and arrays like calling `requestUpdate` or `hasUpdated` or how Lit handles reactive properties changes.

e111077 (changes requested):

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
e111077: oh and add a recap in that section

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: Good idea. How about also adding like a recommend prerequisite tutorial if applicable?

Something like
> Prior knowledge: [Defining a custom element, rendering a template, and reactive properties](http://localhost:5415/docs/components/overview/).
> Recommended prerequisite: [Intro to Lit](https://lit.dev/tutorials/intro-to-lit/).

to be placed under the introductory description of the tutorial.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: As I'm thinking about this change, I'm wondering whether this much spoon feeding is the best way to write interactive tutorials. In my former teaching experience, I'd never want to give away the answer but provide enough clues for students to figure out themselves which makes it more memorable. I know during the kickoff you mentioned people are lazy and want braindead solutions but in that case they can just click the solve button to see the solution. And if they really want to just see working code, they can get that in the doc section or the completed playground examples.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
e111077: That's the idea! Hopefully once we get a bunch of tutorials up i can go through and link to each other

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
e111077: It might be doable skipping that for this tutorial since there's only one JS file and one HTML file. Other tutorials have multiple files and due to an oddity on how playground-ide works, it sometimes points to weird files

augustjk: Oh, I'm happy to add filenames for snippets but I was more getting at not providing copy-pasteable code. Prompt can say use the `map` prototype method to render an array of lit-html templates that wrap each item of the array in a list item tag, and the reader has to come up with the code based on that. Maybe I'm thinking too much like a learn-how-to-code classroom kind of tutorial where making the learners think seem more important.

e111077: also works. As long as it's glance-able

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,68 @@
...
+)}
+```
+
+{% endswitchable-sample %}
+
+Confirm that this is working correctly by checking an item and changing the
+sort order. The check mark should now be moving to be next to the originally
+checked list item.
+
+Keep in mind that the key function provided must return a unique key for an
```
e111077, line 65: Remove "Keep in mind that" and maybe change this to a positive aside:

```md
<aside class="positive">

The key function provided must return a unique key for an item.

A key function that simply returns the index would have the same behavior as using `map`. See the [documentation](/docs/templates/directives/#repeat)
for more details.

</aside>
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,68 @@
...
+item. A key function that simply returns the index would have the same behavior
+as using `map`. See the [documentation](https://lit.dev/docs/templates/directives/#repeat)
```
e111077, line 67: remove `https://lit.dev` and make this an absolute link rather than a full URL

e111077 (changes requested):

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
augustjk: Thanks. Updated here and for all other internal links in this tutorial.

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
augustjk: Done

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,51 @@
...
+    ${templates}
+  </ul>
+`
+```
+
+{% endswitchable-sample %}
+
+This is especially useful when you're not directly working with a single
+iterable and are buliding up a list from multiple sources. However, if working
+with a single iterable, there is more declarative way of writing this shown in
```
AndrewJakubowicz, line 50: Nit: "there is **a** more"

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
AndrewJakubowicz: Thanks Elliott! The fact that `map` works on the `Iterable` type is quite powerful :P

[A quick playground experiment](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy5qcyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcbmltcG9ydCB7bWFwfSBmcm9tICdsaXQvZGlyZWN0aXZlcy9tYXAuanMnO1xuXG5mdW5jdGlvbiogd2lsZEl0ZXJhYmxlKCkge1xuICBmb3IgKGxldCBpID0gMDsgaSA8IDQ7IGkrKykge1xuICAgIHlpZWxkIGk7XG4gIH1cbn1cblxuZXhwb3J0IGNsYXNzIFNpbXBsZUdyZWV0aW5nIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIHJlbmRlcigpIHtcbiAgICByZXR1cm4gaHRtbGA8dWw-JHttYXAod2lsZEl0ZXJhYmxlKCksICh2KSA9PiBodG1sYDxsaT4ke3Z9PC9saT5gKX08L3VsPmA7XG4gIH1cbn1cbmN1c3RvbUVsZW1lbnRzLmRlZmluZSgnc2ltcGxlLWdyZWV0aW5nJywgU2ltcGxlR3JlZXRpbmcpO1xuIn0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8IURPQ1RZUEUgaHRtbD5cbjxoZWFkPlxuICA8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL3NpbXBsZS1ncmVldGluZy5qc1wiPjwvc2NyaXB0PlxuPC9oZWFkPlxuPGJvZHk-XG4gIDxzaW1wbGUtZ3JlZXRpbmcgbmFtZT1cIldvcmxkXCI-PC9zaW1wbGUtZ3JlZXRpbmc-XG48L2JvZHk-XG4ifSx7Im5hbWUiOiJwYWNrYWdlLmpzb24iLCJjb250ZW50Ijoie1xuICBcImRlcGVuZGVuY2llc1wiOiB7XG4gICAgXCJsaXRcIjogXCJeMi4wLjBcIixcbiAgICBcIkBsaXQvcmVhY3RpdmUtZWxlbWVudFwiOiBcIl4xLjAuMFwiLFxuICAgIFwibGl0LWVsZW1lbnRcIjogXCJeMy4wLjBcIixcbiAgICBcImxpdC1odG1sXCI6IFwiXjIuMC4wXCJcbiAgfVxufSIsImhpZGRlbiI6dHJ1ZX1d)

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,33 @@
+This tutorial will be exploring different ways of rendering a list of
+items in Lit!
+
+## Previous knowledge
+It's recommended to have an understanding of
+[defining a component](/docs/components/defining/),
+[rendering](/docs/components/rendering/),and
+[reactive properties](/docs/components/properties/) before
+working on this tutorial.<br>
+Recommended prerequisite: [Intro to Lit](/tutorials/intro-to-lit/).
```
kevinpschaaf, line 10: Nit: Found it odd (both language and format) to have a sentence above recommending certain docs, and then a non-sentence immediately below it recommending more. Maybe format as a bulleted list, or else just add the tutorial to the sentence?

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,70 @@
...
+Confirm that this is working correctly by checking an item and changing the
+sort order. The check mark should now be moving to be next to the originally
+checked list item.
+
+<aside class="positive">
+The key function provided must return a unique key for an item. A key function
+that simply returns the index would have the same behavior as using
+<code>map</code>. See the <a href="/docs/templates/directives/#repeat">
+documentation</a> for more details.
+</aside>
```
AndrewJakubowicz, line 70: This fourth step is extremely good. I love that the initial code state is broken in a very clear way (which is described). Then the explanation and solution makes perfect sense and solves the original issue.

And TIL!

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+## Rendering a list
+Begin by rendering a list of names from an array. The example already has set
+up a custom element with the `names` property that is an array of string.
```
kevinpschaaf, line 15:
```suggestion
up a custom element with the `names` property that is an array of strings.
```
(or "of type string").

AndrewJakubowicz:
Went through the tutorial and it's great!!! Only nit is that the `map` directive can be expanded.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,51 @@
+When not working with an array as a source of data, an array of templates to
+render can be manually built, for example, by using a `for ... of` loop.
+
+{% switchable-sample %}
+
+```ts
+// my-element.ts
+const templates = [];
```
kevinpschaaf, line 8: More stylistic question for @e111077: I had my snippets include the indentation needed so that after copy/paste you didn't manually indent them to make it look ok. But it does make the snippets look kinda weird since they're floating right more. Maybe something to decide and normalize once all the tutorials are committed.

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,36 @@
+The previous step showed how a `for ... of` loop can be used to populate an
+array of templates to render. However, Lit also provides a built-in
+[`map`](/docs/templates/directives/#map) directive for a
+more declarative way of writing templates.
```
kevinpschaaf, line 4: +1
```suggestion
array of templates to render. However, Lit also provides a built-in
[`map`](/docs/templates/directives/#map) directive that will map
over any iterable, including `Map`s, `Set`s, and even
generators, in addition to arrays.
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,36 @@
...
+
+The previous example can be solved by first importing the `map` directive.
```
kevinpschaaf, line 6:
```suggestion
The previous example can be simplified by first importing the `map` directive.
```
?

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,51 @@
+When not working with an array as a source of data, an array of templates to
+render can be manually built, for example, by using a `for ... of` loop.
```
kevinpschaaf, line 2: I didn't notice the example had changed from an array to a `Set` at first, so might be worth calling attention to that change first?
```suggestion
Sometimes you may want to render a list from something other than
an array, such as a `Set`. Note we've changed the example
data from an array to a `Set`.

You can manually manually build an array of templates from the `Set`
by using a `for ... of` loop, for example.
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,70 @@
+The rendering methods shown so far make it so that Lit keeps track of the
+DOM elements and will update them accordingly when the property items change.
```
kevinpschaaf, line 2: Suggestion on wording improvement:
```suggestion
When rendering an array, Lit will create a list of DOM nodes and update them
in the order of the data in the array (reusing any previously rendered nodes) 
eacn render.
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
arthurevans: I think this is a really interesting idea. I think Tour of Go, for example, does things like this--poses a problem and lets you solve it. And I think we could do that--especially with the addition of the Check button. 

It might make sense, for example, to add a separate "Exercise" step (or steps in a longer tutorial) where you give general instructions and let the users work out a solution. (With the backup of the Check/Solve buttons.) 

Another way to do this is by adding suggestions--typically at the end of a step or the end of a tutorial--of things the user could try. For example, the "Extra credit" items at the end of Steve's tooltip tutorial: https://pr740-77daa6d---lit-dev-5ftespv5na-uc.a.run.app/tutorials/tooltip?mods=tutorialCatalog#09

Encouraging people to play with the code + giving that extra reward of figuring it out yourself. As you said, it really helps make the information stick.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,33 @@
...
+
+The `map` array prototype method can be utilized to generate a list item
```
sorvell, line 17: Def defer to @arthurevans here but I think we want to try to avoid passive voice and be more direct. Perhaps something like "Use Array's `map`  method to generate a list..."

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
sorvell: It might also be nice to have a  really brief outline here to setup what's going to happen.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,33 @@
...
+{% switchable-sample %}
+
+```ts
+// my-element.ts
+${this.names.map((name) => html`<li>${name}</li>`)}
+```
+
+```js
+// my-element.js
+${this.names.map((name) => html`<li>${name}</li>`)}
```
sorvell, line 30: A little more description of what's happening here would be great. In particular, this is leveraging Lit's (super power?) ability easily to compose templates.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
sorvell: +1

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,51 @@
...
+// my-element.js
+const templates = [];
+for (const item of this.items) {
+  templates.push(html`<li>${item}</li>`);
+}
+```
+
+{% endswitchable-sample %}
+
+It can then be placed inside the template returned from the render method.
```
sorvell, line 24: This might be too advanced but there's an opportunity here to mention (perhaps in an aside) that Lit templates can host any javascript expression, but if you're using statements like this, you can build up a result to later place in a template (statements can also be easily used by returning a result from a function... hopefully there's a link in the docs we could just call out to?).

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+```js
+// my-element.js
+${map(this.items, (item) => html`<li>${item}</li>`)}
+```
+
+{% endswitchable-sample %}
+
+The index of the item is also made available to the callback function. Check
```
sorvell, line 39: Let's try to add some guidance (re-emphasizing the point in the previous slide) about when you should do one way v. the other. E.g. something like... Use the `map` directive for simple cases that easily lend themselves to declarative description; use imperative code for more complex situations.

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,70 @@
...
+This is usually fine and efficient but sometimes it's preferable to control how
```
sorvell, line 3: I think this section needs more careful treatment. In particular, the code shown here demonstrates a BUG and should be called out as such. 

We need to clearly associate this with state being stored in the DOM that's not managed by Lit, as is shown here.

My concern is that otherwise, this seems pretty scary and like a footgun.

sorvell: ^^ We explain this in the docs directly and you call it out in the next slide, but I think it should be mentioned right away up front. Otherwise, again, it seems scary.

sorvell (changes requested):

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
e111077: TBH I think it's fine removing the indents if it makes it look weird but i gotta say was nice for speed running your tutorial

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,33 @@
...
+It's recommended to have an understanding of
+[defining a component](/docs/components/defining/),
+[rendering](/docs/components/rendering/), and
+[reactive properties](/docs/components/properties/) before
+working on this tutorial. The [Intro to Lit](/tutorials/intro-to-lit/) tutorial
+is a great place to start.
+
+
+## Rendering a list
+Begin by rendering a list of names from an array. The example already has set
```
justinfagnani, line 14: I think there's a bit of conceptual info that could be added here to make list support click for devs.

"List" support is really a combination of two things: lit-html supports nested template results as values, and it supports iterables of any supported value type. Combine those and you get iterables of template results.

It might be nice to start with something _very_ simple, like:

```ts
  render() {
    return html`
      <p>A list of my friends</p>
      ${['Chandler', 'Phoebe', 'Joey', 'Monica', 'Rachel', 'Ross']}
    `;
  }
```

And then say that you might want to include markup to you can do a bulleted list, so you can transform an array with `.map()`...

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,55 @@
+Sometimes there may be a need to render a list from something other than an
```
justinfagnani, line 1: I'd mention "iterable" in here, since that's what Lit supports, so like "render a list from another type of iterable, such as a `Set`"

justinfagnani, line 1: I think there's a question of ordering here... is .map(), imperative, map(), repeat(), the best order? map() is probably 99% of the time more readable than imperative, so maybe it should come first?

packages/lit-dev-content/site/tutorials/content/working-with-lists/04.md
```diff
@@ -0,0 +1,13 @@
+Congratulations on completing the **Working with Lists** tutorial!
```
justinfagnani, line 1: I think two more steps should be: events (talked on chat about that) and heterogenous lists. Since Lit just support iterables of any renderable value, the template results in an array don't have to be the same. So you can conditionally render different templates.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: Combined the "Previous knowledge" section into sentences, and added a bulleted list of "What you'll learn"

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
augustjk: Fixed!

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: Elliott has pointed this out in other places too. Reworded.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: Good call, especially since this one's geared toward beginners. Added a note below the code snippet.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
augustjk: Added an aside stating any javascript expressions resulting in valid values for expressions with link to doc.

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
augustjk: Yes, that's good direction. Added.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: Yes, that's a great suggestion! I did see that in the Lists doc section but didn't want to dedicate a whole step. Starting the example with that is a great way to bring it into the conversation.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
augustjk: That makes sense. I felt that starting with `map()` then reverting to imperative for the set would be silly, but perhaps the imperative step could be combined with the heterogeneous list idea below and to follow `map()`? What do you think about the placing of the events step? Before or after `repeat()`? I feel it could go either way, but leaning toward after.

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
augustjk: The first paragraph was reworded based on Kevin's earlier suggestion and I also expanded the explanation in the following paragraph that explains the current buggy behavior of the example mentioning that the checkbox state lives with the DOM elements that didn't move.

packages/lit-dev-content/samples/tutorials/working-with-lists/02/after/my-element.ts
```diff
@@ -0,0 +1,49 @@
...
+      templates.push('🐶 Pets 🐱');
+    }
+    if (this.includePets) {
+      for (const pet of this.pets) {
+        templates.push(html`<li>${pet}</li>`);
+      }
+    }
+
+    return html`
+      <button @click=${() => this._togglePetVisibility()}>${this.includePets ? 'Hide' : 'Show'} pets</button>
```
e111077, line 33: This looks messy in JS it looks like this:

```ts
return html`
   <button @click=${() => this._togglePetVisibility()}>${
   this.includePets ? 'Hide' : 'Show'
 } pets</button>
   <button @click=${() => this._toggleSeparator()}>${
   this.includeSeparator ? 'Hide' : 'Show'
 } separator</button>
   <p>My family</p>
   <ul>
     <!-- TODO: Render templates. -->
   </ul>
 `;
```

Maybe put the button content on a newline like:

```ts
html`
<button @click=${() => this._togglePetVisibility()}>
  ${this.includePets ? 'Hide' : 'Show'} pets
</button>
...
`
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,93 @@
+The methods used in the previous steps are great when working with a single
+iterable as a source of data but sometimes the situation might call for a more
+imperative approach.
+
+In this example, we have a string array property `members`, a string array
+property `pets`, a boolean state `includePets` controlled by a button, and
+another boolean state `includeSeparator` also controlled by a button. The
+desired result is to render a list item for each member but also include pets
```
e111077, line 8: The readers YEARN for instruction! suggest changing:

> The desired result is to render a list...

to:

> (insert newline) Render a list... so that...

Maybe even make it a bulleted list of tasks & requirements

also because writing the sentence in the infinitive case ("to render...") leads to weird sentence structure

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,93 @@
...
+if (this.includePets) {
+  for (const pet of this.pets) {
+    templates.push(html`<li>${pet}</li>`);
+  }
+}
+```
+
+{% endswitchable-sample %}
+
+It can then be placed inside the template returned from the render method.
```
e111077, line 51: Take command of this ship, captain!

> It can then be placed...

to:

> Now place...

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,93 @@
...
+Remember that any valid JavaScript expressions can be placed within the tagged
+template literal that result in valid renderable values described
+[here](/docs/templates/expressions/). This could be a variable
+containing an array of `TemplateResult` and strings as done here, an array
+method call returning an array like seen in the previous step, or even an
+invocation of a custom written function that returns something renderable.
+
+<br>
+
+Try refactoring the code here to abstract out the logic from the `render` method
```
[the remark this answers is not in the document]
e111077: Also what is "here"

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,75 @@
+When rendering an array, Lit will create a list of DOM nodes and update them
+in the order of the data in the array (reusing any previously rendered nodes) 
+each render. This is usually fine and efficient but sometimes it's preferable
+to control how each item is tied to the DOM element so that the DOM element is
+moved if needed rather than its content updated.
+
+For instance, take a look at the example component presented here where each
+todo item is rendered as a list item with a checkbox. There are buttons to sort
+these items in ascending or descending alphabetical order. Check one of the
```
e111077, line 9: > these items...

to

> the list items...

e111077, line 9: > Check one of the items...

This is a command. Start a new paragraph

> (newline) Check one of the items in the preview...

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,75 @@
...
+items and change the sort order. Notice that the item's label text changes but
+not the position of the checked box. This is because Lit keeps track of the
+list of DOM nodes with the checkbox state as they are and just updates the
+values that go into the expression. Ss the checkbox state is not controlled by
```
e111077, line 13: remove `Ss`

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,75 @@
...
+Lit, only the label text changes.
```
e111077, line 14: > Lit, only the label text changes.

comma splice. Add conjunction or change comma to a semi-colon or an em-dash (–)

> Lit, thus only the label text changes.

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,75 @@
...
+
+{% endswitchable-sample %}
+
+Confirm that this is working correctly by checking an item and changing the
+sort order. The check mark should now be moving to be next to the originally
+checked list item.
+
+<aside class="positive">
+
+The key function provided must return a unique key for an item. A key function
```
e111077, line 69: if you use litdev-aside. The first sentence could be its own line and be bolded

packages/lit-dev-content/site/tutorials/content/working-with-lists/03.md
```diff
@@ -0,0 +1,75 @@
...
+list of DOM nodes with the checkbox state as they are and just updates the
+values that go into the expression. Ss the checkbox state is not controlled by
+Lit, only the label text changes.
+
+If the intention is to keep the list item and checkbox tied to the specific
+item of the array, the `repeat` directive can be of help. Using this directive
+with a key function lets Lit maintain the key-to-DOM association between
+updates by moving the generated DOM when required.
+
+To update the current implementaion, start by importing the `repeat` directive.
```
e111077, line 21: remove

> To update the current implementation,

packages/lit-dev-content/site/tutorials/content/working-with-lists/04.md
```diff
@@ -0,0 +1,69 @@
...
+is being rendered as list items. There's a state called `lastIndex` that should
+be updated to the index of the last list item that the mouse was over.
+
+To achieve this, first create a method that will be called when the `mouseenter`
+happens that will take a new index as its parameter.
+```ts
+// my-element.ts
+class MyElement extends LitElement {
+  ...
+  private _updateLastIndex(newIndex: number) {
```
e111077, line 15: You forgot to wrap all of these in the switchable samples macro

packages/lit-dev-content/site/tutorials/content/working-with-lists/04.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+Now mousing over the different city names should update the one displayed up
+top. Note that an inline arrow function is used here to create closures such
+that each list item gets a function that calls `_updateLastIndex` with its own
+index.
+
+<aside class="info">
+
+While closures will be fine for most cases, there are other ways to achieve
```
e111077, line 58: If you have to use a `<br>` consider finding a way to make your aside content more concise or merge ideas

I think `<litdev-aside>` is good for this because it bolds the "topic" of the aside and forces you to keep these to a single idea

e111077 (changes requested):

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,46 @@
+Sometimes you may want to render a list from another type of iterable,
+such as a `Set`. Note the data in this example has been changed from an
+array to a `Set`.
+
+Lit provides a built-in `map` directive that will map over any iterable
+including `Map`s, `Set`s, and even generators, in addition to arrays.
```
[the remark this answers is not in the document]
arthurevans: Also kind of awkward to say, "a map directive that will map over ... maps"
When pluralizing a class name like this, I usually try to either make it generic ("maps, sets, and arrays"), add a noun like "objects": "`Map` and `Set` objects", or rephrase the sentence to use the singular. 

Maybe: "... a built-in `map` directive that works on any iterable: arrays, maps, sets, and even generators."

arthurevans (commented):
I had some pending comments which may or may not be outdated. Submitting now seems to the only valid escape from this realm of uncertainty. Please ignore if outdated.

arthurevans:
Hi @augustjk ... I think you were going to do some structural changes here so I'm holding off on further reviews ATM. Please ping me when you're ready for another review.

augustjk:
> Hi @augustjk ... I think you were going to do some structural changes here so I'm holding off on further reviews ATM. Please ping me when you're ready for another review.

@arthurevans Yes, thanks so much for the comments! I am indeed working on changes as discussed offline but I will incorporate the suggestions as I move the steps around. Will ping when it's ready for review again.

justinfagnani:
Is something preventing new previews from being deployed? The last preview link is from 4 days ago, but usually there's at least a new link right away for a new push. Maybe it's the draft status? @aomarks ?

augustjk:
> Is something preventing new previews from being deployed? The last preview link is from 4 days ago, but usually there's at least a new link right away for a new push. Maybe it's the draft status? @aomarks ?

Maybe it is the draft status. Switching it back didn't automatically add it though, had to manually re-trigger.

augustjk:
> > Is something preventing new previews from being deployed? The last preview link is from 4 days ago, but usually there's at least a new link right away for a new push. Maybe it's the draft status? @aomarks ?
> 
> Maybe it is the draft status. Switching it back didn't automatically add it though, had to manually re-trigger.

Doh, it just re-added the same url from before. I'll get to my machine and push up another commit to re-trigger.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,24 @@
+This tutorial will be exploring different ways of rendering a list of items in
```
justinfagnani, line 1: Do we want to say "different ways of rendering a list" or "different ways of rendering lists"?

You can render more than one list at once, so I'd lean towards plural. @arthurevans ?

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,24 @@
...
+Lit!
+
+Lit makes it easy to render any lists of items. It has built-it support for
+iterables of any supported values. This, along with Lit's ability to render
+nested templates and helpful built-in directives, allow for rendering any
+variety of lists.
```
justinfagnani, line 7:
```suggestion
Lit makes it easy to render lists of items, with has built-it support for
iterables: pass an array (or other iterable) to a child expression and Lit will render each item in the array. This, combined with Lit's ability to render nested templates, and JavaScript patterns for transforming lists of data into lists of templates, allows for very flexible list handling.
```

^ Not sure if this is too complicated or not, but I thought it might summarize list support quickly.

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,24 @@
...
+
+## What you'll learn
+- Utilizing the `map` directive
+- Using array prototype methods
+- Building static list and using the `range` directive
```
justinfagnani, line 12:
```suggestion
- Building static lists and using the `range` directive
```

justinfagnani, line 12: The term "static list" might not be immediately apparent. Maybe "Generating lists with the range directive"

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
```diff
@@ -0,0 +1,24 @@
...
+- Imperatively building an array of items to render
+- Utilizing the `repeat` directive
```
justinfagnani, line 14: "Utilizing" is repeated here. I wonder if we want to add more description of what readers will do with each thing, like:

```
 - Rendering lists with the map() directive
 - Transforming arrays into templates with array.map()
 - Generating lists with the range() directive
 - Imperatively building arrays of items
 - Diffing lists of stateful elements with the repeat() directive
 - Using event listeners in a list

**Review by justinfagnani — CHANGES_REQUESTED** [review]

File: packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md [hunk 58]
```diff
@@ -0,0 +1,66 @@
...
+```js
+// my-element.js
+import {map} from 'lit/directives/map.js';
+```
+
+{% endswitchable-sample %}
+
+Use the directive in the rendered template, providing a callback function that
+returns a template wrapping each item with a list item (`<li></li>`) tag, and
+place it within the unordered list (`<ul></li>`) of the component's template.
```
justinfagnani, line 29:
```suggestion
place it within the unordered list (`<ul></li>`) in the component's template.
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+```js
+// my-element.js
+import {map} from 'lit/directives/map.js';
+```
+
+{% endswitchable-sample %}
+
+Use the directive in the rendered template, providing a callback function that
```
justinfagnani, line 27: I don't know about "rendered template" - all templates are ultimately rendered, yeah? So just "template" or maybe "component's template" if you want something more?

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: @arthurevans do you have thoughts on whether we should add the `()` after the directive names? Scanning through https://lit.dev/docs/templates/directives/ has examples of both with and without the parens.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
justinfagnani: You could also just refer to the `<ul>` tag rather than "the unordered list":
```suggestion
place it inside the `<ul>` tag in the template.
```

justinfagnani: Oh, and `</li>` needs to be updated
```suggestion
place it within the unordered list (`<ul></ul>`) of the component's template.
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,66 @@
...
+returns a template wrapping each item with a list item (`<li></li>`) tag, and
+place it within the unordered list (`<ul></li>`) of the component's template.
+
+{% switchable-sample %}
+
+```ts
```
justinfagnani, line 33: Do we need separate ts/js samples when they're exactly the same? @aomarks

packages/lit-dev-content/site/tutorials/content/working-with-lists/02.md
```diff
@@ -0,0 +1,48 @@
+Any valid JavaScript expressions that result in valid renderable values
+described in the [expressions documentation](/docs/templates/expressions/) can
+be placed within the tagged template literal. When working with arrays, their
```
justinfagnani, line 3: so the expressions have to be placed in template literal _expressions_. Saying that would be three "expressions" in one sentence though. Maybe the sentence could be reordered:

Lit template literal expressions can contain any JavaScript that returns a renderable value as described in...

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
```diff
@@ -0,0 +1,66 @@
+While Lit allows rendering values straight out of an iterable, it would be nice
+to have the ability to apply some transformation like adding markup for each
+item. Lit provides a built-in `map` directive that works on any iterable such as
```
justinfagnani, line 3:
```suggestion
Lit supports rendering iterables of items directly in templates, but you often need to render a _template_ that uses the items, rather than the items directly. Lit provides a built-in `map` directive that lets you transform the items in an iterable, usually to create a template from them. The map directive works with any iterable such as...
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
augustjk: For most of these, I've added separate samples because of the first line comment that names the file and wanted the file extensions to match the selected language. I had a comment about adding the file name for where the code snippet should go, though arguably since there's only ever 1 file you're editing for this tutorial it could be omitted.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
augustjk: I like the shorter "`<ul>` tag".

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
arthurevans: That's a good question. Originally I thought we were going to go with parens after both method and directive names. The current state is inconsistent for both, unfortunately. I lean towards using the parens everywhere. But at any event, make sure they're consistent in your tutorial. I'll put this on the agenda for a future docs meeting.

e111077: I'm personally against parens even for methods because the arguments may be important and it'd be kinda silly to include the method signature when simply referring to it

arthurevans: Yeah, using empty parens is kind of convention for marking, "this is a function/method". It's recommended by the Google developer docs style guide, which we approximately follow:

https://developers.google.com/style/code-in-text#methods

Back in the polymer days, I bucked this trend because I'd worked on Java interfaces in the past and the JavaDoc rule is that you don't include empty parens--because of the way Java function overloading works, `foo()` may literally be a different function than `foo(bar)`. 

At present, it seems like like the JS world is divided on the value of the parens. React docs appear to use them: https://reactjs.org/docs/rendering-elements.html Vue docs don't. Svelte seems to use them for platform APIs like `event.preventDefault()`, but not for Svelte's own APIs, which seems odd. (Also, they prefix their own method names with a `.` to I guess show it's a member of an object, which also looks odd to me, so let's not do that.)

I don't have a super strong feeling about this one other than a general desire to fall in line with the developer docs style guide. But whatever we do, we should do it consistently.

augustjk: I can see arguments for both but lean towards @e111077's reasoning. I currently have consistently no parens within this tutorial and will keep it that way for now then.

Semi related to this, is do we have patterns for adding `.` in front of prototype methods like when referring to `.map`, `.filter` for arrays? For Step 3 of this tutorial, I've added the `.` if anything at least to make it stand out from the `map` directive.

arthurevans: We've never used that pattern that I'm aware of. It looks like it's just the one place you mention them, so I'd suggest saying something like: "the `filter` or `map` array methods". It seems to me that `.map` might be reasonably clear if the reader was aware you were trying to make the distinction between the map array method and the map directive. 

But the dot is easy to miss and not necessarily transparent to a more novice programmer. I'd suggest erring on the side of being a bit more explicit.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
e111077: I'd say it's okay to omit the switchable-sample macro in these cases

packages/lit-dev-content/site/tutorials/content/working-with-lists/00.md
augustjk: Ooh I did not see that comment about it being in the Google developer docs style guide. I'd be happy to lean on that for this tutorial then. Thanks for the link! I'll also update the array method portion.

packages/lit-dev-content/site/tutorials/content/working-with-lists/01.md
augustjk: Should I do
```
// my-element.(js|ts)
```
for the first line comment in the samples then, or omit the filename comment all-together for this tutorial since the reader is only ever working on the one file?

Maybe this is another thing that we'd want a consistency pass across all tutorials as well. @arthurevans what do you think?

justinfagnani: Oh... that's enough of a difference to keep the two then. I missed that and thought they were identical.

e111077 (changes requested):

packages/lit-dev-content/samples/tutorials/working-with-lists/04/before/my-element.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    #board {
+      display: grid;
+      grid-template-columns: repeat(8, 12.5%);
+      grid-template-rows: repeat(8, 12.5%);
+      border: 2px solid #404040;
+      box-sizing: border-box;
+      height: 100%;
+    }
+    .square {
```
sorvell, line 22: Remove `.square` completely here as this isn't included in the instructions and breaks the step.

packages/lit-dev-content/samples/tutorials/working-with-lists/04/before/my-element.ts
```diff
@@ -0,0 +1,50 @@
...
+  static styles = css`
+    /* playground-fold */
+    :host {
+      display: block;
+      width: 400px;
+      height: 400px;
+    }
+    #board {
+      display: grid;
+      grid-template-columns: repeat(8, 12.5%);
```
sorvell, line 16: Can we remove the hard-coded `8`'s here and elsewhere in this module? It should either be a `const size = 8` at the top of the module or settable in the render method and passed to css via a custom property. I also think this can just be `1fr` instead of `12.5%`.

packages/lit-dev-content/site/tutorials/content/working-with-lists/05.md
```diff
@@ -0,0 +1,104 @@
+When using the `map()` directive or just providing an iterable within the
+template, Lit creates a list of DOM nodes and updates them in the iteration
+order, efficiently reusing any previously rendered nodes. This is **good** and
+should be utilized when **all states are being controlled by Lit**.
```
sorvell, line 4: How about: `These techniques should only be used when Lit controls all of the rendered state as is commonly the case.`

packages/lit-dev-content/site/tutorials/content/working-with-lists/05.md
```diff
@@ -0,0 +1,104 @@
...
+
+However, when relying instead on the DOM nodes to keep some state that are not
```
sorvell, line 6:
```suggestion
However, when relying instead on the DOM nodes to maintain state that is not
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/05.md
```diff
@@ -0,0 +1,104 @@
...
+controlled by Lit, buggy behavior could arise. This is an example of when the
```
sorvell, line 7:
```suggestion
controlled by Lit, unexpected behavior can occur, as we'll see. In this case, use the `repeat()` directive instead.
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/05.md
```diff
@@ -0,0 +1,104 @@
...
+`repeat()` directive should be utilized.
```
sorvell, line 8:
```suggestion

```

packages/lit-dev-content/site/tutorials/content/working-with-lists/06.md
```diff
@@ -0,0 +1,112 @@
+There are several ways of adding event listeners on each item getting rendered
```
sorvell, line 1:
```suggestion
There are several ways of adding event listeners to list items
```

packages/lit-dev-content/site/tutorials/content/working-with-lists/06.md
```diff
@@ -0,0 +1,112 @@
...
+from a list, but care must be taken to ensure the correct information is made
```
sorvell, line 2:
```suggestion
, but care must be taken to ensure the correct information is made
```

sorvell (changes requested):
This is getting really good. It just needs a few more changes. See comments and, in addition, I recommend:
* replace all uses of `utilize` with `use` (see this [explanation](https://grammarist.com/grammar/use-vs-utilize/))
* avoid using `states` plural and instead refer to `the state` as singular

packages/lit-dev-content/samples/tutorials/working-with-lists/04/before/my-element.ts
augustjk: I agree that as a finished product, I'd like to not hardcode the 8. Making it settable and passing into css via custom property sounds great. I think that would be great as a playground example too.

But as a tutorial step, I feel that may be too much. I initially had them as variables but thought it added extra noise in the instruction having to say `rows` by `columns` size board, instead of a fixed 8 by 8 board, and especially for the paragraph under the code sample that mentions `range()` generating integers from 0 to 7. Elliott has already mentioned concerns about this section having a lot going on. What do you think?

Perhaps it can be mentioned as extra credit for this step to make it more versatile. @justinfagnani and @e111077 have been talking about a way to add optional extra steps to these tutorials. Or maybe linking out to a playground example for a solution to the extra credit?

Agreed on the `1fr` which would work even if the size was dynamic.

packages/lit-dev-content/site/tutorials/content/working-with-lists/04.md
```diff
@@ -0,0 +1,79 @@
...
+{% switchable-sample %}
+
+```ts
+// my-element.ts
+  render() {
+    return html`
+      ⋮
+      <div id="board">
+        ${map(range(8), (row) => map(range(8), (col) => html`
+          <div class="${getColor(row, col)}">${getLabel(row, col)}</div>
```
arthurevans, line 43: I think this is missing `.square` here.

```suggestion
          <div class="square ${getColor(row, col)}">${getLabel(row, col)}</div>
```

augustjk: I actually dropped the square class since all it was doing was adding some padding, and selected for `#board`'s children to add that styling instead. The solution already reflects this but I forgot to change the before codes. This will be fixed in my next push.

e111077 (approved):

Issue #746, opened 2022-04-07
StyleInfo link in generated docs doesn't work


AndrewJakubowicz:
The generated link to the return type `StyleInfo`, doesn't work.

Reproduce by navigating to: https://lit.dev/docs/api/directives/#styleMap
Click `StyleInfo`.

Url changes to https://lit.dev/docs/api/directives/#StyleInfo and page scrolls to top.

Issue: There is no `StyleInfo` id on that page.

AndrewJakubowicz:
Ah it's because of the anchorFilter for the directives page defined in `lit-2.ts`.

@aomarks - Is it ok to relax this filter, or maybe expand it to also include StyleInfo (interfaces)?

aomarks:
> Ah it's because of the anchorFilter for the directives page defined in `lit-2.ts`.
> 
> @aomarks - Is it ok to relax this filter, or maybe expand it to also include StyleInfo (interfaces)?

Maybe? TBH I can't remember the reason for that filter. There must otherwise be things showing up under directives that don't make sense?

aomarks:
> > Ah it's because of the anchorFilter for the directives page defined in `lit-2.ts`.
> > @aomarks - Is it ok to relax this filter, or maybe expand it to also include StyleInfo (interfaces)?
> 
> Maybe? TBH I can't remember the reason for that filter. There must otherwise be things showing up under directives that don't make sense?

I would say go ahead and relax it, and we can take a look at what the result is.

Issue #344, opened 2021-04-23
RefOrCallback link does not lead to API docs


Christian24:
Hello,

On the page about builtin directives (https://lit.dev/docs/api/directives/#ref) the link for `RefOrCallback` does not lead to the API for `RefOrCallback`. It instead just leads to the top of the page.

AndrewJakubowicz:
Thank you! This has been addressed. https://lit.dev/docs/api/directives/#RefOrCallback now links correctly.

PR #747, opened 2022-04-09
Remove function filter from generated directive page.

AndrewJakubowicz:
Note: Made this PR to test what changes without filter and start a conversation about this change.

Fixes: https://github.com/lit/lit.dev/issues/746

Also fixes: https://github.com/lit/lit.dev/issues/344

### Context

The filter was stripping the id attribute from non function types. This prevents linking to the directive classes and also interfaces such as `StyleInfo` (how this issue was discovered).

Removing the `id` attribute prevents fragment links to the symbol, which also prevent indexing the symbol in the Lit.dev search bar. A downside, is that any jsdoc `{@link <symbol>}` syntax will still generate a link, even if the fragment is missing.

I personally think it's worthwhile linking to all the symbols, as it makes using `{@link}` syntax more powerful. However maybe we want to instead prevent the search from indexing internal symbols that we don't want to show up?

### Alternatives considered

We can replace the anchor filter with: `anchorFilter: (node) => node.kindString !== 'Class'`. This slightly relaxes the current filter by allowing interfaces without adding all the directive classes.

However the limitation is that there are still broken links created by this, and issue https://github.com/lit/lit.dev/issues/344 would not be fixed via this path.

### Next steps (which we can block merging this PR on).

The main impact I can see of relaxing this filter is that `ref` and `Ref` appear in the search. Since both the function and class appear for each directive. Maybe we want to filter out the classes from the search index, without removing the id which breaks links.

### Tested

Tested manually and via the preview link. Noted that the directive classes are now searchable via the lit.dev search bar, and the StyleInfo links work now.

### DIFF
--- a/packages/lit-dev-api/api-data/lit-2/pages.json
+++ b/packages/lit-dev-api/api-data/lit-2/pages.json
@@ -8731,6 +8731,10 @@
             "moduleSpecifier": "lit/directives/async-append.js"
           }
         ],
+        "location": {
+          "page": "directives",
+          "anchor": "AsyncAppendDirective"
+        },
         "heritage": [
           {
             "type": "reference",
@@ -9382,6 +9386,10 @@
             "moduleSpecifier": "lit/directives/async-replace.js"
           }
         ],
+        "location": {
+          "page": "directives",
+          "anchor": "AsyncReplaceDirective"
+        },
         "heritage": [
           {
             "type": "reference",
@@ -9696,6 +9704,10 @@
             "moduleSpecifier": "lit/directives/cache.js"
           }
         ],
+        "location": {
+          "page": "directives",
+          "anchor": "CacheDirective"
+        },
         "heritage": [
           {
             "type": "reference",
@@ -10152,6 +10164,10 @@
             "moduleSpecifier": "lit/directives/class-map.js"
           }
         ],
+        "location": {
+          "page": "directives",
+          "anchor": "ClassMapDirective"
+        },
         "heritage": [
           {
             "type": "reference",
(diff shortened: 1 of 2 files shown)


aomarks:
I see, so one of the changes here is that all of the internal classes like `CacheDirective`, `GuardDirective` etc. are now documented here.

I'm not sure if a user would ever need to know about or interact with these classes? If not, then I'm not sure if we should include them here, since they are potentially confusing, and should consider keeping the filter but adding an exception for `StyleInfo` (and any other types that seem like they would be useful to users).

AndrewJakubowicz:
They were always documented, but we removed the id attribute which as far as I can tell would only remove the search index.

This results in broken links. See the return signature on production lit.dev https://lit.dev/docs/api/directives/#:~:text=Signature-,cache(v)%3A,-DirectiveResult%3CCacheDirective%3E

aomarks:
> They were always documented, but we removed the id attribute which as far as I can tell would only remove them from the search index.
> 
> This results in broken links. See the return signature on production lit.dev https://lit.dev/docs/api/directives/#:~:text=Signature-,cache(v)%3A,-DirectiveResult%3CCacheDirective%3E
> 
> Which is trying to link to the class documented directly below it: `<h2 id="">CacheDirective</h2>`

Oh, now I remember. The purpose was to keep the TOC on the right-hand-side simpler. Since users won't really care about all of these classes, it seemed unfortunate to make the TOC less useful:

### Before
image

### After
image

I'm still wondering if we should include most of these types at all?

AndrewJakubowicz:
That is helpful info! I agree about not adding extra noise to that list.

Will need to revisit the implementation. I think the `id` attribute should remain such that explicit links to the fragments resolve. But the contents should remain only the functions.
Thank you!

aomarks:
> That is helpful info! I agree about not adding extra noise to that list.
> 
> Will need to revisit the implementation. I think the `id` attribute should remain such that explicit links to the fragments resolve. But the contents should remain only the functions. Thank you!

I'm not necessarily strongly opposed to this change, just something to consider. I am genuinely unsure whether these types are useful or not. If they are, then they should probably be included. But maybe it's just a subset of them.

PR #752, opened 2022-04-12
Implement redesigned asides

### LINKED ISSUE #751: Implement redesigned asides
we have new aside mocks from ocupop. We need to implement them.

e111077:
fixes #751

The goal in the end is to SSR these without hydration, but due to limitations with declarative shadow dom, we cannot do this for the tutorials section as it will mess up the styling. But as of now all the content is in the light dom so it doesn't necessarily matter right now.

They look like so:

image

### DIFF
--- a/packages/lit-dev-content/.eleventy.js
+++ b/packages/lit-dev-content/.eleventy.js
@@ -77,6 +77,9 @@ module.exports = function (eleventyConfig) {
     '../lit-dev-content/samples/tutorials/**/tutorial.json'
   );
   eleventyConfig.addWatchTarget('../lit-dev-content/samples/tutorials/**/*.md');
+  for (const component of componentsToSSR) {
+    eleventyConfig.addWatchTarget(component);
+  }
 
   // Placeholder shortcode for TODOs
   // Formatting is intentional: outdenting the HTML causes the
@@ -547,6 +550,7 @@ ${content}
 
   eleventyConfig.addPlugin(litPlugin, {
     componentModules,
+    ignoreGlobs: ['**/tutorials/content/**/*']
   });
 
   return {
--- a/packages/lit-dev-content/.eleventy.js
+++ b/packages/lit-dev-content/.eleventy.js
@@ -550,7 +550,7 @@ ${content}
 
   eleventyConfig.addPlugin(litPlugin, {
     componentModules,
-    ignoreGlobs: ['**/tutorials/content/**/*']
+    ignoreGlobs: ['**/tutorials/content/**/*'],
   });
 
   return {
--- a/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
+++ b/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
@@ -164,28 +164,47 @@ customElements.define('my-element', MyElement);
 You can also insert an aside in your instructions by using the following format:
 
 ```html
-<aside class="positive">
-  Make sure to do <code>this</code>!
-</aside>
+<litdev-aside type="positive">
 
-<aside class="negative">
-  Make sure <b>NOT</b> to do <code>this</code>!
-</aside>
+The first line is always a bolded header.
 
-<aside class="info">
-  Check out more info <a href="https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html">in this docs section</a>.
-</aside>
+Make sure to do `this`!
+
+</litdev-aside>
+
+<litdev-aside type="warn" no-header>
+
+The `no-header` will make sure that this line is not bolded.
+
+Beware of `this`!
+
+</litdev-aside>
+
+<litdev-aside type="negative">
+
+Make sure NOT to do `this`!
+
+The following non-header lines here make sense to explain the assertion in
+the header line above.
+
+</litdev-aside>
+
+<litdev-aside type="info">
+
+Check out more info [in this docs section](https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html).
+
+</litdev-aside>
 ```
 
-*Note:* markdown is not rendered inside an aside, you must use HTML.
+*Note:* markdown will only be parsed as markdown if there is an empty line
+between the text and the HTML tag.
 
 The available asides are:
 
 * `positive`
-* `warning`
+* `warn`
 * `negative`
 * `info`
-* `labs`
 
 ## Good tutorial Habits
 
--- a/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
+++ b/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
@@ -191,7 +191,7 @@ the header line above.
 
 <litdev-aside type="info">
 
-Check out more info [in this docs section](https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html).
+Check out more info [in this docs section](/docs/templates/expressions/#well-formed-html).
 
 </litdev-aside>
 ```
--- a/packages/lit-dev-content/site/tutorials/content/advanced-templating/01.md
+++ b/packages/lit-dev-content/site/tutorials/content/advanced-templating/01.md
@@ -33,6 +33,6 @@ the header line above.
 
 <litdev-aside type="info">
 
-Check out more info [in this docs section](https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html).
+Check out more info [in this docs section](/docs/templates/expressions/#well-formed-html).
 
 </litdev-aside>
--- a/packages/lit-dev-content/site/tutorials/content/advanced-templating/01.md
+++ b/packages/lit-dev-content/site/tutorials/content/advanced-templating/01.md
@@ -5,3 +5,34 @@ See also these tutorials maybe?
 * list
 * list
 * list
+
+<litdev-aside type="positive">
+
+The first line is always a bolded header.
+
+Make sure to do `this`!
+
+</litdev-aside>
+
+<litdev-aside type="warn" no-header>
+
+The `no-header` will make sure that this line is not bolded.
+
+Beware of `this`!
+
+</litdev-aside>
+
+<litdev-aside type="negative">
+
+Make sure NOT to do `this`!
+
+The following non-header lines here make sense to explain the assertion in
+the header line above.
+
+</litdev-aside>
+
+<litdev-aside type="info">
+
+Check out more info [in this docs section](https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html).
+
+</litdev-aside>
--- a/packages/lit-dev-content/src/components/litdev-aside.ts
+++ b/packages/lit-dev-content/src/components/litdev-aside.ts
@@ -1,0 +1,79 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {css, html, LitElement} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
+import {greenCheckIcon} from '../icons/green-check-icon.js';
+import {redXIcon} from '../icons/red-x-icon.js';
+import {yellowBangIcon} from '../icons/yellow-bang-icon';
+import {blueInfoIcon} from '../icons/blue-info-icon';
+
+export type AsideVariant = 'positive' | 'negative' | 'warn' | 'info';
+
+@customElement('litdev-aside')
+export class LitDevAside extends LitElement {
+  static styles = css`
+    :host {
+      color: currentColor;
+      display: block;
+      margin: 1em 0;
+    }
+
+    aside {
+      display: flex;
+      border-style: solid;
+      border-width: 1px;
+      border-color: var(--lit-dev-aside-border-color, #ccc);
+      padding: 1em 1em 1em 0em;
+    }
+
+    slot {
+      display: block;
+      flex-grow: 1;
+    }
+
+    svg {
+      width: 1.5em;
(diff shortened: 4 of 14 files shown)


packages/lit-dev-content/.eleventy.js
```diff
@@ -77,6 +77,9 @@ module.exports = function (eleventyConfig) {
     '../lit-dev-content/samples/tutorials/**/tutorial.json'
   );
   eleventyConfig.addWatchTarget('../lit-dev-content/samples/tutorials/**/*.md');
+  for (const component of componentsToSSR) {
```
AndrewJakubowicz, line 80: Tiny nit, does the `ssr.js` file need a license header?

```
/**
 * @license
 * Copyright 2022 Google LLC
 * SPDX-License-Identifier: BSD-3-Clause
 */
```

Just so I understand - this is adding watch for dev-mode for the SSR'd components.

AndrewJakubowicz (approved):
Awesome, awesome! Minor feedback!

AndrewJakubowicz:
Weird rendering bug with the cross icon on Firefox:
Screen Shot 2022-04-12 at 4 50 35 PM

arthurevans:
These look way better than the old ones, I think. 

Are these only for the tutorial section? I looked at the asides in the regular docs and I *think* they're the old ones? Or maybe they look different here because they're on a dark background?

Wanted to make sure that the existing asides that use a run in head rather than a separate heading line look OK. That is:

**Don't touch that dial!** This is a job for Superman. Or at least someone in a swanky cape.

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -164,28 +164,47 @@ customElements.define('my-element', MyElement);
...
+Make sure NOT to do `this`!
+
+The following non-header lines here make sense to explain the assertion in
+the header line above.
+
+</litdev-aside>
+
+<litdev-aside type="info">
+
+Check out more info [in this docs section](https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html).
```
augustjk, line 194: Doesn't seem like we need the mods here?

aomarks:
> The goal in the end is to SSR these without hydration, but due to limitations with declarative shadow dom, we cannot do this for the tutorials section as it will mess up the styling. But as of now all the content is in the light dom so it doesn't necessarily matter right now.

It will cause layout shift on every page that uses an side, though, right? Previously, the styles for asides were defined in the main stylesheet. Now the styles won't load until the component loads.

packages/lit-dev-content/site/tutorials/content/advanced-templating/01.md
```diff
@@ -4,4 +4,35 @@ See also these tutorials maybe?
...
+Make sure NOT to do `this`!
+
+The following non-header lines here make sense to explain the assertion in
+the header line above.
+
+</litdev-aside>
+
+<litdev-aside type="info">
+
+Check out more info [in this docs section](https://lit.dev/docs/templates/expressions/?mods=tutorialCatalog#well-formed-html).
```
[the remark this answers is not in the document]
e111077: oops great catch!

packages/lit-dev-content/src/components/litdev-aside.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  @property({type: String})
+  type: AsideVariant = 'info';
+
+  @property({type: Boolean, reflect: true, attribute: 'no-header'})
+  noHeader = false;
+
+  render() {
+    return html`
+      <aside>
```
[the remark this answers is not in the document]
e111077: it is essentially `role="complementary"` personally I'm not a fan of forcing a role on host but i can be swayed. You think I should omit it?

packages/lit-dev-content/src/components/litdev-aside.ts
```diff
@@ -0,0 +1,79 @@
...
+    switch (this.type) {
+      case 'positive':
+        return greenCheckIcon;
+      case 'negative':
+        return redXIcon;
+      case 'warn':
+        return yellowBangIcon;
+      case 'info':
+        return blueInfoIcon;
+    }
```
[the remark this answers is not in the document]
e111077: I'm torn here because I'm a fan of HTML "failing successfully", but also you don't write these components in a place where `lit-plugin` is available and you have an increased chance to write a wrong `type`. Though, we don't really have a CI check to check for thrown exceptions, so it might just pass this error onto the user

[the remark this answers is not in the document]
augustjk: It seems valid to not want to throw when the type is coming from a user supplied attribute, not something hard coded. Perhaps something like this? (not sure about log level, warn vs error)
```ts
default:
  const exhaustiveCheck: never = this.type;
  logger.warn(`Received unexpected type for <litdev-aside>: ${exhaustiveCheck}`);
```
Then TS will make sure we've handled all expected cases when authoring, without runtime throwing.

e111077:
@arthurevans this is for both. Just putting them into the docs should have the right light theme. Ocupop made sure that the icons and stuff should have the right contrast etc for light and dark mode

e111077:
@aomarks as of now it will, but i haven't changed any of the old ones yet. Looking to find a solution for lit/lit#2727 before migrating the rest of the site. Though for tutorials section, JS should already be loaded so it won't be an issue here.

packages/lit-dev-content/src/icons/blue-info-icon.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: Ocupop Design
+export const blueInfoIcon = html`<svg x="0px" y="0px" viewBox="0 0 125 125">
```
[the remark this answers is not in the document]
e111077: removed. Did not add xmlns due to requests in previous PRs to remove it

packages/lit-dev-content/src/icons/blue-info-icon.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import {html} from 'lit';
+
+// Source: Ocupop Design
+export const blueInfoIcon = html`<svg viewBox="0 0 125 125">
+  <desc>A white letter i in a blue circle</desc>
+  <g>
+    <circle fill="#324fff" cx="62.5" cy="62.5" r="61.5" />
+    <g>
```
aomarks, line 14: The `<g>`s in this svg seem unnecessary

packages/lit-dev-content/src/icons/yellow-bang-icon.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import {html} from 'lit';
+
+// Source: Ocupop Design
+export const yellowBangIcon = html`<svg viewBox="0 0 125 125">
+  <desc>A white exclamation mark in a yellow circle</desc>
+  <g>
```
aomarks, line 12: The `<g>`s in this svg seem unnecessary

packages/lit-dev-content/src/components/litdev-aside.ts
```diff
@@ -0,0 +1,79 @@
...
+import {yellowBangIcon} from '../icons/yellow-bang-icon.js';
+import {blueInfoIcon} from '../icons/blue-info-icon.js';
+
+export type AsideVariant = 'positive' | 'negative' | 'warn' | 'info';
+
+@customElement('litdev-aside')
+export class LitDevAside extends LitElement {
+  static styles = css`
+    :host {
+      color: currentColor;
```
aomarks, line 20: Isn't this a no-op? `color` is already `currentColor`, by definition.

packages/lit-dev-content/src/components/litdev-aside.ts
```diff
@@ -0,0 +1,79 @@
...
+    :host(:not([no-header])) ::slotted(:first-child) {
+      font-weight: bold;
+    }
+
+    ::slotted(*) {
+      margin: 0;
+    }
+  `;
+
+  @property({type: String})
```
aomarks, line 52:
```suggestion
  @property()
```

aomarks:
> @aomarks as of now it will, but i haven't changed any of the old ones yet. Looking to find a solution for [lit/lit#2727](https://github.com/lit/lit/issues/2727) before migrating the rest of the site. Though for tutorials section, JS should already be loaded so it won't be an issue here.

Ah ok

packages/lit-dev-content/src/components/litdev-aside.ts
```diff
@@ -0,0 +1,85 @@
...
+  type: AsideVariant = 'info';
+
+  @property({type: Boolean, reflect: true, attribute: 'no-header'})
```
aomarks, line 55: Is there any reason for this to reflect?

packages/lit-dev-content/src/components/litdev-aside.ts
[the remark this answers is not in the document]
AndrewJakubowicz: I do not wish to sway you, and you have swayed me :D Thanks for the explanation!

aomarks (changes requested):

packages/lit-dev-content/src/icons/blue-info-icon.ts
e111077: aww gee!

packages/lit-dev-content/src/components/litdev-aside.ts
e111077: this lets color seep into the shadow DOM, but you're right, it's all slotted so it is indeed a no-op

packages/lit-dev-content/src/components/litdev-aside.ts
e111077: the attribute is used for styling. Should make all efforts to make sure that it is applied, no? TBH it doesn't really need to be a reactive property, but in some future cases lit-plugin may error / warn.

WDYT?

packages/lit-dev-content/src/components/litdev-aside.ts
aomarks: I think all inheritable css properties like `color`, `font-family`, etc. leak through shadow dom automatically: [playground](https://lit.dev/playground/#project=W3sibmFtZSI6InNpbXBsZS1ncmVldGluZy50cyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcbmltcG9ydCB7Y3VzdG9tRWxlbWVudCwgcHJvcGVydHl9IGZyb20gJ2xpdC9kZWNvcmF0b3JzLmpzJztcblxuQGN1c3RvbUVsZW1lbnQoJ3NpbXBsZS1ncmVldGluZycpXG5leHBvcnQgY2xhc3MgU2ltcGxlR3JlZXRpbmcgZXh0ZW5kcyBMaXRFbGVtZW50IHtcbiAgQHByb3BlcnR5KClcbiAgbmFtZSA9ICdTb21lYm9keSc7XG5cbiAgcmVuZGVyKCkge1xuICAgIHJldHVybiBodG1sYDxwPkhlbGxvLCAke3RoaXMubmFtZX0hPC9wPmA7XG4gIH1cbn1cbiJ9LHsibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPHNjcmlwdCB0eXBlPVwibW9kdWxlXCIgc3JjPVwiLi9zaW1wbGUtZ3JlZXRpbmcuanNcIj48L3NjcmlwdD5cbjwvaGVhZD5cbjxib2R5PlxuICA8c3R5bGU-XG4gICAgYm9keSB7XG4gICAgICBjb2xvcjogcmVkO1xuICAgICAgZm9udC1mYW1pbHk6IG1vbm9zcGFjZTtcbiAgICB9XG48L3N0eWxlPlxuICA8c2ltcGxlLWdyZWV0aW5nIG5hbWU9XCJXb3JsZFwiPjwvc2ltcGxlLWdyZWV0aW5nPlxuPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjIuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMS4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4yLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ)

packages/lit-dev-content/src/components/litdev-aside.ts
aomarks: Oh sorry, I missed that it was used in the css. Makes sense.

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
augustjk: This is a different file than where Andrew pointed out, btw.

aomarks (approved):

augustjk (approved):

AndrewJakubowicz (approved):
AWESOME!

e111077:
teamwork dreamwork, y'all!

e111077:
oh no! some of those svg elements were to make the icons white

PR #755, opened 2022-04-18
Filter inline TOC by attribute instead of by deleting id.

AndrewJakubowicz:
A better implementation of https://github.com/lit/lit.dev/pull/747.

Fixes: https://github.com/lit/lit.dev/issues/746
Also fixes: https://github.com/lit/lit.dev/issues/344

### Context

The `anchorFilter` originally deleted the `id` attribute from some of the generated documentation in order to simplify the table of contents generated on the page. This resulted in some fragment urls not linking correctly, i.e.: https://lit.dev/docs/api/directives/#RefOrCallback which doesn't link correctly because the `id` has been removed.

### How

Fixed by using the [`data-toc-exclude`](https://github.com/JordanShurmer/eleventy-plugin-nesting-toc#:~:text=can%20add%20the-,data%2Dtoc%2Dexclude,-attribute%20to%20exclude) boolean attribute that [eleventy-plugin-nesting-toc](https://github.com/JordanShurmer/eleventy-plugin-nesting-toc) uses to filter the TOC.

### Result

 - Some previously unsearchable symbols are now searchable. For example `StyleInfo`.
 - Fixes `{@link}` jsdocs to non function directive symbols.

### Testing

The filter is only used on the directive page. I manually compared the table of contents between [preview deploy](https://pr755-2297b6b---lit-dev-5ftespv5na-uc.a.run.app/docs/api/directives/) and the [production deploy](https://lit.dev/docs/api/directives/).

And the `RefOrCallback` link example now works: https://pr755-551f95e---lit-dev-5ftespv5na-uc.a.run.app/docs/api/directives/#RefOrCallback

### DIFF
--- a/packages/lit-dev-api/api-data/lit-2/pages.json
+++ b/packages/lit-dev-api/api-data/lit-2/pages.json
@@ -8163,7 +8163,8 @@
                     "name": "AsyncAppendDirective",
                     "location": {
                       "page": "directives",
-                      "anchor": "AsyncAppendDirective"
+                      "anchor": "AsyncAppendDirective",
+                      "excludeFromTOC": true
                     }
                   }
                 }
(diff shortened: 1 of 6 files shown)


aomarks (approved):
Great solution, gives us the best of both worlds! Great find with `data-toc-exclude`.

PR #757, opened 2022-04-20
Use wireit in most of our scripts to track dependencies, add caching, and as a universally useful watch mode

rictic:
This should behave exactly the same, only now the dependencies are explicit (e.g. if you run `npm test` it will build the typescript if necessary), does caching (so rebuilding or rerunning tests is instant if no TS source files have changed), and supports a watch mode (e.g. with `npm test -- watch`)

### DIFF
--- a/packages/lit-dev-tools-cjs/package.json
+++ b/packages/lit-dev-tools-cjs/package.json
@@ -6,9 +6,34 @@
   "author": "Google LLC",
   "license": "BSD-3-Clause",
   "scripts": {
-    "build": "npm run build:ts",
-    "build:ts": "../../node_modules/.bin/tsc",
-    "test": "uvu ./lib \".spec.js$\""
+    "build": "wireit",
+    "build:ts": "wireit",
+    "test": "wireit"
+  },
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:ts"
+      ]
+    },
+    "build:ts": {
+      "files": [
+        "src/**",
+        "package.json"
+      ],
+      "output": [
+        "lib"
+      ],
+      "command": "../../node_modules/.bin/tsc"
+    },
+    "test": {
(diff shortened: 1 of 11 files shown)


packages/lit-dev-tools-cjs/package.json
```diff
@@ -6,9 +6,34 @@
...
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:ts"
+      ]
+    },
+    "build:ts": {
+      "files": [
+        "src/**",
+        "package.json"
```
aomarks, line 22: We do have a tsbuildinfo file declared, and we set incremental/composite here: https://github.com/lit/lit.dev/blob/c7ca68dcf85519b597efbc487140718e8ea72b65/packages/lit-dev-tools-cjs/tsconfig.json#L7

So this might be doing incremental build.

In which case, you need to include the tsbuildinfo file, and set `"clean": "if-file-deleted"`

rictic: +1, done

The tsbuildinfo file is included, as it's emitted into the lib directory

rictic:
See https://github.com/lit/lit.dev/pull/759 to fix the external links

packages/lit-dev-tools-cjs/package.json
```diff
@@ -6,9 +6,36 @@
...
+  "wireit": {
+    "build": {
+      "dependencies": [
+        "build:ts"
+      ]
+    },
+    "build:ts": {
+      "files": [
+        "src/**",
+        "package.json",
```
aomarks, line 22: This should be `tsconfig.json` I think?

rictic: Erp, yes! Will fix here and elsewhere

aomarks (approved):

PR #759, opened 2022-04-21
Ignore twitter links when checking external links

rictic:
It seems to be serving 404s in CI for some (but not all) ordinary working URLs. e.g. see https://github.com/lit/lit.dev/runs/6104284767?check_suite_focus=true


aomarks (approved):

aomarks:
> It seems to be serving 404s in CI for some (but not all) ordinary working URLs. e.g. see https://github.com/lit/lit.dev/runs/6104284767?check_suite_focus=true

Already approved but, it does seem unfortunate to lose all 404 coverage for twitter.com. Maybe we are getting rate limited or something?

rictic:
Our rate seems very low, we only make like six checks in each run, and there's a long time between the first and second check, but the first check tends to succeed and the second check tends to fail? I don't know how to explain this

aomarks:
Looks like Elliott might have found another solution: https://github.com/lit/lit.dev/pull/761

rictic:
AFAICT that didn't fix it

Issue #723, opened 2022-03-15
Implement new tutorial catalog design


e111077:
waiting on final stuff from ocupop

PR #760, opened 2022-04-21
[tutorials] Add featured section to the catalog

e111077:
Related to #723

Featured section should always have a max 3 items and `intro-to-lit` should likely always be the first item.

Question to reviewers: When the screen resizes to be smaller, should I be hiding the "overflow" featured tutorials?


packages/lit-dev-content/site/css/tutorial-catalog.css
```diff
@@ -38,6 +40,15 @@ main {
   grid-template-columns: repeat(3, 1fr);
 }
 
+section.featured,
```
AndrewJakubowicz, line 43: Is `section` required?

packages/lit-dev-content/site/tutorials/index.html
```diff
@@ -11,22 +11,18 @@
...
+  <div class="category-subtitle"></div>
   <div class="center">
     <div class="card-grid">
-      {% for tutorial in tutorials %}
-        {% if tutorial.category == 'Learn' %}
+      {% for tutorial in tutorials | sort(false, false, 'featured.position') %}
+        {% if tutorial.featured.position !== undefined and tutorial.featured.position < 3 %}
           <litdev-tutorial-card
-              size="{{ tutorial.size }}"
+              size="small"
```
AndrewJakubowicz, line 22: What happens if there is too much content - I.e. the featured tutorial was originally on a larger card?

packages/lit-dev-content/site/tutorials/tutorials.11tydata.js
```diff
@@ -15,14 +15,14 @@ const path = require('path');
  * @returns {Promise<Object>} Tutorial data to be consumed by
  *     /tutorials/index.html
  */
-const loadTutorialData = async (dirname) => {
+const loadTutorialData = async (dirname, options = {featured: {position: undefined}}) => {
```
AndrewJakubowicz, line 18: Is it worth breaking out the `options` type into a separate interface, and documenting that `featured` is an optional `number` between 0 and 2, inclusive?

packages/lit-dev-content/site/css/tutorial-catalog.css
```diff
@@ -1,6 +1,8 @@
-main {
+main,
+.not-featured {
   display: flex;
   flex-direction: column;
+  z-index: 0;
```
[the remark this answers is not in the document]
e111077: it is. tbh weird behavior

packages/lit-dev-content/site/tutorials/index.html
e111077: I was thinking about this. The real thing would be to add "featured" description, but that's way too much work for the author. In the meantime, if it's too long, then it won't be featured. This is an issue we can tackle down the line

AndrewJakubowicz (approved):
I like the current behavior of the featured section but easy either way.

Awesome!

PR #761, opened 2022-04-21
Fix twitter links on broken link checker

e111077:


rictic:
This doesn't look like it fixed it to me: https://github.com/lit/lit.dev/runs/6104952768?check_suite_focus=true

PR #763, opened 2022-04-21
Update docker node version, deps, and lit-dev-tools-cjs deps

rictic:
As a bonus, this was the only way I could get the dockerfile to work at all for me locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Official lightweight Node.js image
 # https://hub.docker.com/_/node
-FROM node:17-slim
+FROM node:18-slim
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
   # Dependencies of Playwright Chromium for Playground SSR
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,14 @@
 # Official lightweight Node.js image
 # https://hub.docker.com/_/node
-FROM node:15-slim
+FROM node:17-slim
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
   # Dependencies of Playwright Chromium for Playground SSR
   libgtk-3-0 \
   libnss3 \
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -1,12 +1,14 @@
 # Official lightweight Node.js image
 # https://hub.docker.com/_/node
-FROM node:15-slim
+FROM node:17-slim
```
aomarks, line 3: I think we should use Node 16, since that's Active LTS. Node 17 end of life is in only about a month (https://nodejs.org/en/about/releases/).

aomarks (changes requested):

Dockerfile
rictic: I get an odd puppeteer error with Node 16, but Node 18 works, it'll be in LTS soon, and it's supported through 2025

aomarks: Feels like we should run our tests with the same node version that we run in production. I guess we could bump our github actions here to 18 as well.

packages/lit-dev-tests/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "build": "tsc",
     "start": "cd ../.. && npm run start",
-    "test:integration": "playwright test --config=lib/playwright.config.js",
+    "test:integration": "playwright install && playwright test --config=lib/playwright.config.js",
```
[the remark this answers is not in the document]
rictic: Some of the github actions manually run `npx playwright install-deps` which is probably helping, but this is fast to run and will also help users outside of CI

Dockerfile
aomarks: Thoughts?

rictic: +1, bumped all github actions to node 18, let's see how it goes

aomarks (approved):

PR #764, opened 2022-04-21
Add wireit to a couple more packages

rictic:
Still no changes to semantics, just caching and watch and all that good stuff

### DIFF
--- a/packages/lit-dev-server/package.json
+++ b/packages/lit-dev-server/package.json
@@ -24,7 +24,7 @@
       ],
       "files": [
         "src/**",
-        "package.json",
+        "tsconfig.json",
         "../../tsconfig.base.json"
       ],
       "output": [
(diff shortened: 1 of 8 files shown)


packages/lit-dev-server/package.json
```diff
@@ -8,9 +8,38 @@
...
+        "build:ts"
+      ]
+    },
+    "build:ts": {
+      "dependencies": [
+        "../lit-dev-tools-cjs:build"
+      ],
+      "files": [
+        "src/**",
+        "package.json",
```
aomarks, line 27: Should this be `tsconfig.json` instead of `package.json`? Same below.

aomarks (changes requested):

packages/lit-dev-server/package.json
rictic: +1, done

aomarks (approved):

PR #770, opened 2022-04-22
Increase renderer timeout

rictic:
Been seeing this error in some runs that have stalled out in the wireit branches, I think because we're doing more things in parallel it's meant that some eleventy rendering has been timing out


aomarks (approved):

aomarks:
> Been seeing this error in some runs that have stalled out in the wireit branches, I think because we're doing more things in parallel it's meant that some eleventy rendering has been timing out

Locally or on CI?

rictic:
on CI

PR #771, opened 2022-04-22
Hook up wireit into the tests package

rictic:


packages/lit-dev-tests/package.json
```diff
@@ -32,6 +32,33 @@
...
+      ]
+    },
+    "test:integration": {
+      "dependencies": [
+        "build"
+      ],
+      "files": [
+        "src/**/*.png"
+      ],
+      "command": "playwright install && playwright test --config=lib/playwright.config.js"
```
aomarks, line 48: Add `"output": []` so that it can be cached if nothing changed?

packages/lit-dev-tests/package.json
```diff
@@ -32,6 +32,33 @@
...
+    },
+    "test:integration:update-golden-screenshots": {
+      "dependencies": [
+        "build"
+      ],
+      "files": [],
```
aomarks, line 54: Add `"output": []` so that it can be cached if nothing changed?

packages/lit-dev-tests/package.json
```diff
@@ -32,6 +32,33 @@
...
+      "command": "playwright install && playwright test --config=lib/playwright.config.js --update-snapshots"
+    },
+    "test:links:redirects": {
+      "dependencies": [
+        "build"
+      ],
+      "command": "node lib/check-redirects.js"
```
aomarks, line 61: Add `"files": [], "output": []` so that it can be cached if nothing changed?

aomarks (changes requested):
nice!

packages/lit-dev-tests/package.json
rictic: This updates the screenshots, so it needs that as output, done

packages/lit-dev-tests/package.json
rictic: Makes sense

aomarks (approved):

PR #772, opened 2022-04-22
Test links as part of integration test

rictic:
Also roll up testing internal and external links into a single command now that it's consistent. I've manually verified that it fails if there's a missing internal or a missing external link.

I put it with the integration tests rather than the unit tests because building the site is the majority of the time for both the integration tests and the links test, while the unit tests now run completely in like 20s on CI.


aomarks (approved):
Cool

PR #785, opened 2022-04-28
[tutorials] optimize images

e111077:
convert them to webp and in the litdev card, load lazily


AndrewJakubowicz (approved):
Nice!

aomarks:
Does safari support webp?

This indicates maybe only in TP? https://caniuse.com/webp

e111077:
It does support webp. It just indicates that you need MacOS 11+

PR #791, opened 2022-04-29
Reactivity tutorial

arthurevans:
Not sure if this version will build on github. Failing locally but it seems to be related to cached files.


packages/lit-dev-content/site/tutorials/content/reactivity/04.md
```diff
@@ -0,0 +1,9 @@
+One way your component can react to property changes is to compute values based on the new and existing property values.
+
+If you need to compute a value to render, you can compute it in the `render()` method itself. However a more general solution is to compute values in the `willUpdate()` callback. This has a couple of advantages:
```
sorvell, line 3: Let's lead with the right way: do this in `willUpdate`. Computing state (properties are are saved on the element) in render is not something we should generally recommend since we want to encourage `render` to just be for rendering and not side-effectful.

Computing ephemeral values needed just for rendering in `render` is a-ok, but stateful computed values should be done in willUpdate.

You typically want `changedProperties` to know if you want to compute and be able to see old values. You get this info in willUpdate, but not in render.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
```diff
@@ -0,0 +1,9 @@
...
+
+-   You can make computed properties available to [reactive controllers](/docs/composition/controllers/). The `willUpdate()` callback is called before the `hostUpdate()` callback is called on any [reactive controllers](/docs/composition/controllers/).
```
sorvell, line 5: This is super advanced and could easily be removed.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
```diff
@@ -0,0 +1,9 @@
...
+
+-   If you subclass the component, you can supply a new `render()` method and still use  the computed properties.
+
+Updating a reactive property during `willUpdate()` does not trigger a new update cycle.
```
sorvell, line 9: It does not in render either. Let's either say this more fully or omit.

During the update cycle, setting any property will not trigger an additional update until the last callback in the cycle: `updated`.

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
```diff
@@ -0,0 +1,45 @@
+For some reactive properties, you may want to customize _change detection_. By default, Lit uses a strict equality comparison to determine whether the new value is different from the old value (`newValue !== oldValue`).
```
sorvell, line 1: This is pretty advanced for a beginner tutorial, and I think it can be left out. If you do include it, I think it would be important to explain when to use shouldUpdate v. hasChanged: 
* shouldUpdate: allows you to opt-out of an update based on the full set of pending changes
* hasChanged: allows you to opt-in/out of an update based on a single property being set.

packages/lit-dev-content/site/tutorials/content/reactivity/00.md
```diff
@@ -0,0 +1,14 @@
+Reactive properties and the reactive update cycle are an important part of the Lit library. Updating a reactive property can trigger a reactive update cycle, which in turn produces side effects—like re-rendering the element, updating its attributes, or firing events.
+
+If you did the [Intro to Lit](/tutorials/intro-to-lit/) tutorial, you've already used some reactive properties. This tutorial digs a little deeper into reactive properties and the reactive update cycle:
+
+*   How reactive properties trigger an update (and how to triggering an update manually).
+*   Customizing when the component updates.
```
sorvell, line 6: I think performing an action, a custom side-effect, when a property changes is more important to explain than `shouldUpdate` or `hasChanged`.

This would be implemented via `updated`, and could be just as basic as alert|console.log(`${this.name}!`)

sorvell (changes requested):

packages/lit-dev-content/site/tutorials/content/reactivity/00.md
```diff
@@ -0,0 +1,14 @@
+Reactive properties and the reactive update cycle are an important part of the Lit library. Updating a reactive property can trigger a reactive update cycle, which in turn produces side effects—like re-rendering the element, updating its attributes, or firing events.
```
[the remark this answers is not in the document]
arthurevans: Hmm... Right, I was thinking like what we describe here: https://lit.dev/docs/components/lifecycle/#updatecomplete ... But it's confusing in this context, will remove.

packages/lit-dev-content/site/tutorials/content/reactivity/00.md
arthurevans: I'd be happy to add an `updated()` example. It felt kind of strange not having something there. But I had a hard time finding reasonable use cases.

A console log or alert would demonstrate that `updated()` exists, but not why you'd use it.

The only example we've mentioned so far is measuring the element. (You also mention that changing a property in `updated` triggers a new update, so presumably there's a place that's useful, like triggering a transition animation?)

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
arthurevans: Yeah, I basically included this to explain the advantages of `willUpdate` over `render` ... I was going to mention that you got `changedProperties` as well, but I think that you have a change that exposes `changedProperties` on the component, which weakens that argument...

Anyway--if we just present `willUpdate` as the default way to do this, we don't need to explain the advantages.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,22 @@
+How does a reactive property work? When you add a reactive property, Lit adds accessors for that property. When you set the property, Lit checks to see if the property has changed. If the property has changed, Lit starts an update cycle.
+
+By default, Lit uses a strict equality check to determine whether a property has changed. If you mutate an existing array, Lit won't detect a change. Try clicking the "x" button next to one of the list items in the preview pane. You'll notice the list doesn't update.
+
+One way around this is to treat the array as immutable. Always create a new array for every change. Try replacing the existing `splice()` call with the following line:
```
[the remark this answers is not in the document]
arthurevans: Do you have a suggested destination for immutable data patterns? I'm not sure what would be a good, technology-neutral source to point people to...

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
```diff
@@ -0,0 +1,30 @@
+Setting a reactive property or calling `requestUpdate()` triggers a reactive update cycle. The update cycle is _asynchronous_, so that updating several properties in the course of a single event handler, for example, only triggers a single update.
+
+At the beginning of an update cycle, Lit calls the `shouldUpdate()` method, passing in a map of changed properties.
+
+You can override `shouldUpdate()` to determine whether the component needs to update itself.
```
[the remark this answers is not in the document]
e111077: I was thinking the opposite – why is duration a reactive prop anyway rather than a public class member unless you're using the attribute to pre-select the default option.

+1 to steve's suggestion

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
```diff
@@ -0,0 +1,37 @@
...
+      "hasAfter": true
+    },
+    {
+      "title": "Should the component update?",
+      "hasAfter": true
+    },
+    {
+      "title": "Computing values during an update",
+      "hasAfter": true
+    },
```
e111077, line 27:
```suggestion
    {
      "title": "Computing values during an update",
      "noSolve": true
    },
```

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
e111077, line 5: I think this list could include something along the line that it's a good spot to reconcile between reactive properties and won't trigger rerender unlike in the render function (right?)

packages/lit-dev-content/samples/tutorials/reactivity/05/before/my-element.js
```diff
@@ -0,0 +1,37 @@
+import {LitElement, html, css, PropertyValues} from 'lit';
```
e111077, line 1:
```suggestion
import {LitElement, html, css} from 'lit';
```

packages/lit-dev-content/samples/tutorials/reactivity/05/before/my-element.ts
```diff
@@ -0,0 +1,33 @@
+import {LitElement, html, css, PropertyValues} from 'lit';
```
e111077, line 1:
```suggestion
import {LitElement, html, css} from 'lit';
```

packages/lit-dev-content/samples/tutorials/reactivity/05/before/my-element.js
```diff
@@ -0,0 +1,37 @@
...
+
+class MyElement extends LitElement {
+  // TODO: add hasChanged function on the date property
+  static properties = {
+    date: {}
+  }
+
+  constructor() {
+    this.date = new Date();
+  }
```
e111077, line 14:
```suggestion
  constructor() {
    super();
    this.date = new Date();
  }
```

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
e111077: I'm not sure if there's enough content to warrant a `Reactivity (intermediate)`

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
```diff
@@ -0,0 +1,45 @@
...
+}
+```
+
+```js
+function dateHasChanged(newDate, oldDate) {
+  const hasChanged = (oldDate == undefined) ||
+    !(newDate.getFullYear() == oldDate.getFullYear() &&
+    newDate.getMonth() == oldDate.getMonth() &&
+    newDate.getDate() == oldDate.getDate());
+  return hasChanged;
```
e111077, line 27:
```suggestion
  return hasChanged;
}
```

packages/lit-dev-content/samples/tutorials/reactivity/05/after/index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script type="module" src="my-element.ts"></script>
```
e111077, line 4:
```suggestion
    <script type="module" src="my-element.js"></script>
```

packages/lit-dev-content/samples/tutorials/reactivity/05/after/blinky-box.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function formatDate(date: Date|undefined): string {
+  console.log(date);
+  let formatted = '';
+  if (date && date instanceof Date) {
+    formatted = `${date.getUTCMonth()+1}/${date.getUTCDate()}/${date.getUTCFullYear()}`;
+  }
+  console.log(formatted);
+  return formatted;
+}
```
e111077, line 12:
```suggestion
function formatDate(date: Date|undefined): string {
  let formatted = '';
  if (date && date instanceof Date) {
    formatted = `${date.getUTCMonth()+1}/${date.getUTCDate()}/${date.getUTCFullYear()}`;
  }
  return formatted;
}
```

packages/lit-dev-content/samples/tutorials/reactivity/05/after/blinky-box.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+      `;
+
+  willUpdate(changedProperties: PropertyValues<this>) {
+    if (changedProperties.has('date')) {
+      this._updated = true;
+    }
+    if (changedProperties.has('_updated')) {
+      console.log("Just checking");
+    }
```
e111077, line 48:
```suggestion
```

packages/lit-dev-content/site/tutorials/content/reactivity/06.md
```diff
@@ -0,0 +1,26 @@
+Congratulations! That's the end of this tutorial. You learned:
+
+*   How reactive properties trigger the reactive update cycle.
+*   How to trigger an update manually.
+*   How to control when the component updates.
+*   How to compute values during the update.
+*   How to customizing property change detection.
```
e111077, line 7:
```suggestion
*   How to customize property change detection.
```

e111077 (changes requested):

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
```diff
@@ -0,0 +1,37 @@
+{
+  "header": "Reactivity",
+  "difficulty": "Beginner",
+  "size": "medium",
```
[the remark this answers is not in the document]
arthurevans: Yep, thanks! Already made this change locally.

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
```diff
@@ -0,0 +1,37 @@
...
+  "duration": 70,
+  "category": "Learn",
+  "steps": [
+    {
+      "title": "Reactivity",
+      "hasAfter": true
```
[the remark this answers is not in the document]
arthurevans: Done, thanks! 

... Any way to add a confirmation before deleting? I accidentally removed `01/after` first ...

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
```diff
@@ -0,0 +1,37 @@
+{
+  "header": "Reactivity",
+  "difficulty": "Beginner",
+  "size": "medium",
+  "duration": 70,
```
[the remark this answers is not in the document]
arthurevans: Don't even know where that number came from.  I think the durations are mostly fictional, but I think 20 is more in the ballpark.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,22 @@
+How does a reactive property work? When you add a reactive property, Lit adds accessors for that property. When you set the property, Lit checks to see if the property has changed. If the property has changed, Lit starts an update cycle.
```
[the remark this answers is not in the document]
arthurevans: Done.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,22 @@
...
+
+By default, Lit uses a strict equality check to determine whether a property has changed. If you mutate an existing array, Lit won't detect a change. Try clicking the "x" button next to one of the list items in the preview pane. You'll notice the list doesn't update.
```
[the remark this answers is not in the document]
arthurevans: Done.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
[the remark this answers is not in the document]
arthurevans: Implemented suggestions, minus the link for immutable data patterns.

(But did add a link to Immer.)

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
arthurevans: Done.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
arthurevans: Clarified (I hope).

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
arthurevans: For now I removed the `hasChanged()` step and added an `updated()` step instead. Might add an aside to the `shouldUpdated()` step in the future to point to the `hasChanged()` section in the docs... Or maybe not.

packages/lit-dev-content/site/tutorials/content/reactivity/00.md
arthurevans: OK, added an `updated()` example that does a simple animation in a not-very simple way.

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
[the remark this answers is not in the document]
arthurevans: Done.

packages/lit-dev-content/samples/tutorials/reactivity/01/before/my-element.js
```diff
@@ -0,0 +1,22 @@
+import {LitElement, html} from 'lit';
+
+class MyElement extends LitElement {
+  result = '';
```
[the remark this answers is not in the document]
arthurevans: Done.

packages/lit-dev-content/site/tutorials/content/reactivity/01.md
```diff
@@ -0,0 +1,23 @@
+A reactive property is a property that triggers the component to update whenever the property's value changes.
+
+The sample on the right includes a property, `result`, but it isn't reactive. Pressing the button changes the property value, but the component never changes its display.
+
+To make it a reactive property, add the following code to the beginning of the class.
+
+{% switchable-sample %}
+
+```ts
+  @property()
```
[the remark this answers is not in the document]
arthurevans: Done.

arthurevans:
Updated, addressed existing feedback and added a new step (After the update) and removed the `hasChanged()` step.

Still need to:

- Add replace sample code on intro & recap pages with something more interesting.

arthurevans:
Also, make sure the correct file is selected for all steps.

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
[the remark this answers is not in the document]
e111077: oh nooooo!!

e111077:
Not sure if it's GH's bad UI that hides comments, but there are several un-addressed comments in the middle

arthurevans:
Grr... yeah, apparently I pushed changes that hid the previous conversations. Some of which I never saw in the first place. Not sure how that worked.

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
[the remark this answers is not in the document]
arthurevans: @sorvell The animation happens when the user clicks a button. Triggering an update when the animation speed changes doesn't show anything, because there's no animation running. I did think about something like a key or unique ID--but a) this didn't really make sense to me, unless "Bob"s record never changes, and b) then I'd need to .... I don't know, pick from a list of users with associated data objects? Not sure how to make this into a demo.

I also considered something like an object with a "timestamp" field, where you only rerender if the timestamp is newer than the previous one.  But again, it's a fairly complicated demo setup.

@e111077 Yeah, I had the same though, and my only justification is that you could set it from an attribute. It is not much of a justification, but there you have it.

packages/lit-dev-content/samples/tutorials/reactivity/05/after/index.html
arthurevans: This has been fixed.

packages/lit-dev-content/samples/tutorials/reactivity/05/before/my-element.js
arthurevans: This sample is no longer in the tut.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
arthurevans: Rewrote this section per Steve's input. Apparently only changes in `updated()` trigger rerender.

packages/lit-dev-content/site/tutorials/content/reactivity/01.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+```js
+  static properties = {
+    result: {},
+  };
```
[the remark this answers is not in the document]
arthurevans: Done.

arthurevans:
OK, think I addressed the previously unaddressed comments, made sure the right file was selected, and made sure each component file exports the component class, for consistency.

Steve asked for some sample changes to the `shouldUpdate()` and maybe the `requestUpdate()` steps. I'm not sure whether those changes are doable in our timeframe, but I definitely don't have the brainpower to handle them today.

arthurevans:
Ping on this one @sorvell @e111077 ... I'm still hoping to get this one in...

e111077:
reviewing now

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+```js
+  updated(changedProperties) {
+    if (changedProperties.has('_showMessage')) {
+      const final = this._message.getBoundingClientRect().width;
+      const starting = 0 - final;
+      var player = this._message.animate([
+        { left: `${starting}px` },
+        { left: `0` }
```
e111077, line 32: Consider using transform: translateX since it causes less jank in the demo

```suggestion
        { transform: `translateX(${starting}px)` },
        { transform: `translateX(0)` }
```

packages/lit-dev-content/samples/tutorials/reactivity/05/after/my-element.ts
```diff
@@ -0,0 +1,45 @@
...
+    `;
+  }
+
+  protected updated(changedProperties: PropertyValues<this>): void {
+    if (changedProperties.has('_showMessage')) {
+      const final = this._message.getBoundingClientRect().width;
+      const starting = 0 - final;
+      var player = this._message.animate([
+        { left: `${starting}px` },
+        { left: `0` }
```
e111077, line 38:
```suggestion
        { transform: `translateX(${starting}px)` },
        { transform: `translateX(0)` }
```

e111077 (changes requested):

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
```diff
@@ -0,0 +1,37 @@
...
+
+{% switchable-sample %}
+
+```ts
+  protected willUpdate(changedProperties: PropertyValues<this>): void {
+    // Only need to check changed properties for an expensive computation.
+    if (changedProperties.has('name')) {
+      computeSHA(this.name).then((sha) => this.sha = sha);
+    }
+  }
```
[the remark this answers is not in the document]
arthurevans: Yeah... I like the idea of a synchronous action much better. I was having trouble justifying, say, rot13ing the name, because why not just do that in `render()`? But I think it would make a simpler demo without the weird note about async actions...

The example in the docs (which I think is from @sorvell but I could be wrong) uses a `computeSHA` function so that's what I tried to kludge together. I dunno if @sorvell has thoughts on this.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,36 @@
...
+    this.groceries = [...this.groceries.slice(0, i), ...this.groceries.slice(i+1)];
+```
+
+Since `groceries` is now a new array, the change triggers an update.
+
+<litdev-aside type="info">
+<span>Immutable data can be tricky.</span> The immutable approach looks simple with an array of strings, but correctly cloning a deeply-nested object can be complicated. If you're already using an immutable data pattern, it should just work with Lit. If you're not familiar with immutable data, you may be better off starting with a library like <a href="https://immerjs.github.io/immer/">Immer</a> rather than writing something from scratch.
+</litdev-aside>
+
+Sometimes you may need to mutate data and trigger an update as a result. That's what `requestUpdate()` is for.
```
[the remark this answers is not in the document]
arthurevans: OK, I didn't know that we preferred `requestUpdate()`. Will make the requested update to this section.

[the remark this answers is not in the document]
sorvell: I think it's fine not to mention the perf trade offs here. It's generally ok unless you have a ton of data mutating really frequently... we should handle that separately in the docs (later) and maybe just ensure we're linking out to the right spot.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,40 @@
...
+By default, Lit uses a strict inequality check (`!==`) to determine whether a property has changed. If you mutate an existing object or array, Lit won't detect a change. Try clicking the "x" button next to one of the list items in the preview pane. You'll notice the list doesn't update.
+
+Lit's default equality check works great when using immutable data patterns. For example:
+
+```js
+this.array = [...this.array, newThing]
+this.obj = {...this.obj}
+```
+
+To use this approach in the current component, Replace the existing `splice()` call with the following line:
```
e111077, line 13:
```suggestion
To use this approach in the current component, replace the existing `splice()` call with the following line:
```

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
```diff
@@ -0,0 +1,37 @@
+One way your component can react to property changes is to compute values based on the new and existing property values. 
+
+The `willUpdate()` callback is the ideal place to do these calculations in Lit. It's called near the beginning up the update cycle, before the component renders, so any properties you calculate during `willUpdate()` are available in your `render()` method.
+
+The `willUpdate()` method receives a map of changed properties as an argument, where the keys are property names and the values are the old values for each changed property. In this case, we'll compute a hash value based on the `name` property.
```
e111077, line 5: 2 paragraphs back you call it a callback. Or maybe you want them both to be method idk

```suggestion
The `willUpdate()` callback receives a map of changed properties as an argument, where the keys are property names and the values are the old values for each changed property. In this case, we'll compute a hash value based on the `name` property.
```

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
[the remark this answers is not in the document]
sorvell: Seems fine as is. We could also mention that in `shouldUpdate` you get the complete list of changes and can decide to filter them as appropriate. In this case, for example, you could do `return !(props.size === 1 && props.has('duration'))`... feels more natural for this case.

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
```diff
@@ -0,0 +1,53 @@
...
+      ], {
+        duration: 500,
+        easing: 'ease-out',
+      });
+    }
+  }
```
e111077, line 22:
```suggestion
  protected updated(changedProperties: PropertyValues<this>): void {
    if (changedProperties.has('_showMessage')) {
      const final = this._message.getBoundingClientRect().width;
      const starting = 0 - final;
      var player = this._message.animate([
        { transform: `translateX(${starting}px)` },
        { transform: `translateX(0)` }
      ], {
        duration: 500,
        easing: 'ease-out',
      });
    }
  }
```

e111077 (changes requested):
small changes. Looks good except my prior reservations on the SHA one

packages/lit-dev-content/site/tutorials/content/reactivity/05.md
```diff
@@ -0,0 +1,53 @@
+Sometimes you need to do something after the component has updated. For example, you may need to measure the component after the update, in order to perform an animation. There's a callback for that, and unsurprisingly it's called `updated()`.
```
sorvell, line 1:
```suggestion
Sometimes you need to do something after the component has updated. For example, you may need to measure the component after the update, in order to perform an imperative action like an animation. There's a callback for that, and unsurprisingly it's called `updated()`.
```

sorvell (changes requested):
Getting very close. I think if we just refine step 5, we're g2g.

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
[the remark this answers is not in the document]
arthurevans: Makes sense. Oh, should probably update this to `changedProperties` which is what I used in another step & is more self-documenting.

packages/lit-dev-content/site/tutorials/content/reactivity/00.md
```diff
@@ -0,0 +1,14 @@
+Reactive properties and the reactive update cycle are an important part of the Lit library. Updating a reactive property can trigger a reactive update cycle, which in turn produces side effects—like re-rendering the element and updating its attributes.
+
+If you did the [Intro to Lit](/tutorials/intro-to-lit/) tutorial, you've already used some reactive properties. This tutorial digs a little deeper into reactive properties and the reactive update cycle:
```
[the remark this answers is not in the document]
arthurevans: Added this as the last paragraph of the page, leading into the next step.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,40 @@
...
+
+```ts
+    this.groceries = [...this.groceries.slice(0, i), ...this.groceries.slice(i+1)];
+```
+
+Since `groceries` is now a new array, the change triggers an update.
```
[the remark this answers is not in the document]
arthurevans: Done.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,40 @@
...
+
+{% aside "info" %}
+
+Immutable data can be tricky.
+
+The immutable approach looks simple with an array of strings, but correctly cloning a deeply-nested object can be complicated. If you're already using an immutable data pattern, it should just work with Lit. If you're not familiar with immutable data, you may be better off starting with a library like [Immer](https://immerjs.github.io/immer/) rather than writing something from scratch.
```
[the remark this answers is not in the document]
arthurevans: Sure. Done, link removed.

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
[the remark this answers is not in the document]
arthurevans: Fixed.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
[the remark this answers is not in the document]
arthurevans: Makes sense. I think I'll use Elliott's forward/backward text example... It's a little funky, but we've done first name/last name/name tag so many times in the docs that I think it's nice to take a break.

(If syncing two ways is too complicated, I could always just make it rot-13.)

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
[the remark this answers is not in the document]
arthurevans: OK, just left this as is for now. May add a link after I merge the new section on mutating object- and array- valued properties.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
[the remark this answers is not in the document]
arthurevans: Went with @e111077 's demo, it's kind of random and fun.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
```diff
@@ -0,0 +1,37 @@
...
+    // Only need to check changed properties for an expensive computation.
+    if (changedProperties.has('name')) {
+      computeSHA(this.name).then((sha) => this.sha = sha);
+    }
+  }
+```
+
+{% endswitchable-sample %}
+
+Updating a reactive property during an update cycle callback like `willUpdate()` or `render()` does not trigger a new update cycle. However, if you initiate an asynchronous action from a lifecycle callback, it can trigger a new update. For example, the `computeSHA()` function here returns a promise. Setting `this.sha` in the promise callback *does* trigger a new update.
```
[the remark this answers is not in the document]
arthurevans: Done.

packages/lit-dev-content/site/tutorials/content/reactivity/04.md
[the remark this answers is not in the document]
arthurevans: Added this, with slight tweaks.

arthurevans:
Updated step 5, added new end samples, and addressed other feedback.

e111077 (approved):
I just have to say 👐  that these words 👌  are the best words ☝️👌 the most ☝️ wonderful words ever written 👐 nobody can write better words 👋

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
```diff
@@ -0,0 +1,41 @@
...
+
+{% aside "info" %}
+
+Immutable data can be tricky.
+
+The immutable approach looks simple with an array of strings, but correctly cloning a deeply-nested object can be complicated. If you're already using an immutable data pattern, it should just work with Lit. If you're not familiar with immutable data, you may want to starting with a library that handles immutable data.
+
+{% endaside %}
+
+Sometimes you may need to mutate data and trigger an update as a result. That's what `requestUpdate()` is for.
```
sorvell, line 30: We mention 2 ways to do this but we don't suggest when you'd want to use one way v. the other. Is that an issue? It's sort of out of scope, but are there any docs we can link to about this? I think it's fine as is, but just wanted to mention it in case there's a quick fix.

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
```diff
@@ -0,0 +1,27 @@
+Setting a reactive property or calling `requestUpdate()` triggers a reactive update cycle. At the beginning of an update cycle, Lit calls the `shouldUpdate()` method, passing in a map of changed properties.
+
+You can override `shouldUpdate()` to determine whether the component needs to update itself.
```
sorvell, line 3: Let's explain that by default this returns true.

sorvell (approved):
LGTM!! Just a couple of optional minor suggestions.

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
```diff
@@ -0,0 +1,37 @@
+{
+  "header": "Reactivity",
+  "difficulty": "Beginner",
```
sorvell, line 3: Can we make this intermediate? The concepts are given a light treatment here, but it's definitely not beginner content imo.

packages/lit-dev-content/site/tutorials/content/reactivity/02.md
arthurevans: Removed mention of `requestUpdate()`, linked to new section.

packages/lit-dev-content/site/tutorials/content/reactivity/03.md
arthurevans: Done.

packages/lit-dev-content/samples/tutorials/reactivity/tutorial.json
arthurevans: Done.

PR #793, opened 2022-04-30
implement tutorial gtag

e111077:
Fires tutorial start and end events. Also sets `document.title` which is also useful for gtag events

```
tutorial_start: {
  category: tutorials,
  label: tutorial folder name,
  value: step tutorial page was loaded,
}

tutorial_end {
  category: tutorials,
  label: tutorial folder name,
}
```

## tutorial_start event

fired when the page and manifest area loaded, and includes the step the page was loaded on.

## tutorial_end event

fired when the next button is pressed and the page index matches the length of steps in the laoded manifest -1. Signaling progression

## faults

- start is fired whenever the page loads so that may include reloads
  - can likely be deduplicated by session ID cloud-side
  - hopefully can fire automated event if session has start index of 0 and also fires end event
- if initial page loaded is the final page, then next button cannot be pressed unless they go back and press right
  - doesn't signal progression so this is fine, but may have a mismatched start to end

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-tutorial.ts
+++ b/packages/lit-dev-content/src/components/litdev-tutorial.ts
@@ -649,7 +649,6 @@ export class LitDevTutorial extends LitElement {
       this._idx++;
       this._writeUrl();
       if (this._manifest.steps.length > 1 && this._projectLocation) {
-        // User has advanced in the tutorial and we are on the last step.
         const eventFired = reportTutorialMetrics({
           idx: this._idx,
           tutorialUrl: this._projectLocation,
--- a/packages/lit-dev-content/src/components/litdev-tutorial.ts
+++ b/packages/lit-dev-content/src/components/litdev-tutorial.ts
@@ -34,6 +34,10 @@ import './litdev-icon-button.js';
 import {Task, TaskStatus} from '@lit-labs/task';
 import {PostDoc} from 'postdoc-lib';
 import type {PlaygroundPreview} from 'playground-elements/playground-preview';
+import {
+  reportTutorialMetrics,
+  TutorialMetricEvent,
+} from '../util/gtag-helpers.js';
 
 const CHECK_TIMEOUT_MS = 10000;
 export interface TutorialStep {
@@ -324,7 +328,15 @@ export class LitDevTutorial extends LitElement {
       this._oldManifestValue = manifestTaskValue;
       this._readUrl();
       // Manifest loaded. Good indicator that the tutorial has initially loaded.
-      this.reportMetrics(this._idx, this._projectLocation);
+      const eventFired = reportTutorialMetrics({
+        idx: this._idx,
+        numSteps: this._manifest.steps.length,
+        tutorialUrl: this._projectLocation,
+        hasRecordedStart,
+        hasRecordedEnd,
+      });
+
+      this._handleTutorialMetricEvent(eventFired);
     }
 
     const htmlTaskValue = this._htmlTask.value;
@@ -636,13 +648,17 @@ export class LitDevTutorial extends LitElement {
       this._clearCheckingTimeout();
       this._idx++;
       this._writeUrl();
-      if (
-        this._manifest.steps.length > 1 &&
-        this._idx === this._manifest.steps.length - 1 &&
-        this._projectLocation
-      ) {
+      if (this._manifest.steps.length > 1 && this._projectLocation) {
         // User has advanced in the tutorial and we are on the last step.
-        this.reportMetrics(this._idx, this._projectLocation);
+        const eventFired = reportTutorialMetrics({
+          idx: this._idx,
+          tutorialUrl: this._projectLocation,
+          numSteps: this._manifest.steps.length,
+          hasRecordedStart,
+          hasRecordedEnd,
+        });
+
+        this._handleTutorialMetricEvent(eventFired);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -753,6 +773,23 @@ export class LitDevTutorial extends LitElement {
...
+      });
+      hasRecordedStart = true;
+    } else if (idx === this._manifest.steps.length - 1 && !hasRecordedEnd) {
+      window.gtag?.('event', 'tutorial_end', {
+        category: 'tutorials',
+        event_label: tutorialUrl,
+      });
+      hasRecordedEnd = true;
+    }
+  }
```
AndrewJakubowicz, line 792: Is it worth splitting out the gtag related logic into a separate analytics file, and making this a pure function that isn't bound to the component?

AndrewJakubowicz (commented):
This is awesome! I tested with https://tagassistant.google.com/ and everything looks great.

Two nitpicks:
 - Can we split and document the analytics code separate from the component. Even if just in the same file?
 - Is it worth adding an extra event like `tutorial_progress` that triggers at each step with the value of the step? Then we can see if certain tutorials have a drop off point?

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -753,6 +773,23 @@ export class LitDevTutorial extends LitElement {
...
+
+  private reportMetrics(idx: number, tutorialUrl: string) {
+    if (idx === 0 || !hasRecordedStart) {
+      window.gtag?.('event', 'tutorial_start', {
+        category: 'tutorials',
+        event_label: tutorialUrl,
+        value: idx,
+      });
+      hasRecordedStart = true;
+    } else if (idx === this._manifest.steps.length - 1 && !hasRecordedEnd) {
```
aomarks, line 785: Shouldn't we record each step between start and end, too?

packages/lit-dev-content/src/types/gtag.d.ts
```diff
@@ -0,0 +1,7 @@
+interface Window {
```
aomarks, line 1: License header

e111077:
WOW TIL @AndrewJakubowicz! Thanks so much for that link!

Moved the gtag logic to its own file but still requires a local private method to track state.

Added licenses.

Added `tutorial_progress` event

Possible confusions for progress is that if they go back, it won't record back progress but it will record if they go back and then right. WDYT? Should It also record back?

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -774,20 +789,17 @@ export class LitDevTutorial extends LitElement {
...
+  private _handleTutorialMetricEvent(event: TutorialMetricEvent) {
+    switch (event) {
+      case 'tutorial_start':
+        hasRecordedStart = true;
+        break;
+      case 'tutorial_end':
+        hasRecordedEnd = true;
+        break;
+      case 'tutorial_progress':
+      default:
```
aomarks, line 801: A `never` check would be good. Otherwise the `default` case doesn't need to be here.

packages/lit-dev-content/src/util/gtag-helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+  hasRecordedStart: boolean;
+  hasRecordedEnd: boolean;
+  numSteps: number;
+};
+
+export type TutorialMetricEvent =
+  | 'tutorial_start'
+  | 'tutorial_progress'
+  | 'tutorial_end'
+  | null;
```
aomarks, line 19:
```suggestion
  | null;

```

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -58,6 +62,9 @@ interface ExpandedTutorialStep {
 
 type CheckStatus = 'INDETERMINATE' | 'CHECKING' | 'PASSED' | 'FAILED';
 
+let hasRecordedStart = false;
```
aomarks, line 65: This is a module scoped variable, so does it assume new tutorials are always loaded with a full page reload? Seems like that could be a little brittle, since a refactoring to make this a little more SPAish could break that assumption.

aomarks (changes requested):

AndrewJakubowicz (approved):
Small code nits. Pre-approving. The analytic events are working great!

packages/lit-dev-content/src/components/litdev-tutorial.ts
```diff
@@ -300,6 +307,13 @@ export class LitDevTutorial extends LitElement {
     return this;
   }
 
+  update(changed: PropertyValues<this>): void {
+    if (this._manifest.header) {
+      document.title = `${this._manifest.header} Tutorial – Lit`;
+    }
+    super.update(changed);
+  }
```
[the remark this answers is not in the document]
e111077: It doesn't really matter because it's not like this is run on the server. Personally I think earlier in the lifecycle is better

packages/lit-dev-content/src/util/gtag-helpers.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export type TutorialMetricsOptions = {
```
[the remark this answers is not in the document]
e111077: lol nice catch! this one was evolved from a bunch of `'string'|'string2'`

e111077:
Question: should I be recording progress when they hit the back button? Currently only broadcasting forward

packages/lit-dev-content/src/util/gtag-helpers.ts
```diff
@@ -24,21 +24,26 @@ export const reportTutorialMetrics = ({
   hasRecordedEnd,
   numSteps,
 }: TutorialMetricsOptions): TutorialMetricEvent => {
+  // A user can load a tutorial from a URL at a step that is not 0, hence
+  // we shoul consider whether start has not been recorded already.
```
AndrewJakubowicz, line 29: Nit: "should"

AndrewJakubowicz:
I continue to happily LGTM this PR! No strong opinion on back button. Wouldn't it technically be inferable from multiple duplicate progress steps provided the user doesn't hit the end?

AndrewJakubowicz (approved):

aomarks (approved):

aomarks:
Sorry for delay

PR #795, opened 2022-05-02
[tutorials] Update list tutorial

augustjk:
- Change duration to 20 minutes
- Fix property misname in step 4
- Change solution in step 4 to use `.forEach` to avoid infinite loop while typing

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/working-with-lists/03/after/my-element.ts
+++ b/packages/lit-dev-content/samples/tutorials/working-with-lists/03/after/my-element.ts
@@ -1,5 +1,6 @@
 import {LitElement, html} from 'lit';
 import {customElement, property, state} from 'lit/decorators.js';
+import type {TemplateResult} from 'lit';
 
 @customElement('my-element')
 class MyElement extends LitElement {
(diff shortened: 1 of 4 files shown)


packages/lit-dev-content/samples/tutorials/working-with-lists/03/after/my-element.ts
```diff
@@ -1,5 +1,6 @@
 import {LitElement, html} from 'lit';
 import {customElement, property, state} from 'lit/decorators.js';
+import type {TemplateResult} from 'lit';
```
AndrewJakubowicz, line 3: This type import and the `const listItems: TemplateResult[] = [];` initialization need to be updated in the `before` file.

AndrewJakubowicz (approved):
Pre approval - one `before` file needs to be updated with the type import.

Great way to get around the infinite loop!

packages/lit-dev-content/samples/tutorials/working-with-lists/03/after/my-element.ts
augustjk: Oops, thanks for catching that! It's weird that prior to this TS was happily inferring the type for the array. I guess it has trouble following what happens inside `.forEach`'s callback.

e111077 (approved):
sad to hear about the loops. LGTM pending andrew's suggestion

PR #796, opened 2022-05-03
make aside headers same line as content

e111077:
@arthurevans pointed out that aside vertical height is precious space so make bolded title inline instead.

also seems to have been a slight bug with before no-header. OOPS!

| variant | before | after |
| -- | -- | -- |
| with header | image | image |
| `no-header` | image | image |

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
+++ b/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
@@ -189,36 +189,36 @@ customElements.define('my-element', MyElement);
 You can also insert an aside in your instructions by using the following format:
 
 ```html
-<litdev-aside type="positive">
+{% aside "positive" %}
 
 The first line is always a bolded header.
 
 Make sure to do `this`!
 
-</litdev-aside>
+{% endaside %}
 
-<litdev-aside type="warn" no-header>
+{% aside "warn" no-header %}
 
 The `no-header` will make sure that this line is not bolded.
 
 Beware of `this`!
 
-</litdev-aside>
+{% endaside %}
 
-<litdev-aside type="negative">
+{% aside "negative" %}
 
 Make sure NOT to do `this`!
 
 The following non-header lines here make sense to explain the assertion in
 the header line above.
 
-</litdev-aside>
+{% endaside %}
 
-<litdev-aside type="info">
+{% aside "info" %}
 
 Check out more info [in this docs section](/docs/templates/expressions/#well-formed-html).
 
-</litdev-aside>
+{% endaside %}
 ```
 
 *Note:* markdown will only be parsed as markdown if there is an empty line
--- a/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
+++ b/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
@@ -197,7 +197,7 @@ Make sure to do `this`!
 
 {% endaside %}
 
-{% aside "warn" no-header %}
+{% aside "warn" "no-header" %}
 
 The `no-header` will make sure that this line is not bolded.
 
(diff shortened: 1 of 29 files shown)


AndrewJakubowicz:
Is it possible to get a space between the heading and the next sentence? Otherwise if there isn't any punctuation the heading touches the next line.

E.g.:
Screen Shot 2022-05-03 at 9 26 18 AM

e111077:
ugh this is because the compiler minifies the spaces between them 
image

that's why it wasn't an issue in screenshots above

e111077:
I have changed this PR to require the user to use the `{% aside "info" "no-header" %}` shortcode. This shortcode will insert the `<!-- htmlmin:ignore -->` comments which turns off minification for the contents.

Other positives is that it will notify you at build time that you have set an attribtue wrong unlike the `<litdev-aside class="positive">` which it should be `type="positive"` instead of `class`

packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md
```diff
@@ -189,36 +189,36 @@ customElements.define('my-element', MyElement);
...
 
 The first line is always a bolded header.
 
 Make sure to do `this`!
 
-</litdev-aside>
+{% endaside %}
 
-<litdev-aside type="warn" no-header>
+{% aside "warn" no-header %}
```
AndrewJakubowicz, line 200: Does this need wrapping quotes around `"no-header"`?

e111077: si

packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
```diff
@@ -263,6 +264,73 @@ export const playgroundPlugin = (
...
+        // This will throw an error at runtime if it does not match and will
+        // throw a TS build time error if we add another variant and forget to
+        // update this file.
+        neverReach(
+          variant,
+          `Invalid {% aside ${variant} %}.` +
+            ` variant "${variant}" is not an acceptable variant of:` +
+            ` ${acceptableVariants.join(', ')}.`
+        );
+      }
```
AndrewJakubowicz, line 318: I wouldn't act on this. For fun I attempted this using a switch. Attempted some code gold.

```
      switch(variant) {
        case acceptableVariants[0]:
        case acceptableVariants[1]:
        case acceptableVariants[2]:
        case acceptableVariants[3]:
          break;
        default:
          // This will throw an error at runtime if it does not match and will
          // throw a TS build time error if we add another variant and forget to
          // update this file.
          neverReach(
            variant,
            `Invalid {% aside ${variant} %}.` +
              ` variant "${variant}" is not an acceptable variant of:` +
              ` ${acceptableVariants.join(', ')}.`
          );
      }
```

AndrewJakubowicz (approved):
Reviewed and also manually tested! Nice!

packages/lit-dev-content/src/components/litdev-aside.ts
```diff
@@ -43,13 +43,17 @@ export class LitDevAside extends LitElement {
...
+      display: inline;
+    }
+
+    ::slotted(:first-child) {
       margin-block-start: 0;
     }
 
-    ::slotted(:first-of-type),
-    ::slotted(:last-of-type) {
+    ::slotted(:first-child),
```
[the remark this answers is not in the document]
e111077: this was a bug. it would bold both these:

```
<litdev-aside>

bolded first line

`bolded second line`

</litdev-aside>
```
because first is a `<p>` second is a `<code>` and they are each first-of-type. oops!

packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
e111077: yeah I usually do switches. I wonder how they minify in comparison.

PR #801, opened 2022-05-03
Add section on mutable objects. Fixes #694.

### LINKED ISSUE #694: How to handle mutable object should be more clear
We've heard that it is sometimes not clear how objects are handled across ReactiveElement and lit-html with respect to change detection.

By default, ReactiveElement property setters only request an update of the new value is not identical to the old value, so if you mutate an existing object, ReactiveElement will not re-render. You can handle mutable objects by overriding `hasChanged` to be `() => true`. You can also use an immutable pattern to changes are expressed as new objects.

Our example of `hasChanged` uses a string value, which might not be the most common case. We also don't seem to talk about this concern either independently enough or high enough on the reactive property page.

And it might be important to note that lit-html has different behavior: it assume objects may have change and only bails on updating the DOM (attributes, text, or properties) when _primitives_ are equal. `live()` might be part of this topic too.

arthurevans:
Fixes a big hole in the reactive properties chapter. Could be better, but this is a start.

### DIFF
--- a/packages/lit-dev-content/site/docs/components/properties.md
+++ b/packages/lit-dev-content/site/docs/components/properties.md
@@ -264,13 +264,40 @@ When a property changes, the following sequence occurs:
 
 1.  The property's setter is called.
 1.  The setter calls the component's `requestUpdate` method.
-1.  The property's old and new values are compared. If the property has a `hasChanged` function, it's called with the property's old and new values.
+1.  The property's old and new values are compared.
+    -  By default Lit uses a strict inequality test to determine if the value has changed (that is `newValue !== oldValue`).
+    -  If the property has a `hasChanged` function, it's called with the property's old and new values.
 1.  If the property change is detected, an update is scheduled asynchronously. If an update is already scheduled, only a single update is executed.
 1.  The component's `update` method is called, reflecting changed properties to attributes and re-rendering the component's templates.
+
+Note that if you mutate an object or array property, it won't trigger an update, because the object itself hasn't changed. For more information, see [Mutating object and array properties](#mutatingobjectandarrayproperties).
 
 There are many ways to hook into and modify the reactive update cycle. For more information, see [Reactive update cycle](/docs/components/lifecycle/#reactive-update-cycle).
 
 For more information about property change detection, see [Customizing change detection](#haschanged).
+
+### Mutating object and array properties
+
+Mutating an object or array doesn't change the object reference, so it won't trigger an update. You can handle object and array properties in one of two ways:
+
+-   Treat objects and arrays as immutable. For example, to remove an item from `myArray`, construct a new array:
+
+    ```js
+    this.myArray = [...this.myArray.slice(0, indexToRemove),
+                    ...this.myArray.slice(indexToRemove+1)];
+    ```
+
+    While this example is simple, using immutable data patterns with deeply nested objects can be complicated and error-prone, unless you're using an immutable data library like [Immer](https://immerjs.github.io/immer/).
+
+-   Call `requestUpdate()` to trigger an update directly.
+
+    ```js
+    this.myArray.splice(indexToRemove, 1);
+    this.requestUpdate();
+    ```
+
+    When called with no arguments, `requestUpdate()` schedules an update, without calling a `hasChanged()` function.
+
 
 ## Attributes {#attributes}
 
--- a/packages/lit-dev-content/site/docs/components/properties.md
+++ b/packages/lit-dev-content/site/docs/components/properties.md
@@ -283,8 +283,7 @@ Mutating an object or array doesn't change the object reference, so it won't tri
 -   **Immutable data pattern.** Treat objects and arrays as immutable. For example, to remove an item from `myArray`, construct a new array:
 
     ```js
-    this.myArray = [...this.myArray.slice(0, indexToRemove),
-                    ...this.myArray.slice(indexToRemove+1)];
+    this.myArray = this.myArray.filter((_, i) => i !== indexToRemove);
     ```
 
     While this example is simple, it's often helpful to use a library like [Immer](https://immerjs.github.io/immer/) to manage immutable data. This can help avoid tricky boilerplate code when setting deeply nested objects.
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,41 @@ When a property changes, the following sequence occurs:
...
 1.  If the property change is detected, an update is scheduled asynchronously. If an update is already scheduled, only a single update is executed.
 1.  The component's `update` method is called, reflecting changed properties to attributes and re-rendering the component's templates.
 
+Note that if you mutate an object or array property, it won't trigger an update, because the object itself hasn't changed. For more information, see [Mutating object and array properties](#mutatingobjectandarrayproperties).
+
 There are many ways to hook into and modify the reactive update cycle. For more information, see [Reactive update cycle](/docs/components/lifecycle/#reactive-update-cycle).
 
 For more information about property change detection, see [Customizing change detection](#haschanged).
 
+### Mutating object and array properties
```
AndrewJakubowicz, line 279: Worth adding a shorter heading `id`, e.g.: `{#mutating-properties}`

AndrewJakubowicz (approved):
This is a great addition to the documentation and would have really helped me out starting with Lit!

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,41 @@ When a property changes, the following sequence occurs:
...
+Mutating an object or array doesn't change the object reference, so it won't trigger an update. You can handle object and array properties in one of two ways:
+
+-   Treat objects and arrays as immutable. For example, to remove an item from `myArray`, construct a new array:
+
+    ```js
+    this.myArray = [...this.myArray.slice(0, indexToRemove),
+                    ...this.myArray.slice(indexToRemove+1)];
+    ```
+
+    While this example is simple, using immutable data patterns with deeply nested objects can be complicated and error-prone, unless you're using an immutable data library like [Immer](https://immerjs.github.io/immer/).
```
sorvell, line 290: Suggestion but feel free to ignore: maybe we can structure this recommendation more positively. Instead of, "this can be bad unless you do...", what about something like "It's often helpful to use a library to manage immutable data... [Immer seems fine to recommend] This can help avoid tricky boilerplate when setting deeply nested objects..."

sorvell (approved):
Small optional recommendation, but otherwise good!

packages/lit-dev-content/site/docs/components/properties.md
arthurevans: SG.

packages/lit-dev-content/site/docs/components/properties.md
arthurevans: Good catch. Done!

packages/lit-dev-content/site/docs/components/properties.md
arthurevans: Done.

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,46 @@ When a property changes, the following sequence occurs:
...
 1.  The property's setter is called.
 1.  The setter calls the component's `requestUpdate` method.
-1.  The property's old and new values are compared. If the property has a `hasChanged` function, it's called with the property's old and new values.
+1.  The property's old and new values are compared.
+    -  By default Lit uses a strict inequality test to determine if the value has changed (that is `newValue !== oldValue`).
+    -  If the property has a `hasChanged` function, it's called with the property's old and new values.
 1.  If the property change is detected, an update is scheduled asynchronously. If an update is already scheduled, only a single update is executed.
 1.  The component's `update` method is called, reflecting changed properties to attributes and re-rendering the component's templates.
 
+Note that if you mutate an object or array property, it won't trigger an update, because the object itself hasn't changed. For more information, see [Mutating object and array properties](#mutatingobjectandarrayproperties).
```
augustjk, line 273:
```suggestion
Note that if you mutate an object or array property, it won't trigger an update, because the object itself hasn't changed. For more information, see [Mutating object and array properties](#mutating-properties).
```

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,46 @@ When a property changes, the following sequence occurs:
...
+### Mutating object and array properties {#mutating-properties}
+
+Mutating an object or array doesn't change the object reference, so it won't trigger an update. You can handle object and array properties in one of two ways:
+
+-   **Immutable data pattern.** Treat objects and arrays as immutable. For example, to remove an item from `myArray`, construct a new array:
+
+    ```js
+    this.myArray = [...this.myArray.slice(0, indexToRemove),
+                    ...this.myArray.slice(indexToRemove+1)];
+    ```
```
augustjk, line 288: What do you think about this for creating a new array without an item?
```suggestion
    ```js
    this.myArray = this.myArray.filter((_, i) => i !== indexToRemove);
    ```
```

augustjk (commented):
I really like the additions. Very informative!

packages/lit-dev-content/site/docs/components/properties.md
arthurevans: Thanks!

packages/lit-dev-content/site/docs/components/properties.md
arthurevans: Good question... Is there a reason to favor one over the other? It seems like slice would be optimized for large arrays but ... I'm going to play the "I am not an engineer" card here.

sorvell: I prefer the filter.

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,46 @@ When a property changes, the following sequence occurs:
...
+-   **Manually triggering an update.** Mutate the data and `requestUpdate()` to trigger an update directly. For example:
+
+    ```js
+    this.myArray.splice(indexToRemove, 1);
+    this.requestUpdate();
+    ```
+
+    When called with no arguments, `requestUpdate()` schedules an update, without calling a `hasChanged()` function. But note that `requestUpdate()` only causes the _current_ component to update. That is, if a component uses the code shown above, **and** the component passes `this.myArray` to a subcomponent, the subcomponent will detect that the array reference hasn't changed, so it won't update.
+
+**In general, using top-down data flow with immutable objects is best for most applications.** It ensure that every component that needs to render a new value does (and does so as efficiently as possible, since parts of the data tree that didn't change won't cause components that rely on them to update).
```
sorvell, line 301:
```suggestion
**In general, using top-down data flow with immutable objects is best for most applications.** It ensures that every component that needs to render a new value does (and does so as efficiently as possible, since parts of the data tree that didn't change won't cause components that rely on them to update).
```

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,46 @@ When a property changes, the following sequence occurs:
...
+
+Mutating data directly and calling `requestUpdate()` should be considered an advanced use case. In this case, you the user (or some other system) needs to identify all the components that use the mutated data and call `requestUpdate()` on each one. When those components are spread across an application, this gets hard to manage. Not doing so robustly means that you might modify an object that's rendered in two parts of your application, but only have one part update.
```
sorvell, line 303:
```suggestion
Mutating data directly and calling `requestUpdate()` should be considered an advanced use case. In this case, you (or some other system) need to identify all the components that use the mutated data and call `requestUpdate()` on each one. When those components are spread across an application, this gets hard to manage. Not doing so robustly means that you might modify an object that's rendered in two parts of your application, but only have one part update.
```

packages/lit-dev-content/site/docs/components/properties.md
```diff
@@ -264,14 +264,46 @@ When a property changes, the following sequence occurs:
...
+
+If you know that a given piece of data is only used in a single part of the application, mutating the data directly is practical, and in some cases may be the most efficient option.
```
sorvell, line 305: I'd like to stay away from the efficient point cuz it's a bit of a can of worms, so maybe something like:

```suggestion
In very simple cases when you know that a given piece of data is only used in a single component, mutating the data directly and calling `requestUpdate` can be safely done if preferred.
```

arthurevans: Compromise text:

```suggestion
In simple cases, when you know that a given piece of data is only used in a single component, it should be safe to mutate the data and call `requestUpdate()`, if you prefer.
```

PR #803, opened 2022-05-05
Make dev command more resilient

e111077:
Until google/wireit#33 or 11ty/eleventy#984 are resolved we need to do this


aomarks (approved):
Nice

arthurevans:
🎉

PR #808, opened 2022-05-06
[tutorial] WC to Lit take 2

e111077:
Lots of changes to clarify this long tutorial. e.g.

* Moved the instructions before the code samples rather than just explaining what happened after.
* Split up a bunch of code samples.

This one is a honker, and maybe still a bit clunky but it's better than before IMO

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/wc-to-lit/06/before/rating-element.ts
+++ b/packages/lit-dev-content/samples/tutorials/wc-to-lit/06/before/rating-element.ts
@@ -10,7 +10,7 @@ export class RatingElement extends HTMLElement {
     this.shadowRoot!.querySelector<HTMLElement>('.rating')!.innerText = `${this.rating}`;
   }
 
-  set rating(value) {
+  set rating(value: number) {
     this._rating = value;
 
     if (!this.shadowRoot) {
--- a/packages/lit-dev-content/samples/tutorials/wc-to-lit/07/before/rating-element.ts
+++ b/packages/lit-dev-content/samples/tutorials/wc-to-lit/07/before/rating-element.ts
@@ -10,9 +10,7 @@ export class RatingElement extends HTMLElement {
     this.shadowRoot!.querySelector<HTMLElement>('.rating')!.innerText = `${this.rating}`;
   }
 
-  static get observedAttributes() {
-    return ['rating'];
-  }
+  static observedAttributes = ['rating'];
 
   attributeChangedCallback(attributeName: string, _oldValue: string, newValue: string) {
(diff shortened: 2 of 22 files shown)


sorvell (changes requested):
I suggest not merging this tutorial. I think it has too many issues, would need a lot of reworking, and is of questionable value. Concerns:

* The rating-element seems very strange. Having a `rating` and a `vote` setting seems both unusual, unnecessary, and complex for no reason.
* The style of the tutorial requires a very attentive, dedicated student to understand that the many many steps to implement the element from scratch should *not* turn you off because you'll fix them with Lit.
* Numerous places demonstrate bad practices and then fix them. This, again, seems like a bad pedagogical style since it requires a lot from the student: e.g. "Ok, I see, this bad way I just learned is actually wrong."
* I do *not* think we should demonstrate a custom element putting elements in its light DOM as is done here before introducing Shadow DOM.
* We should *not* use colors that are not a11y.
* We have chosen to de-emphasize stand alone use of lit-html, but this tutorial spends time showing how to do this before introducing `LitElement`.

e111077:
@sorvell this component makes sense maybe I can change rating to "initialRating" and expose a read-only "rating" but up and down vote make sense because imagine a YouTube rating, you need to be able to initialize it as "already voted". If I were to make this component perfect (but more complex) I'd do something [like this](https://lit.dev/playground/?mods=gists#gist=82468bf7f8b8b37ec35059bdc2939b5a)

I agree it's quite long, but unfortunately building a vanilla web component is a bad experience. I can change the difficulty to intermediate, though I think the concepts are "beginner" and the long time estimation clarifies ahead of time that it's a commitment. I think the concepts here are really important though.

Can you clarify what points are bad practices so I can work on fixing them?

Putting elements in light DOM – [Web.dev](http://web.dev/) & YT are built this way, and it's more of a stylistic / requirements choice. Regardless I think it's an excellent way to introduce the reasons behind Shadow DOM by "working through it" and reaching for WC tools as you need them.

Red/green is unavoidable. I can make it a11y compliant by changing the background if you want though.

I don't think separating Lit templates to their own step is much of an issue. It's a single step that shows a web component without LitElement, and our users should know what utility it provides. Also many people use it outside of Lit Element in the real world like haunted and Chrome Dev Tools which just released a viral article detailing how they don't use Lit element only lit html. I doubt this single step will convince our users to drop LitElement

sorvell:
The amount of work needed here seems unrealistic to try to get done for our next big push. I appreciate that work has already gone into this one, but I suggest we set it aside for now.

e111077:
agreed, I've already hidden the tutorial as a draft in another PR

PR #809, opened 2022-05-06
Further refine Wireit integration

aomarks:
- Bump to latest version of Wireit to bring in https://github.com/google/wireit/pull/189, which makes it safe to have multiple concurrent Wireit processes handling overlapping build graphs.

- Added `output:[]` to some server commands, so that they can run concurrently given the above new restriction.

- Renamed `:not-site` scripts to `:assets`, and `:site` scripts to `:eleventy`. I think these are more intuitive names.

- Separated out `dev:build:assets` and `prod:build:assets` scripts. This means that watching dev no longer runs rollup, which was unnecessary because we serve tsc output directly in dev mode.

- Updated the top-level `dev` script:
  - Ensure that the server is built first
  - Simplified to take advantage of safe concurrent Wireit processes

I'll be working on https://github.com/google/wireit/issues/33 next, which should simplify things even further.

### DIFF
--- a/packages/lit-dev-content/package.json
+++ b/packages/lit-dev-content/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "fonts:manrope": "wireit",
     "build": "wireit",
-    "build:not-site": "wireit",
+    "build:assets": "wireit",
     "build:site": "wireit",
     "build:site:watch": "wireit",
     "build:ts": "wireit",
@@ -37,7 +37,7 @@
     },
     "prod:build": {
       "dependencies": [
-        "build:not-site"
+        "build:assets"
       ],
       "files": [
         "site/**",
@@ -54,7 +54,7 @@
     },
     "build:site": {
       "dependencies": [
-        "build:not-site"
+        "build:assets"
       ],
       "output": [
         "_site"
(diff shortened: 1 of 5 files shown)


packages/lit-dev-content/package.json
```diff
@@ -65,23 +66,23 @@
...
+    "build:eleventy:watch": {
       "dependencies": [
-        "build:not-site"
+        "prod:build:assets"
       ],
       "output": [
         "_site"
       ],
-      "#comment": "This is awkward. We want to do the watch part for build:not-site ourselves, but let eleventy do its own watch mode itself. Maybe --incremental will do a quick rebuild of just the changed parts, and we don't need --watch?",
+      "#comment": "This is awkward. We want to do the watch part for build:assets ourselves, but let eleventy do its own watch mode itself. Maybe --incremental will do a quick rebuild of just the changed parts, and we don't need --watch?",
```
rictic, line 76: Should this comment be updated?

rictic (approved):

packages/lit-dev-content/package.json
aomarks: Not really, same issue still applies.

I think server mode will pretty much solve this, because we can annotate both the actual server, and the built-in `eleventy --watch` command with `"server":true` to indicate that they are long-running, and then have a `dev` script which depends on them both. Then we'll annotate which dependencies of those scripts require a restart (none for `eleventy`, the server's `build:ts` for the server), and then `npm run dev watch` should just do all the right things.

https://github.com/11ty/eleventy/issues/984 is the feature request to support `.tsbuildinfo` style cold start incremental for Eleventy (it's closed, but I think Eleventy does a weird "lodash style" thing where they close all feature requests regardless of whether they will be implemented). This would be even better, since then Wireit could be in control of the timing of when Eleventy re-builds, too.

augustjk (approved):

aomarks:
Also added `output:[]` to the server commands, so that Wireit doesn't prevent them from running concurrently, which was why tests were timing out.

Issue #810, opened 2022-05-06
Improve Gists UI


aomarks:
The Gists UI is OK, but could be a lot better.

![image](https://user-images.githubusercontent.com/48894/167198511-9e78da0d-dc85-4bf1-9214-782d277a3e3e.png)

### Ideas

- A *modal* share button, similar to the GitHub "Create pull request" and "Merge pull request" buttons, where you can switch between Long URLs and GitHub actions, and it's sticky.

![image](https://user-images.githubusercontent.com/48894/167198757-ceb71526-9288-4429-aec2-af9e0665032a.png)

- More information about what Gist is being displayed. Something a lot like GitHub's own Gists viewer:
![image](https://user-images.githubusercontent.com/48894/167198676-712e3855-5816-41b4-8e55-76637562bc00.png)

PR #811, opened 2022-05-06
Launch gists and remove old share button

aomarks:
Enables the GitHub Gists integration for everyone, without a mod.

image

Fixes https://github.com/lit/lit.dev/issues/533

Filed https://github.com/lit/lit.dev/issues/810 to track further improvements to the UI.


AndrewJakubowicz (approved):
Tested manually on Chrome. I think the only nitpick I realized while testing this is that if you use the shortcut `cmd+S` the focus remains in the editor, so it's not possible to use tab and enter to navigate the popup (between copy url and the github button).

sorvell (approved):
Love it!

e111077 (approved):

e111077 (approved):

aomarks:
> Tested manually on Chrome. I think the only nitpick I realized while testing this is that if you use the shortcut `cmd+S` the focus remains in the editor, so it's not possible to use tab and enter to navigate the popup (between copy url and the github button).

Done, using `delegatesFocus`. It doesn't seem to work in Firefox, even though everything seems to indicate that Firefox supports `delegatesFocus`. Works fine in Chrome and Safari...

e111077:
so goes the tale of delegatesFocus

aomarks:
> so goes the tale of delegatesFocus

There seem to be no issues filed about it not working correctly. https://bugzilla.mozilla.org/show_bug.cgi?id=1413836 is marked fixed since Firefox 94.

Am I using it wrong?

AndrewJakubowicz (approved):
This is an improvement! We may just need to track an issue to follow up on Firefox.

aomarks:
> This is an improvement! We may just need to track an issue to follow up on Firefox.

Yeah, https://bugzilla.mozilla.org/show_bug.cgi?id=1735667 seems like it is this issue, but it's also been marked fixed since 94 as well.

aomarks:
> Yeah, https://bugzilla.mozilla.org/show_bug.cgi?id=1735667 seems like it is this issue, but it's also been marked fixed since 94 as well.

Commented at https://bugzilla.mozilla.org/show_bug.cgi?id=1735667#c10 with a repro

Issue #753, opened 2022-04-16
[tutorial] @eventName reference on intro-to-lit/03 confusing


grayside:
### Problem
On https://lit.dev/tutorial/#03 there is a bit of text:

```
@eventName is a special syntax for adding an event handler using an expression.
```

This is not explained anywhere in the page, the referenced Event page, or the solution.

### Impact

This was a distraction that led to taking a few extra minutes trying to figure out how to follow the steps on the page.

arthurevans:
Thanks for the report. The sentence is supposed to explain the step you've just done, and obviously it didn't work.

Before I try to improve the wording, can you add a little more detail about what was confusing? 

 For example, was it unclear that the _eventName_  `@eventName` was a placeholder? Or was there something else about the sentence that didn't make sense?

grayside:
I thought `@eventName` was a literal directive like `@property` that I was meant to place above something. I don't really understand how the @property annotation is used as I'm not familiar with that syntax.

I wasn't sure which of the following was intended:

### Confusion:
```
@eventName
@input=${this.changeName}
```

### Maybe it was for the next instruction?
```
@eventName
changeName(event: Event) {
  const input = event.target as HTMLInputElement;
  this.name = input.value;
}
```

### Big confusion
```
@eventName
    return html`
      <p>Hello, ${this.name}</p>
      @input=${this.changeName}
      <input @input=${this.changeName} placeholder="Enter your name">
    `;
```

arthurevans:
Ack! Yeah, that would be super confusing. And re-reading the step it says to add the "@input" expression "inside the tag" which is a little too vague.

Here's a rewrite:

Find the input element and add a declarative event listener:

```
<input @input=${this.changeName} placeholder="Enter your name">
```

This `@input` expression adds `this.changeName` as an event listener for the `input` event. (You can use this syntax for any event listener—just replace `input` with any event name and `this.changeName` with any JavaScript expression that evaluates to an event handler function.)

Issue #703, opened 2022-02-11
Inconsistencies in tutorial code


arthurevans:
Review of #697 identified several other inconsistencies in the tutorial code.

Specifically:

- In Step 6 there's indentation that changes on the solve.
- Everything up to Step 6 omits the `export` from the class definition.
- Step 6 doesn't have `{attribute: false}` on the listItems property, but subsequent steps do.

Original discussion:
https://github.com/lit/lit.dev/pull/697#pullrequestreview-879687175

PR #794, opened 2022-04-30
launch the lit tutorial catalog

e111077:
Targeted launch day is Monday May 9th

fixes #788

PR #813, opened 2022-05-07
Intro to lit updates

arthurevans:
Fixes several reported issues and a couple of unreported issues in the intro tutorial. 

- Removes `requestUpdate()` call and favors immutable data pattern instead.
- Adds missing export statements and other inconsistencies noted in #703.
- Fixes a confusingly-written note reported in #753.
- Fixes two errors in JS starting files that I noticed while making the other changes.

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/intro-to-lit/01/before/my-element.js
+++ b/packages/lit-dev-content/samples/tutorials/intro-to-lit/01/before/my-element.js
@@ -1,2 +1,1 @@
 import {LitElement, html} from 'lit';
-import {customElement} from 'lit/decorators.js';
--- a/packages/lit-dev-content/samples/tutorials/intro-to-lit/01/before/my-element.js
+++ b/packages/lit-dev-content/samples/tutorials/intro-to-lit/01/before/my-element.js
@@ -1,1 +1,2 @@
-export {};
+import {LitElement, html} from 'lit';
+import {customElement} from 'lit/decorators.js';
--- a/packages/lit-dev-content/samples/tutorials/intro-to-lit/05/after/todo-list.ts
+++ b/packages/lit-dev-content/samples/tutorials/intro-to-lit/05/after/todo-list.ts
@@ -2,12 +2,12 @@ import {LitElement, html} from 'lit';
 import {customElement, property, query} from 'lit/decorators.js';
 
 @customElement('todo-list')
-class ToDoList extends LitElement {
-  @property()
+export class ToDoList extends LitElement {
+  @property({attribute: false})
   listItems = [
-      { text: 'Start Lit tutorial', completed: true },
-      { text: 'Make to-do list', completed: false }
-    ];
+    { text: 'Start Lit tutorial', completed: true },
+    { text: 'Make to-do list', completed: false }
+  ];
 
   render() {
     return html`
@@ -25,9 +25,9 @@ class ToDoList extends LitElement {
   input!: HTMLInputElement;
 
   addToDo() {
-    this.listItems.push({text: this.input.value, completed: false});
+    this.listItems = [...this.listItems,
+        {text: this.input.value, completed: false}];
     this.input.value = '';
-    this.requestUpdate();
   }
 }
 
--- a/packages/lit-dev-content/samples/tutorials/intro-to-lit/05/after/todo-list.ts
+++ b/packages/lit-dev-content/samples/tutorials/intro-to-lit/05/after/todo-list.ts
@@ -1,10 +1,10 @@
 import {LitElement, html} from 'lit';
-import {customElement, property, query} from 'lit/decorators.js';
+import {customElement, state, query} from 'lit/decorators.js';
 
 @customElement('todo-list')
 export class ToDoList extends LitElement {
-  @property({attribute: false})
-  listItems = [
+  @state()
+  private _listItems = [
     { text: 'Start Lit tutorial', completed: true },
     { text: 'Make to-do list', completed: false }
   ];
@@ -13,7 +13,7 @@ export class ToDoList extends LitElement {
     return html`
       <h2>To Do</h2>
       <ul>
-        ${this.listItems.map((item) =>
+        ${this._listItems.map((item) =>
           html`<li>${item.text}</li>`)}
       </ul>
       <input id="newitem" aria-label="New item">
@@ -25,7 +25,7 @@ export class ToDoList extends LitElement {
   input!: HTMLInputElement;
 
   addToDo() {
-    this.listItems = [...this.listItems,
+    this._listItems = [...this._listItems,
         {text: this.input.value, completed: false}];
     this.input.value = '';
   }
--- a/packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
+++ b/packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
@@ -19,7 +19,7 @@ In Lit, most things start with defining a _component_. Here we've given you a st
 
     {% endswitchable-sample %}
 
-    The `MyElement` class provides the implementation for your new component, and the `@customElement` decorator registers it with the browser as a new element type named `my-element`.
+    The `MyElement` class provides the implementation for your new component. <ts-js><span slot="ts">The <code>@customElement</code> decorator registers it with the browser as a new element type named <code>my-element</code>.</span><span slot="js">The <code>customElements.define</code> call registers it with the browser as a new element type named <code>my-element</code>.</span></ts-js>
 
     The "TODO" text should disappear from the **Result** pane, showing that the component is defined. But your component is still missing something—it doesn't display anything.
 
@@ -36,3 +36,12 @@ In Lit, most things start with defining a _component_. Here we've given you a st
     The `render()` method defines your component's internal DOM. The `html` tag function processes a template literal and returns a `TemplateResult`—an object that describes the HTML for Lit to render. Every Lit component should include a `render()` method.
 
 You should see the greeting message in the **Result** pane.
+
+{% aside "info" %}
+
+Decorators.
+
+The TypeScript version of this code uses the `@customElement` decorator to register your class with the browser as a new element. The JavaScript version uses `customElements.define()` to do the same thing. This is a matter of style. You can use decorators—or not use decorators—in either language. For more information, see [Decorators](/docs/components/decorators/).
+
+{% endaside %}
+
--- a/packages/lit-dev-content/site/tutorials/content/intro-to-lit/03.md
+++ b/packages/lit-dev-content/site/tutorials/content/intro-to-lit/03.md
@@ -8,13 +8,13 @@ Here we've provided a name tag component with a message and an input element. In
 
 *   **Add a declarative event listener.**
 
-    Find the input element and add this expression inside the tag:
+    Find the input element and add a declarative event listener:
 
-    ```html
-    @input=${this.changeName}
+    ```js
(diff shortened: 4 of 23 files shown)


packages/lit-dev-content/samples/tutorials/intro-to-lit/01/before/my-element.js
```diff
@@ -1 +1,2 @@
-export {};
+import {LitElement, html} from 'lit';
+import {customElement} from 'lit/decorators.js';
```
arthurevans, line 2:
```suggestion
import {LitElement, html} from 'lit';
```

Oops. Might've been in a bit of a hurry there.

packages/lit-dev-content/site/tutorials/content/intro-to-lit/03.md
```diff
@@ -8,13 +8,13 @@ Here we've provided a name tag component with a message and an input element. In
 
 *   **Add a declarative event listener.**
 
-    Find the input element and add this expression inside the tag:
+    Find the input element and add a declarative event listener:
 
-    ```html
-    @input=${this.changeName}
+    ```js
```
arthurevans, line 13:
```suggestion
    ```html
```

packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
```diff
@@ -19,7 +19,7 @@ In Lit, most things start with defining a _component_. Here we've given you a st
 
     {% endswitchable-sample %}
 
-    The `MyElement` class provides the implementation for your new component, and the `@customElement` decorator registers it with the browser as a new element type named `my-element`.
+    The `MyElement` class provides the implementation for your new component. <ts-js><span slot="ts">The <code>@customElement</code> decorator registers it with the browser as a new element type named <code>my-element</code>.</span><span slot="js">The <code>customElements.define</code> call registers it with the browser as a new element type named <code>my-element</code>.</span></ts-js>
```
e111077, line 22: The HTML minifier unfortunately deletes this space in the prod build (it doesn't run in the dev build). We just need to add a `&nbsp;` to force a space there, minification or not.

```suggestion
    The `MyElement` class provides the implementation for your new component.&nbsp;<ts-js><span slot="ts">The <code>@customElement</code> decorator registers it with the browser as a new element type named <code>my-element</code>.</span><span slot="js">The <code>customElements.define</code> call registers it with the browser as a new element type named <code>my-element</code>.</span></ts-js>
```

packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
```diff
@@ -36,3 +36,12 @@ In Lit, most things start with defining a _component_. Here we've given you a st
...
 
 You should see the greeting message in the **Result** pane.
+
+{% aside "info" %}
+
+Decorators.
+
+The TypeScript version of this code uses the `@customElement` decorator to register your class with the browser as a new element. The JavaScript version uses `customElements.define()` to do the same thing. This is a matter of style. You can use decorators—or not use decorators—in either language. For more information, see [Decorators](/docs/components/decorators/).
+
+{% endaside %}
```
e111077, line 46: is it worth mentioning the TS/JS toggle. Surprising amount of people online don't notice that it exists and have issues with some of our decorators. Specifically `@query`

packages/lit-dev-content/site/tutorials/content/intro-to-lit/03.md
```diff
@@ -8,13 +8,13 @@ Here we've provided a name tag component with a message and an input element. In
...
-    Find the input element and add this expression inside the tag:
+    Find the input element and add a declarative event listener:
 
     ```html
-    @input=${this.changeName}
+    <input @input=${this.changeName} placeholder="Enter your name">
     ```
 
-    <code>@<var>eventName</var></code> is a special syntax for adding an event handler using an expression.
+    This `@input` expression adds `this.changeName` as an event listener for the `input` event. (You can use this syntax for any event listener—just replace input with any event name and `this.changeName` with any JavaScript expression that evaluates to an event handler function.)
```
e111077, line 17: `event handler function` -> `event handler`. Because AFAIK the only way to do 

`el.addEventListener('name', 'props', {passive: true});`

in javascript is to use an event handler object:

```
const clickHandler = {

    // This fires when the handler is called
    handleEvent(e) { console.log('clicked'); }

    // You can specify event options same as addEventListener
    capture: false;
    passive: true;
}

html`<button @click=${clickHandler}>Click Me</button>`
```

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback

packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
e111077: ah i see you talk about query a later step 👍 👍

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
```diff
@@ -38,21 +38,32 @@ You can use standard JavaScript in Lit expressions to create conditional or repe
     }
 
     addToDo() {
-      this.listItems.push({text: this.input.value, completed: false});
+      this.listItems = [...this.listItems,
+          {text: this.input.value, completed: false}];
```
e111077, line 42: I still personally disagree with this method being the one we choose to highlight.

Especially since it's an input `@property` to the component – any references the user of this component hands to this component is no longer the one used inside this component. The user then has to query the component again and get the new refernce. Then there's idea of cloning objects etc.

e111077 (commented):
Leaving these as comment so I don't block the merge since i'm OOO on mon-tues

packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
arthurevans: Yeah, I talk about that but mentioning the toggle explicitly is a good idea. It was added after the tutorial was written, so there were a number of confusing bits.

packages/lit-dev-content/site/tutorials/content/intro-to-lit/03.md
arthurevans: 👍

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
arthurevans: Yeah, I can see the argument here with regard to it being an input property. It would probably make more sense for it to *not* be an input property, since the component manipulates it. 

@kevinpschaaf and @sorvell both felt we should not be recommending `requestUpdate()` for most use cases.

packages/lit-dev-content/site/tutorials/content/intro-to-lit/03.md
arthurevans:
```suggestion
    This `@input` expression adds `this.changeName` as an event listener for the `input` event. (You can use this syntax for any event listener—just replace input with any event name and `this.changeName` with any JavaScript expression that evaluates to an event handler.)
```

packages/lit-dev-content/samples/tutorials/intro-to-lit/05/after/todo-list.ts
```diff
@@ -2,12 +2,12 @@ import {LitElement, html} from 'lit';
 import {customElement, property, query} from 'lit/decorators.js';
 
 @customElement('todo-list')
-class ToDoList extends LitElement {
-  @property()
+export class ToDoList extends LitElement {
+  @property({attribute: false})
```
sorvell, line 6: This should be `@state`

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
e111077: This is more of a "alright but I'mma complain the entire time" feel free to ignore the noise 🌬️

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
```diff
@@ -38,21 +38,32 @@ You can use standard JavaScript in Lit expressions to create conditional or repe
...
 
-    As the name suggests, `requestUpdate()` triggers the component to update. Setting a reactive property calls this automatically. Since you're not setting `listItems` here, but only mutating the array, you need to call `requestUpdate()` yourself.
-
-    The `@query` decorator is a handy way of getting a reference to a node in your component's internal DOM. It's basically equivalent to calling code like this in your constructor:
+    The `@query` decorator (used in the TypeScript version of the code) is a handy way of getting a reference to a node in your component's internal DOM. It's  equivalent to this code in the JavaScript version:
 
     ```js
-    this.input = this.shadowRoot.querySelector('#newitem');
+    get input() {
+      return this.renderRoot?.querySelector('#newitem') ?? null;
```
sorvell, line 53: Why change this to use `renderRoot`?

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
sorvell: I think we should make this `@state` and then split this into 2 steps. One just intro'ing the `@state` and adding the list and the other doing the `add` stuff.

sorvell (changes requested):

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
kevinpschaaf: If it's too big of a change to split out another step, an aside that says something like this seems fine:

> Notice that `listItems` uses the `@state()` decorator. That makes it reactive just like `@property()`, but is internal to our element, not part of the public API of our element. Read more about [Internal reactive state](https://lit.dev/docs/components/properties/#internal-reactive-state).

arthurevans: I think this latter should work. It'll just be a matter of making sure I fix all the references to listItems. Which should probably also become private or protected `_listItems`?

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
arthurevans: I'm just adding the code that our TS->JS translator generates so this code matches the solved code.

packages/lit-dev-content/site/tutorials/content/intro-to-lit/01.md
arthurevans: Added note, with a little extra switchable sample as a demo.

packages/lit-dev-content/site/tutorials/content/intro-to-lit/05.md
arthurevans: Made this change, and added a new paragraph about state. The asides just seem way to distracting to put inline.

arthurevans:
Made the requested changes:

- Changed `listItems` to internal state, named `_listItems`.
- Added a note about the JS/TS language switch on the intro page.

Also updated the last page to link to new tutorials.

sorvell (approved):

arthurevans:
This is ready to go as soon as #794 merges.

e111077:
Waiting for https://github.com/lit/lit.dev/pull/820#pullrequestreview-966868027

Are we doing this? 👏👏👏

arthurevans:
I added some links to the new tutorials, so I think we should do these basically at the same time. Or this one can even follow #794, there's no hard requirement.

PR #815, opened 2022-05-07
Aside applies the indentation of content

e111077:
When aside is currently in an indent. e.g.

```md
*   Here is a bulleted list that has an aside

    {% aside "info" %}
    Here are the contents
    {% endaside %}
```

The shortcode will translate it into markdown as so:

```md
*   Here is a bulleted list that has an aside

<litdev-aside type="info">
<!-- htmlmin:ignore -->

    Here are the contents

<!-- htmlmin:ignore -->
</litdev-aside>
```

Thus throwing off the MD compiler. In this PR we get the leading whitespace in front of the contents, and prepend them to all of our HTML nodes:

```js
// content looks like this:

"    Here are the contents"
```

The regex matches `^\s*` which is the first 4 spaces and applies it to our HTML nodes

```md
*   Here is a bulleted list that has an aside

    <litdev-aside type="info">
    <!-- htmlmin:ignore -->

    Here are the contents

    <!-- htmlmin:ignore -->
    </litdev-aside
```

This is how it looks now at 3 different indentation levels +0, +1, +2

image

### DIFF
--- a/packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
+++ b/packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
@@ -321,7 +321,7 @@ export const playgroundPlugin = (
 
       // Matches the first line's leading whitespace and applies it to the other
       // HTML nodes to preserve the same indentaiton.
-      const contentIndent =  (content.match(/^\s*/)?? [''])[0];
+      const contentIndent = (content.match(/^\s*/) ?? [''])[0];
 
       // htmlmin:ignore will prevent minifier from formatting the contents.
       // otherwise, in the prod build, there will not be a space between the
--- a/packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
+++ b/packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
@@ -320,7 +320,7 @@ export const playgroundPlugin = (
       const noHeaderAttribute = noHeader === 'no-header' ? ' no-header' : '';
 
       // Matches the first line's leading whitespace and applies it to the other
-      // HTML nodes to preserve the same indentaiton.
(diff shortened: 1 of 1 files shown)


arthurevans:
🥇

packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
```diff
@@ -318,15 +318,20 @@ export const playgroundPlugin = (
       }
 
       const noHeaderAttribute = noHeader === 'no-header' ? ' no-header' : '';
+
+      // Matches the first line's leading whitespace and applies it to the other
+      // HTML nodes to preserve the same indentaiton.
```
AndrewJakubowicz, line 323: Typo: indentation

packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
```diff
@@ -318,15 +318,20 @@ export const playgroundPlugin = (
...
+      const contentIndent = (content.match(/^\s*/) ?? [''])[0];
```
AndrewJakubowicz, line 324: Nit: Because `*` can match zero spaces, as far as I can tell, this will never fail to match and never return `null`. Any string I test without spaces will match zero spaces and return `''`.
Maybe can remove the `?? ['']`.

AndrewJakubowicz (approved):
Nice!

packages/lit-dev-tools-cjs/src/playground-plugin/plugin.ts
AndrewJakubowicz: Oh I see this makes the type checker happy. Resolving.

justinfagnani (approved):

PR #818, opened 2022-05-09
[tutorials] Working with lists: encourage immutable data pattern

augustjk:
Updates step 7 to encourage immutable data pattern by creating a new array rather than mutating and calling `requestUpdate()`. Adds link to that section of the doc.

Also updates all property usages to use `@state` decorator instead for consistency with Intro to Lit tutorial.


sorvell (changes requested):
Can we just do this for all the array properties? (wherever we have `@property({attribute: false})`)

sorvell (approved):

PR #820, opened 2022-05-09
[svg-tutorial] update demo height, based on view height

taylor-vann:
Update `repeat-pattern` to use view height instead of parent height.
This prevents the occasional `0` height issue when a demo is initially rendered.

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/svg-templates/00/before/index.html
+++ b/packages/lit-dev-content/samples/tutorials/svg-templates/00/before/index.html
@@ -2,18 +2,22 @@
 <html>
   <head>
     <style>
-      html, body {
-         height: 100%;
-         padding: 0;
-         margin: 0;
-      }
-      :root {
+      body {
         --background-color: #7acbf1;
         --font-color: #fff;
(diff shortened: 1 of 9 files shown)


AndrewJakubowicz:
Could you rebase or merge in the newest main branch. The images in the tutorials catalog have been removed.

AndrewJakubowicz (changes requested):
Documenting some of the offline chat threads.
Unfortunately still able to repro the issue.

To reproduce issue, open the preview url: https://pr820-5f5e714---lit-dev-5ftespv5na-uc.a.run.app/tutorials/svg-templates/?mods=tutorialCatalog

Alternate `cmd + R` and `cmd + shift + R` and note that the pattern in the preview does not display a lot of the time.

@sorvell Mentioned that it's the `rect` element that is not sized correctly in the error case.

For reference on this PR, the problematic case looks like this:
Screen Shot 2022-05-09 at 11 39 41 AM

taylor-vann:
updated, the behavior no longer happens on local

e111077 (approved):

PR #828, opened 2022-05-10
Misc fixes to tooltip tutorial

aomarks:
- Hid some files on the first step.
- Made file order consistent.
- Fixed some typos and minor grammar things (also found a couple repeated words in some random other pages on the site).
- Added "Here's what you'll do in the upcoming steps:" to the first page, because it seemed like a user could think they were supposed to complete the task list right away.
- Softened language about `fixed` solving stacking context. I think it might actually be a problem fairly often, so "vast majority" seemed a bit strong.
- Removed folding from the last step. It was showing nothing, but seems like we should show the whole final product?

### Open issues

- In step 6, the box 3 tooltip goes back to the wrong place after setting up `floating-ui`, and there is an error in the console: `process is not defined`.

- In step 8, I was slightly confused by the "a flag is used" sentence. Can you clarify what this is supposed to mean? Does "flag" refer to the "once" option?

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/tooltip/00/before/index.html
+++ b/packages/lit-dev-content/samples/tutorials/tooltip/00/before/index.html
@@ -7,6 +7,7 @@
   </style>
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons"
   rel="stylesheet">
+  <script src="./process.js"></script>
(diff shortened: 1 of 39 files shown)


sorvell (approved):
Excellent fixes. Thanks!

PR #831, opened 2022-05-11
[infra] Cloud Build/Run artifact cleanup

### LINKED ISSUE #786: Configure cleaning up closed PR preview deploys.
### Context

Currently we have over 1000 revisions and were temporarily blocked from building and releasing new preview urls.

The solution is to add a cleanup step to the main script:

https://cloud.google.com/run/docs/tutorials/configure-deployment-previews#cloudbuild-cleanupyaml

This should remove all revisions for closed PRs.

augustjk:
Resolves #786 

The Cloud Build and Cloud Run to deploy the lit.dev site for production and PR previews were leaving artifacts like service revisions and docker images that were no longer needed.

This PR adds a `cloudbuild-clean.yaml` that can be manually triggered, or run on schedule, which will do the following:
- Fetch all open PRs for this repo
- Fetch all revisions we have for the main site service and playground service
- Identify revisions that are associated with any open PRs or are younger than 1 week
- Remove any traffic tags for revisions that were not identified in the above step
- Update each service with the trimmed down list of traffic tags
- Delete all revisions that no longer have any traffic tags
- Fetch all docker image tags in GCR and identify ones that are not associated with any active revisions and older than a week
- Delete all tags from image, then delete the image
- Fetch all cache docker images (created by kaniko) and identify ones older than a week
- Delete all tags from image, then delete the image

Run example: https://pantheon.corp.google.com/cloud-build/builds;region=global/b24b1ec5-6947-4e7b-aae9-88ad9505ff18?project=lit-dev-site

Once this is merged, the cloudbuild trigger can be updated to point to `main` branch and scheduled.

### DIFF
--- a/cloudbuild-main.yaml
+++ b/cloudbuild-main.yaml
@@ -119,8 +119,8 @@ steps:
       - beta
       - run
       - services
-      - update-traffic # Service name
-      - lit-dev-playground
+      - update-traffic
+      - lit-dev-playground # Service name
       - '--region=$_DEPLOY_REGION'
       - '--platform=managed'
       - '--quiet'
--- a/cloudbuild-main.yaml
+++ b/cloudbuild-main.yaml
@@ -37,6 +37,8 @@ steps:
       - '--image=$_GCR_HOSTNAME/$PROJECT_ID/$REPO_NAME/lit-dev:$COMMIT_SHA'
       - '--quiet'
       - '--no-traffic'
+      # If this tag format changes, also update route-traffic- steps below and
+      # the cleanup script in packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
       - '--tag=main-$SHORT_SHA'
       # IMPORTANT: If you change --memory, be sure to also change
       # --max-old-space-size in ./Dockerfile, and this same flag in
@@ -70,6 +72,8 @@ steps:
       - '--image=$_GCR_HOSTNAME/$PROJECT_ID/$REPO_NAME/lit-dev:$COMMIT_SHA'
       - '--quiet'
       - '--no-traffic'
+      # If this tag format changes, also update route-traffic- steps below and
+      # the cleanup script in packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
       - '--tag=main-$SHORT_SHA'
       # IMPORTANT: If you change --memory, be sure to also change
       # --max-old-space-size in ./Dockerfile, and this same flag in
--- a/packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
+++ b/packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
@@ -49,9 +49,11 @@ async function main() {
 
   // fetch list of revisions across both services
   console.log('fetching list of revisions');
-  const {data: revisions} = await run.projects.locations.revisions.list({
+  const {data} = await run.projects.locations.revisions.list({
     parent: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}`,
   });
+
+  let revisions = data;
 
   if (!revisions.items) {
     throw new Error('found no revision items');
@@ -77,6 +79,7 @@ async function main() {
 
   // traffic tags must be cleaned up before a revision can be deleted
   const services = ['lit-dev', 'lit-dev-playground'];
+  let serviceUpdated = false;
   await Promise.all(
     services.map(async (serviceName) => {
       console.log(`fetching ${serviceName} service`);
@@ -92,7 +95,8 @@ async function main() {
 
       const trafficToKeep = data.spec.traffic.filter((t) => {
         return (
-          Boolean(t.percent) || // there's traffic being routed here (i.e. prod)
+          // precence of percent indicates traffic is being routed here (i.e. prod)
+          Boolean(t.percent) ||
           revisionsToKeepByRecency.has(t.revisionName!) ||
           openPrs.has(parseInt(t.tag!.slice(2), 10))
         );
@@ -101,7 +105,7 @@ async function main() {
       console.log(`${serviceName} traffic to keep count`, trafficToKeep.length);
 
       if (trafficToKeep.length === 0) {
-        throw new Error('found no ${service} traffic to keep');
+        throw new Error(`found no ${serviceName} traffic to keep`);
       }
 
       if (data.spec.traffic.length !== trafficToKeep.length) {
@@ -111,27 +115,31 @@ async function main() {
           name: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}/services/${serviceName}`,
           requestBody: data,
         });
+        serviceUpdated = true;
       }
     })
   );
 
   // revision status takes some time to update
-  console.log('waiting 30 seconds for traffic tags to update');
-  await sleep(30_000);
+  if (serviceUpdated) {
+    console.log('waiting 30 seconds for traffic tags to update');
+    await sleep(30_000);
 
-  // fetching new list of revisions to get the now untagged (inactive) ones
-  console.log('fetching new revisions list');
-  const {data: newRevisions} = await run.projects.locations.revisions.list({
-    parent: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}`,
-  });
+    // fetch new list of revisions to get the now untagged (inactive) ones
+    console.log('fetching new revisions list');
+    const {data} = await run.projects.locations.revisions.list({
+      parent: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}`,
+    });
+    revisions = data;
+  }
 
-  if (!newRevisions.items) {
+  if (!revisions.items) {
     throw new Error('found no new revision items');
   }
 
   console.log('looking for inactive revisions to delete');
   const imageDigestsToKeep = new Set<string>();
-  for (const rev of newRevisions.items) {
+  for (const rev of revisions.items) {
     if (!rev.metadata) {
       throw new Error('revision has no metadata');
     }
@@ -173,7 +181,7 @@ async function main() {
   // clean up gcr lit-dev images that are no longer used
   {
     console.log('fetching images from gcr');
-    const resp = await request({
+    const response = await request({
       url: `https://us.gcr.io/v2/${PROJECT_ID}/${REPO_NAME}/lit-dev/tags/list`,
       method: 'GET',
       headers: {
@@ -181,7 +189,7 @@ async function main() {
       },
     });
 
-    const data = resp.data as {
+    const data = response.data as {
       manifest: Record<string, {tag: string[]; timeCreatedMs: string}>;
       tags: string[];
     };
@@ -219,7 +227,7 @@ async function main() {
   // clean up gcr lit-dev/cache images created by kaniko
   {
     console.log('fetching cache images from gcr');
-    const resp = await request({
+    const response = await request({
       url: `https://us.gcr.io/v2/${PROJECT_ID}/${REPO_NAME}/lit-dev/cache/tags/list`,
       method: 'GET',
       headers: {
@@ -227,7 +235,7 @@ async function main() {
       },
     });
 
-    const data = resp.data as {
+    const data = response.data as {
       manifest: Record<string, {tag: string[]; timeCreatedMs: string}>;
       tags: string[];
     };
--- a/packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
+++ b/packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
@@ -186,6 +186,8 @@ async function main() {
       tags: string[];
     };
 
+    console.log('image tag count', data.tags.length);
+
     for (const [digest, manifest] of Object.entries(data.manifest)) {
       if (
         !imageDigestsToKeep.has(digest) &&
@@ -216,6 +218,7 @@ async function main() {
 
   // clean up gcr lit-dev/cache images created by kaniko
   {
+    console.log('fetching cache images from gcr');
     const resp = await request({
       url: `https://us.gcr.io/v2/${PROJECT_ID}/${REPO_NAME}/lit-dev/cache/tags/list`,
       method: 'GET',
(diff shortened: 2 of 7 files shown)


cloudbuild-main.yaml
```diff
@@ -119,8 +119,8 @@ steps:
       - beta
       - run
       - services
-      - update-traffic # Service name
-      - lit-dev-playground
+      - update-traffic
+      - lit-dev-playground # Service name
```
aomarks, line 123: Was this re-ordering significant?

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
+import {google} from 'googleapis';
```
aomarks, line 1: Add license header

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+import {Octokit} from 'octokit';
+import {request} from 'gaxios';
```
aomarks, line 3: If Node 18 is available to us here, then we can use the new built in `fetch` API (same as the browser's API) so I think won't need gaxios.

https://nodejs.org/en/blog/announcements/v18-release-announce/

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+  console.log({PROJECT_ID, REPO_NAME, _DEPLOY_REGION});
+  throw new Error(
+    'missing one or more required environment variables: PROJECT_ID, REPO_NAME, _DEPLOY_REGION'
+  );
+}
+
+const sleep = (time: number) =>
+  new Promise((resolve) => setTimeout(resolve, time));
+
+async function main() {
```
aomarks, line 20: Could use top-level await here (Node 14.8+) so the `main` isn't required.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+  const octokit = new Octokit();
+
+  const openPrs = new Set<number>();
+
+  console.log('fetching open PRs from github');
+  const openPrsIterator = octokit.paginate.iterator(octokit.rest.pulls.list, {
+    owner: 'lit',
+    repo: `${REPO_NAME}`,
```
aomarks, line 28:
```suggestion
    repo: REPO_NAME,
```

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+  // fetch list of revisions across both services
+  console.log('fetching list of revisions');
+  const {data} = await run.projects.locations.revisions.list({
+    parent: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}`,
+  });
+
+  let revisions = data;
+
+  if (!revisions.items) {
+    throw new Error('found no revision items');
```
aomarks, line 59: Maybe add a comment that there must at least be revisions for the live sites?

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+
+      if (!data.spec || !data.spec.traffic) {
+        throw new Error(`found no traffic for ${serviceName} service`);
+      }
+
+      console.log(`${serviceName} traffic count`, data.spec.traffic.length);
+
+      const trafficToKeep = data.spec.traffic.filter((t) => {
+        return (
+          // precence of percent indicates traffic is being routed here (i.e. prod)
```
aomarks, line 98:
```suggestion
          // presence of percent indicates traffic is being routed here (i.e. prod)
```

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+          Boolean(t.percent) ||
```
aomarks, line 99: This is false for both `undefined` and `0`. Which seems reasonable, just checking that's the intention (maybe `0` means something subtly different?). Could consider checking both explicitly for clarity.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+    })
+  );
+
+  // revision status takes some time to update
+  if (serviceUpdated) {
+    console.log('waiting 30 seconds for traffic tags to update');
+    await sleep(30_000);
+
+    // fetch new list of revisions to get the now untagged (inactive) ones
```
aomarks, line 128: Curious: do we need to actually fetch the new list of revisions, given that we know which traffic tags we removed earlier?

aomarks (changes requested):
Wow, this is fantastic. Great work figuring out all these APIs!

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
augustjk: I did that initially but the `for await` here can't be done top-level
https://github.com/lit/lit.dev/blob/179b9e2f636c98de076faf03573de4bb013a70bd/packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts#L33

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
augustjk: Due to the way I'm checking it, and maybe also because it's inside an async function(?), TS still thinks `REPO_NAME` could be undefined so it didn't match the param type without the template literal/conversion to string. Ah, I suppose I could do `REPO_NAME as string`.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
augustjk: The way we're using it, we should never see `0` as we always direct 100% of traffic to a single revision. https://cloud.google.com/run/docs/rollouts-rollbacks-traffic-migration#split-traffic
I don't believe cloud run even creates `0` values there unless the yaml is manually edited to be that way. And if the value somehow is `0`, we'll still keep it for a week based on the recency check. No traffic and older than a week seems safe to get rid of.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+
+      console.log(`${serviceName} traffic count`, data.spec.traffic.length);
+
+      const trafficToKeep = data.spec.traffic.filter((t) => {
+        return (
+          // precence of percent indicates traffic is being routed here (i.e. prod)
+          Boolean(t.percent) ||
+          revisionsToKeepByRecency.has(t.revisionName!) ||
+          openPrs.has(parseInt(t.tag!.slice(2), 10))
```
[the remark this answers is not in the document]
augustjk: If we want to err on the safe side for non-matches we do also need to consider tags that look like `main-xxxxxxx` for revisions made from the main branch deploy, that we do wish to delete if older than a week.

Definitely will be adding comments here and other related places, and consider switching to regex.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+        });
+        serviceUpdated = true;
+      }
+    })
+  );
+
+  // revision status takes some time to update
+  if (serviceUpdated) {
+    console.log('waiting 30 seconds for traffic tags to update');
+    await sleep(30_000);
```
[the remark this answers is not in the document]
augustjk: I found no API to actually verify if the update was done. 30 seconds is a conservative guess on my part and I haven't seen it fail with this wait. Before adding the wait, I have seen when we try to delete revisions too soon, it just fails because it still thinks it's associated with a tag. The cleanup job will fail and the next run will catch anything that wasn't deleted due to early termination.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
augustjk: That's a good point. The initial thinking was to just refetch for the active status in case some didn't finish updating yet based on the 30 second wait so we'd only delete ones that are ready to be deleted, but I did find that the new fetch could return revisions that are inactive but still fail to delete because the service wasn't fully updated yet. Since we're going to wait either way, it seems fine to just keep track of the revisions to delete from the code before this. I'll refactor to do that.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+
+    // fetch new list of revisions to get the now untagged (inactive) ones
+    console.log('fetching new revisions list');
+    const {data} = await run.projects.locations.revisions.list({
+      parent: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}`,
+    });
+    revisions = data;
+  }
+
+  if (!revisions.items) {
```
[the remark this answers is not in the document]
augustjk: This will go away in the refactor to not refetch the list.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+      rev.status.conditions.find((cond) => cond.type === 'Active')!.status !==
+        'True' &&
+      // extra timestamp check in case the revision was just created and not active because of that
+      new Date(rev.metadata.creationTimestamp!) < ONE_WEEK_AGO
+    ) {
+      console.log(`deleting revision ${rev.metadata.name}`);
+      await run.projects.locations.revisions.delete({
+        name: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}/revisions/${rev.metadata.name}`,
+      });
+      // wait 1 second to not hit api limit
```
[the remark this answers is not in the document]
augustjk: https://cloud.google.com/run/quotas#api
For writes, API limit is 60 per 60 seconds and I've hit that during testing without the wait. This was also more of an issue when we had hundreds of revisions to delete initially. I suppose we could also chunk the requests to groups of 60 and wait 1 minute in between too but figured this is easier. When run on a schedule, this likely won't have more than 60 to delete at a time so maybe it would be okay to not wait, but also there's no reason to really rush this either. WDYT?

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,272 @@
...
+          console.log(`deleting tag ${tag}`);
+          await request({
+            url: `https://us.gcr.io/v2/lit-dev-site/lit.dev/lit-dev/manifests/${tag}`,
+            method: 'DELETE',
+            headers: {
+              authorization: 'Bearer ' + token,
+            },
+          });
+        }
+        console.log(`deleting image ${digest}`);
```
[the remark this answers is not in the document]
augustjk: The delete request fails due to the still associated tag.

augustjk (commented):
@aomarks Thanks for the thorough review!

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
aomarks: Sounds reasonable. Didn't realize `gaxios` is a `googleapis` package itself.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
aomarks: Oh, weird, didn't know that wasn't supported.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
aomarks: Ah yeah, the `${}` is actually masking the type error there, since that does implicit `undefined` string conversion. Would prefer something explicit.

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,313 @@
...
+        revisionsToKeep.add(t.revisionName);
+        return true;
+      }
+
+      if (t.tag) {
+        // Tags should look like "pr529-8b0837f" or "main-45ace0e"
+        if (t.tag.startsWith('pr')) {
+          // pr529-8b0837f
+          //   ^^^
+          const prNumber = parseInt(t.tag.slice(2), 10);
```
aomarks, line 145: Seems like this is depending on the behavior that `parseInt` ignores extra noise, (e.g. `parseInt("529-8b0837f") // 529`).

I think a regexp might be more explicit/clear:

```ts
const prMatch = t.tag.match(/^pr(\d+)-/);
if (prMatch) {
  const prNumber = Number(t.tag[1]);
  ...
} else if (!t.tag.startsWith('main')) {
  ...
}
```

aomarks (changes requested):
Looks good! Just a suggestion about the pr number extraction.

augustjk:
> If you wrote the above check as if (!PROJECT_ID || !REPO_NAME || !_DEPLOY_REGION) then I think TypeScript will correctly narrow the type, so you won't need the as string below. Plus it's shorter and handles the empty string case.

@aomarks I tried that but TS doesn't narrow the type if the check happens outside the function definition. Since the `main` only runs once anyway, I suppose there's no waste in moving that check to the top of the function instead.

aomarks (approved):
Awesome!

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  // Fetch list of revisions across both services.
+  // This API provides the revision's created time but does not contain data
+  // of any associated traffic tags.
+  console.log('Fetching list of revisions');
+  const {data} = await run.projects.locations.revisions.list({
+    parent: `projects/${PROJECT_ID}/locations/${_DEPLOY_REGION}`,
+  });
+
+  let revisions = data;
```
AndrewJakubowicz, line 83: Ultra minor nit. Can be `const`.

AndrewJakubowicz (approved):
Super nice! Extremely impressive script with effective usage of minimally documents APIs.

Thank you!

packages/lit-dev-cloudbuild-cleanup/src/cleanup.ts
augustjk: Oops good catch!

PR #832, opened 2022-05-11
Use 1-indexed and non-padded tutorial URLs

aomarks:
Previously, we used 0-indexed and left-padded tutorial URLs. For example, the 2nd step of a tutorial had URL `/tutorials/<name>/#01`.

Now it is 1-indexed, and not padded. For example: `/tutorials/<name>/#2`.

I think 0-indexed is more intuitive, because it means the URLs are aligned with the "Step 2/8" text on the page.

I think left-padding was unnecessary in URLs, plus by removing it, we have a convenient way to detect people still on the 0-indexed scheme, and fix the URLs to be 1-indexed. This way nobody will get jumped ahead if they reload before/after the new scheme is deployed.

Also fixes a bug where we would set the URL to `/tutorials/<name>#2` (no trailing slash), but then after reloading it would turn into `/tutorials/<name>/#2`. Now we always set the final form.

### DIFF
--- a/packages/lit-dev-content/src/components/litdev-tutorial.ts
+++ b/packages/lit-dev-content/src/components/litdev-tutorial.ts
@@ -660,6 +660,16 @@ export class LitDevTutorial extends LitElement {
       hash = hash.slice(1);
     }
     this._idx = hash === '' ? 0 : this._slugToIdx(hash) ?? this._idx;
+    if (hash.startsWith('0')) {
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz (approved):
Nice! Also tested manually in the preview url. Went over the word-viewer and async directive checking that solve and all the steps were working.

Nice!

PR #835, opened 2022-05-11
Fix floating-ui and motion dependencies in tooltip and carousel tutorials

aomarks:
- Upgrades playground-elements to bring in https://github.com/google/playground-elements/pull/313, which lets us import `floating-ui` without any special handling.

- Removed special handling for `process.env` error from incorrect module resolution in tooltip tutorial.

- Fixed `"dependencies"` in tooltip and carousel tutorials. Previously they were listed in the tutorial `project.json` file -- which has no effect. Now, they are correctly inside a `package.json` file, so that the playground sees it.

- Also: made the order of files in the carousel tutorials consistent.


AndrewJakubowicz (approved):
LGTM. Also did some manual testing on the preview url and compared with production.

PR #841, opened 2022-05-11
Reference superclass explicitly in the mixin demo

luzhang:
Hey Lit team,

The playground [example](https://lit.dev/playground/#sample=docs/mixins/highlightable) on the live site is not working, the second element is not rendered correctly because `super.styles` does not work here. So I implemented a quick fix to replace `super.styles` with `Highlightable(LitElement).styles`.

Actually the Lit documentation also mentions about explicitly referencing the superclass instead of using `super`: https://lit.dev/docs/components/styles/#inheriting-styles-from-a-superclass

Thanks,

### DIFF
--- a/packages/lit-dev-content/samples/docs/mixins/highlightable/element-two.ts
+++ b/packages/lit-dev-content/samples/docs/mixins/highlightable/element-two.ts
@@ -5,7 +5,7 @@ import {Highlightable} from './highlightable.js'
 @customElement('element-two')
 export class ElementTwo extends Highlightable(LitElement) {
   static styles = [
-    super.styles || [],
+    Highlightable(LitElement).styles || [],
     css`:host { display: block; }`
   ];
   render(){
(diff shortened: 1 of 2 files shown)


arthurevans:
Fixed sample is here: 

https://pr841-fd214f1---lit-dev-5ftespv5na-uc.a.run.app/docs/composition/mixins/#creating-mixins-for-litelement

Original sample is here:

https://lit.dev/docs/composition/mixins/#creating-mixins-for-litelement

Verified that it works in JavaScript, but fails in TypeScript.

I'm inclined to merge this but /cc @kevinpschaaf as resident expert in case there's some other way we should be doing this.

kevinpschaaf:
It would be better to not apply the mixin twice, so would you mind pulling the `HilightableLitElement` out as a const in `element-two.ts`? ala this: https://lit.dev/playground/#gist=0b5486728fc473c171200ca60cbc6bd2

Also, in looking at this I realized it would be safest for the mixin to also be defensive about being applied to a class that already has `styles`, so I've updated the `highlightable.ts` to do that also. Mind folding these changes in?

luzhang:
Thanks Arthur and Kevin!

Yeah caching the mixin makes sense, it will improve the performance a bit. I also applied your suggestion in the `highlightable.ts`. Can you take another look on my changes?

Arthur, you said it failed in TS, can you elaborate it more?

PR #845, opened 2022-05-17
[tutorial] implement custom attribute tutorial suggestions

### LINKED ISSUE #827: [tutorials] Custom attribute converter refinements
General:
- Consider using `navigator.language` instead of an explicitly provided locale, in both the converter and the rendered output
- For each step, would like to see us select the first file to be edited (see individual notes)

Step 2:
- `date-display.ts` file should be selected
- Order of tabs should not change

Step 3:
- `index.html file` should be selected
- Explicitly say to delete the `<script>` tag

Step 4:
- `date-converter.ts` should (probably) be selected
- Consider putting `date-converter.ts` before index.html?
- Very minor: I was a bit disappointed that I was invited to provide my own locale but never got to see it put to use

Step 5:
- `date-display.ts` file should be selected

e111077:
fixes #827

Note: when pressing `solve` in some steps. the `selected` flag may change the currently-viewed file.

Also the dateConverter is now a _function_ that takes in a `locale` and returns a `ComplexAttributeConverter<Date>`. It's what you probably would do in production. LMK if it complicates it too much

### DIFF
--- a/packages/lit-dev-content/samples/tutorials/custom-attribute-converter/01/before/project.json
+++ b/packages/lit-dev-content/samples/tutorials/custom-attribute-converter/01/before/project.json
@@ -1,8 +1,8 @@
 {
   "extends": "/samples/checkable-tutorial-base.json",
   "files": {
-    "date-display.ts": {},
     "index.html": {},
+    "date-display.ts": {"selected": true},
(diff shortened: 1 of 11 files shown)


arthurevans (approved):
LGTM.

PR #850, opened 2022-05-23
SSR & progressive perf improvements to landing page

e111077:
This PR tackles performance on the landing page and some general perf fixes to the header and footer using SSR.

General mobile lighthouse scores are up from low 80s to low 90s. General problems still arise around loading gtag and 200kb of fonts.

Try for yourself:

before: https://pr850-f9aa1c0---lit-dev-5ftespv5na-uc.a.run.app/
after: https://pr850-a7e1b8a---lit-dev-5ftespv5na-uc.a.run.app/

General Changes:

- homepage
  - SSRd a bunch of components
  - Lazy load SVGs and images except those above fold
  - do not load litdev-aside by default
  - removes `dsd-pending` styles on `noscript`
  - Some things in the header and footer are SSRd on every page, so `nohydrate` was removed and hydration helpers loaded every page
- litdev-search
  - now SSRd
  - moved off of lion-combobox because could not be SSRd
  - hydrated on focus
  - now has focus indication with bolded underline
- litdev-version-selector
  - Is now SSRd
  - mods transform is lazy loaded to work on Node
- lazy-svg
  - lazy loads SVGs that load with `<use>`
  - strategies are on idle, on visible, and eagerly
  - SSRd to prevent layout shift


AndrewJakubowicz:
This is awesome!

Two issues found during manual testing.


### Issue 1

The combobox dropdown options currently cause a horizontal scroll to appear.

Repro: Full screen chrome. Open page. Type "dir". Notice a horizontal scroll bar has appeared and it is possible to scroll right.

Screenshot:
Screen Shot 2022-05-24 at 4 23 01 PM

Similarly in the mobile view, there is a slight horizontal movement.

### Issue 2

Switching between full screen search and mobile search is broken. The first one works, but then the second doesn't seem to.

Repro:

Refresh the page. Type in the search that loaded and see a dropdown. Then resize the page to access the other search bar. Typing in the other search bar does not result in suggestions rendering.

Screenshot:

https://user-images.githubusercontent.com/15080861/170148144-fd654a07-f655-4588-8df4-829c2aedc3b8.mov

e111077:
Thanks for the review @AndrewJakubowicz, fixed!

AndrewJakubowicz:
Sorry for these nitpicks. Did an A11y pass to ensure the behavior is the same as the prior search.
VoiceOver is working fantastic on Safari using Mac.
Manually tested keyboard behaviors. The only behavior that I've found is different is with the ESC key.

When using arrow keys to navigate the dropdown, or if the dropdown is open. ESC should close it.
If the focus is in the search input and ESC is pressed, it should also clear the input.

Another very subtle, and related issue. If on a narrow screen and interacting with the search. If you press escape it closes the drawer and makes the search dropdown invisible but still clickable.

Repro (shown in video). Type a search and notice the top link in the mobile search. Click with mouse away from drawer and notice it shuts. After re-opening the drawer the mouse indicates there is no link under the search. Now focus the search to show the dropdown and press ESC. The drawer shuts (but the dropdown is just invisible this time). Re-open the drawer and notice you can now click something beneath the search. Clicking it takes you to the link that is there.

I also had Mac VoiceOver assistant active while recording this. Repro on Safari and Mac.

Video repro:

https://user-images.githubusercontent.com/15080861/170372231-90b607da-d395-43bf-a5df-d883833abd90.mov

e111077:
Resolved!

packages/lit-dev-content/src/components/lazy-svg.ts
```diff
@@ -0,0 +1,144 @@
...
+    </svg>`;
+  }
+
+  updated(changed: PropertyValues<this>) {
+    if (changed.has('loading')) {
+      if (this.loading === 'idle') {
+        this._idleLoad();
+      } else if (this.loading === 'visible') {
+        // Must happen in `updated` because in non-SSR the DOM is not ready
+        this._visibleLoad();
```
AndrewJakubowicz, line 106: If `this.loading === 'visible'`, will this register an Intersection observer every update?

packages/lit-dev-content/src/global/hydrate-common-components.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const searches = Array.from(document.body.querySelectorAll('litdev-search'));
+
+for (const search of searches) {
+  search.addEventListener(
+    'focus',
+    async () => {
+      await import('../components/litdev-search.js');
+
+      search.focus();
```
AndrewJakubowicz, line 27: On a slow network. If someone starts typing, and then the element upgrades. It seems the input is cleared. Is there anyway to preserve the input during hydration?

Repro. Throttle network. Refresh page. Click search and start typing. On upgrade the input is cleared.
Expected behavior, the input is not cleared and maybe the suggestions appear.

https://user-images.githubusercontent.com/15080861/170516948-2b07da1e-5be7-478f-81d8-4751e9cb5a54.mov

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -208,81 +342,145 @@ class LitDevSearch extends LitElement {
...
+      case 'Escape':
+        const oldText = this._searchText;
+        this._searchText = '';
+        this._suggestions = [];
+        this._selectedIndex = -1;
+        // prevent the input from closing dialog if there was text
+        if (oldText.trim()) {
+          e.stopPropagation();
         }
+        break;
```
AndrewJakubowicz, line 424: I'm slightly unsure what `oldText` is being used for here.

Currently when manually testing, pressing escape always closes and clears the input. Previously pressing escape while focused on the suggestions would only close them and not clear input. And pressing escape while focussed on input would clear input.

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -140,60 +170,164 @@ class LitDevSearch extends LitElement {
...
+          aria-owns="items"
+          aria-expanded=${isExpanded ? 'true' : 'false'}
+          aria-autocomplete="list"
+          aria-activedescendant=${activeDescendant}
+          .value=${live(this._searchText)}
+          @input=${this._onInput}
+          @keydown=${this._onKeydown}
+          @focus=${this._onFocus}
+          @blur=${this._onBlur}
+        />
```
AndrewJakubowicz, line 278: Nit, void element. 
```suggestion
        >
```

AndrewJakubowicz (commented):
Really nice implementation of litdev-search and litdev-search-options.

Found a tiny hydration issue on slow networks.

packages/lit-dev-content/src/components/lazy-svg.ts
e111077: Yes, but I think we can skip the complexity of fixing this potential issue since our use cases will never run into it.

packages/lit-dev-content/src/components/litdev-search.ts
e111077: The comment should be updated to say "drawer" instead. This is if you put it in a drawer, you're typing and you hit escape, it will clear it but it won't close the dialog.

If the searchbox is empty and you hit escape, you likely wanted to close the drawer

packages/lit-dev-content/src/global/hydrate-common-components.ts
e111077: it's because of the `live` directive. When it hydrates, it will render `''` in the textfield. I also found another small issue for when there are two of these on the page and they hydrate. the static nature of the loading is causing funkyness

AndrewJakubowicz (approved):
LGTM! Pre-approving with a last nitpick.

With keyboard navigation. Pressing enter on a selection doesn't provide any feedback, whilst clicking causes the dropdown to disappear.

Also did a manual test on phone. Thank you so much for doing this.

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -238,6 +240,17 @@ export class LitDevSearch extends LitElement {
 
   private _listboxClicked = false;
 
+  update(changed: PropertyValues<this>) {
+    if (!this._searchText && this._inputEl) {
+      console.log(this._inputEl.value);
+      // this is required on hydration to make sure we don't blow away the
+      // input text due to the `live` directive which is necessary to prevent
+      // issues with typing on the virtual keyboard in Safari on iOS.
```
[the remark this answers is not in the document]
e111077: https://github.com/material-components/material-web/pull/877

If you type and set `this._searchText` Lit only knows that the binding at `.value` is stale. `live` prevents this from happening because it first checks the value of the input before setting it.

PR #851, opened 2022-05-27
More Home page SSR improvements

e111077:
**This PR addresses several main issues with the landing page:**

* MWC and sub-bundles are being pulled in on first load
* The SSR module graph is interfering with the client-side module graph
* Mobile drawer assets are being pulled in on desktop
* Reduced chain requests on prod that block render **(biggest speed improvement)**

**Results:**

General mobile lighthouse scores are up from low 90s to 95-98.

**In depth changes:**

* implemented `litdev-ripple-icon-button`
  * SSR-able mwc-button lookalike.
  * Doesn't need to be hydrated to work
  * it's square-icle rather than round for fun
  * does not pull in MWC
* consolidated `mobile-nav.js` into `hydrate-common-components.js`
  * created helpers for `onIdle` and `onMediaQuery` for hydration since D.R.Y.
* mwc-drawer logic only pulled in on mobile media query
* Made a JS bundle just for the server / 11ty
  * As we added more files to SSR, it kept making the client bundle split, nearly obviating the bundler
* Load lit.js directly instead of lit-hydrate-support in Prod
  * In prod we manually bundle all of lit.js together so lit/lit-hydrate-support.js is kludged in with lit.js
  * This means that if we load js/global/lit-hydrate-support.js it will ALWAYS make a second request to lit.js

loading lit.js directly (blue line is first render):

| Before | After |
| -- | -- |
| image | image |


packages/lit-dev-content/site/_includes/default.html
```diff
@@ -8,15 +8,22 @@
     <meta name="description" content="Simple. Fast. Web Components.">
     <link rel="icon" href="{{ site.baseurl }}/images/flame-favicon.svg">
     <link rel="mask-icon" href="{{ site.baseurl }}/images/flame-favicon.svg" color="#324fff">
-    <link rel="modulepreload" href="{{ site.baseurl }}/js/global/lit-hydrate-support.js">
+    <script type="modulepreload" src="{{ site.baseurl }}/js/tslib.js"></script>
+    <link rel="modulepreload" href="{{ site.baseurl }}/js/lit.js">
+    <link rel="modulepreload" href="{{ site.baseurl }}/js/global/hydrate-common-components.js">
```
AndrewJakubowicz, line 13: Brilliant!

packages/lit-dev-content/site/_includes/default.html
```diff
@@ -8,15 +8,22 @@
...
     {% inlinecss "mods.css" %}
     {% inlinejs "global/apply-mods.js" %}
     {% inlinejs "global/initialize-typescript-attribute.js" %}
     {% inlinejs "global/mobile-drawer.js" %}
-    <script type="module" src="{{ site.baseurl }}/js/global/lit-hydrate-support.js"></script>
+    {% if env.DEV %}
+      <script type="module" src="{{ site.baseurl }}/js/global/lit-hydrate-support.js"></script>
+    {% else %}
+      <!-- Loads lit-hydration since we manually chunk this -->
+      <link rel="module" href="{{ site.baseurl }}/js/lit.js">
```
AndrewJakubowicz, line 25: A bit confused about this. Should this be `modulepreload`? I couldn't find `module` in the link types: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

packages/lit-dev-content/rollup.config.js
```diff
@@ -101,6 +97,27 @@ export default [
       }),
     ],
   },
+  {
+    name: 'SSR',
```
AndrewJakubowicz, line 101: When testing this manually, a `mods-[hash].js` was created next to the `rollupout/server/ssr.js` file. This contains the mods import that litdev-search requires. I guess it doesn't matter that there is a tiny bit of duplication for the overall benefit.

AndrewJakubowicz (commented):
Part of the way through review. Everything looking great so far. Reviewed build change and manual testing.

Manually tested on phone with throttled network as well as with Mac voiceover assistant for the `litdev-ripple-icon-button` usages.

Still need to review the new component and hydration helper code implementation. Great stuff!

packages/lit-dev-content/src/util/hydration-helpers.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+/**
+ * Calls a function on idleCallback or in 200ms if on Safari. _Note:_ 200ms is
+ * what Astro uses for their idle callback fallback.
+ * @param hydrationCallback Function to be called on idleCallback.
+ */
+export const onIdle = (hydrationCallback: () => void) => {
+  if ('requestIdleCallback' in window) {
+    (window as any).requestIdleCallback(hydrationCallback);
```
AndrewJakubowicz, line 14: `as any` is not required here (I think)?

packages/lit-dev-content/src/util/hydration-helpers.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+};
+
+/**
+ * Calls a given function when window.matchMedia matches for the first time.
+ *
+ * @param queryString Media query string to trigger the callback.
+ * @param hydrationCallback Callback to be called when the given media query matches.
+ */
+export const onMediaQuery = (
```
AndrewJakubowicz, line 27: No action required. Maybe this name can be updated to reflect that it is once. `onMediaQueryOnce`?

packages/lit-dev-content/src/util/hydration-helpers.ts
```diff
@@ -0,0 +1,44 @@
...
+  queryString: string,
+  hydrationCallback: () => void
+) => {
+  const query = window.matchMedia(queryString);
+  const onQuery = (e: MediaQueryList | MediaQueryListEvent) => {
+    if (e.matches) {
+      hydrationCallback();
+      query.removeEventListener('change', onQuery);
```
AndrewJakubowicz, line 35: Instead of removing the event listener, can pass `{once: true}` to the `addEventListener` call.

[Reference](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#:~:text=to%20false.-,once,-A%20boolean%20value)

packages/lit-dev-content/src/global/hydrate-common-components.ts
```diff
@@ -29,3 +31,63 @@ for (const search of searches) {
...
+  ) as HTMLDetailsElement | null;
+  if (activeDetails === null) {
+    return;
+  }
+  activeDetails.open = true;
+
+  const drawer = document.body.querySelector('mwc-drawer');
+  // Wait for the drawer to begin opening.
+  drawer?.addEventListener(
+    'opening',
```
AndrewJakubowicz, line 52: Why is this additional `opening` event required?

AndrewJakubowicz (commented):
Looking fantastic.  A couple clarifying questions.

Also while testing with `npm run dev`, getting: `GET http://localhost:5415/js/lit.js net::ERR_ABORTED 404 (Not Found)` error.

packages/lit-dev-content/src/util/hydration-helpers.ts
e111077: Can't do this because it might call the listener but not match and then remove itself

packages/lit-dev-content/src/global/hydrate-common-components.ts
e111077: TBH I'm pretty sure it never worked before...

confirmed it never worked before:

https://lit.dev/docs/resources/community/
https://pr851-041c2a2---lit-dev-5ftespv5na-uc.a.run.app/docs/resources/community/

open those pages on mobile, make the height very low, and open the side drawer. The `Community` thing should be visible on both

e111077: https://user-images.githubusercontent.com/5981958/172723304-8b05337d-08b1-4398-9a0c-4fa1489b1fd2.mov

packages/lit-dev-content/site/_includes/default.html
e111077: no, because in the rollup config we manually chunk everything in `node_modules/lit` into `lit.js` that means that the files loaded by hydrate helpers are located at the top of this `lit.js` file. If i were to `<script type="module" src="/js/global/lit-hydrate-support.js">` then even in prod it would make 2 render-blocking requests to first `/js/global/lit-hydrate-support.js` then to `/js/lit.js`. We want to stop this from happening so we just load `lit.js` directly

packages/lit-dev-content/rollup.config.js
e111077: I think this is fine since it's just at build time and the workaround would complicate the bundling imo

e111077:
Also fixed the lit.js 404 issue. Forgot to exclude dev build from link[rel="moduelpreload"][href="/js/lit.js"]

packages/lit-dev-content/src/util/hydration-helpers.ts
AndrewJakubowicz: Oops! Missed that behavior.

packages/lit-dev-content/src/global/hydrate-common-components.ts
AndrewJakubowicz: Great repro! Great fix!

packages/lit-dev-content/site/_includes/default.html
AndrewJakubowicz: Makes sense. I think I'm just confused what `link` with `rel="module"` means?

e111077: OMG that is a TYPO!

e111077: how does any of this even work???

e111077: ohh it's because it's in lit.js and anything loading lit ever will setup correctly. Maybe we don't even need an else statement here

e111077: Keeping it in here just to be safe

AndrewJakubowicz (approved):
Nice work! Love the blazing fast site!

PR #853, opened 2022-06-03
[docs] Fix async task example 

### LINKED ISSUE #559: "Names controller" example does not work
The Names Controller example is giving me fetch errors:

https://lit.dev/docs/composition/controllers/#asynchronous-tasks

Looks like CORS + playground
image

augustjk:
Fixes #559 

### Changes
- Replaced json-generator API with [swapi.dev](https://swapi.dev/)
- There was no more a need for kind to id map since swapi had semantic route names
- Removed the "Error: " in the error rendering as it was causing duplicates since stringifying the error object adds it as well.

### DIFF
--- a/packages/lit-dev-content/samples/docs/controllers/names/names-controller.ts
+++ b/packages/lit-dev-content/samples/docs/controllers/names/names-controller.ts
@@ -16,13 +16,13 @@ export class NamesController {
         if (!kind?.trim()) {
           return initialState;
         }
-        const response = await fetch(`${Names.baseUrl}${Names.kindIdMap[kind]}`);
-        const result = await response.json();
-        const error = (result as Names.Error).error;
-        if (error !== undefined) {
-          throw new Error(error);
+        try {
+          const response = await fetch(`${Names.baseUrl}${kind}`);
+          const data = await response.json();
+          return data.results as Names.Result;
+        } catch(error) {
+          throw new Error(`Failed to fetch "${kind}"`);
         }
-        return result as Names.Result;
       }, () => [this.kind]
     );
   }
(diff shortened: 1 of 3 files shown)


packages/lit-dev-content/samples/docs/controllers/names/names-controller.ts
```diff
@@ -16,13 +16,13 @@ export class NamesController {
...
-        const response = await fetch(`${Names.baseUrl}${Names.kindIdMap[kind]}`);
-        const result = await response.json();
-        const error = (result as Names.Error).error;
-        if (error !== undefined) {
-          throw new Error(error);
+        try {
+          const response = await fetch(`${Names.baseUrl}${kind}`);
+          const data = await response.json();
+          return data.results as Names.Result;
+        } catch(error) {
```
AndrewJakubowicz, line 23: Nit: `error` is not referenced. Replace with `} catch {`

augustjk: Ah, yes good point. I considered doing a `console.error(error);` before re-throwing a new error so we don't lose that info but I wasn't sure if it was necessary for this example. What do you think?

AndrewJakubowicz (approved):
Nice! Manually tested as well. Thanks for fixing this example. TIL a bunch more about Task as well!

packages/lit-dev-content/samples/docs/controllers/names/names-controller.ts
AndrewJakubowicz: I think it's fine removing since the error flows into the render. And not sure we expect or want to encourage samples logging into the console.

arthurevans (approved):
LGTM, thanks!

PR #856, opened 2022-06-08
Update decorators doc to stage 3, update babel config.

### LINKED ISSUE #754: Update the decorators documentation (Stage 3)
Finally the decorators have been moved to stage 3!

https://github.com/tc39/proposal-decorators/pull/458

Babel plugin configuration has changed:
https://babeljs.io/docs/en/babel-plugin-proposal-decorators#options

Current Lit documentation:
https://lit.dev/docs/components/decorators/

I think we should update it.

arthurevans:
Fixes #754.

I updated the note about Babel support (line #117). I tried it with this config, and if I understand correctly, this setting will give you the same behavior we've been testing against with previous versions of the Babel plugin, which seems like the safest option.

The "we recommend" here is my best guess as to what we want to recommend.

### DIFF
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -96,20 +96,17 @@ Note, the `@babel/plugin-proposal-class-properties` may not be required with the
 
 To set up the plugins, add code like this to your Babel configuration:
 
-```js
-assumptions = {
+```json
+"assumptions": {
   "setPublicClassFields": true
-};
-
-plugins = [
-  ['@babel/plugin-proposal-decorators',
-    {
-      version: '2018-09',
-      decoratorsBeforeExport: true
-    },
-  ],
-  ["@babel/plugin-proposal-class-properties"],
-];
+},
+"plugins": [
+  ["@babel/plugin-proposal-decorators", {
+    "version": "2018-09",
+    "decoratorsBeforeExport": true
+  }],
+  ["@babel/plugin-proposal-class-properties"]
+]
 ```
 
 <div class="alert alert-info">
@@ -122,20 +119,20 @@ Babel decorator support has been tested with `version: '2018-09'`. This is curre
 
 When using TypeScript with Babel, it's important to order the TypeScript transform before the decorators transform in your Babel config as follows:
 
-```js
+```json
 {
   "assumptions": {
     "setPublicClassFields": true
   },
-  "plugins":[
-    ["@babel/plugin-transform-typescript", {"allowDeclareFields": true}],
-    ['@babel/plugin-proposal-decorators',
-      {
-        version: '2018-09',
-        decoratorsBeforeExport: true
-      },
-    ],
-    ["@babel/plugin-proposal-class-properties"],
+  "plugins": [
+    ["@babel/plugin-transform-typescript", {
+      "allowDeclareFields": true
+    }],
+    ["@babel/plugin-proposal-decorators", {
+      "version": "2018-09",
+      "decoratorsBeforeExport": true
+    }],
+    ["@babel/plugin-proposal-class-properties"]
   ]
 }
 ```
--- a/packages/lit-dev-content/site/docs/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/components/decorators.md
@@ -10,9 +10,19 @@ versionLinks:
 
 Decorators are special functions that can modify the behavior of classes, class methods, and class fields. Lit uses decorators to provide declarative APIs for things like registering elements, reactive properties, and queries.
 
-Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard, which means they're neither finalized nor implemented in browsers yet. Compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for proposed features like decorators by compiling them into standard JavaScript a browser can run.
+Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard,  Currently no browsers implement decorators, but compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for an earlier version of the decorators proposal. Lit decorators work with Babel and TypeScript, and will be updated to work with the final specification when it's implemented in browsers.
 
 See the [Enabling decorators](#enabling-decorators) section for more information.
+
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -102,14 +102,19 @@ assumptions = {
...
 
 plugins = [
-  ['@babel/plugin-proposal-decorators', {decoratorsBeforeExport: true}],
+  ['@babel/plugin-proposal-decorators',
+    {
+      version: '2018-09',
+      decoratorsBeforeExport: true
+    },
+  ],
   ["@babel/plugin-proposal-class-properties"],
```
augustjk, line 111: Nit: in terms of consistency, this line has double quotes whereas the above has single quotes.

It would depend on whether the user uses json or js config file but maybe we should just use JSON syntax with double quotes for both key and value which will works for both.

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -124,7 +129,12 @@ When using TypeScript with Babel, it's important to order the TypeScript transfo
   },
   "plugins":[
     ["@babel/plugin-transform-typescript", {"allowDeclareFields": true}],
-    ["@babel/plugin-proposal-decorators", {"decoratorsBeforeExport": true}],
+    ['@babel/plugin-proposal-decorators',
```
augustjk, line 132: Nit: quote inconsistencies between this line and the ones around it.

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -102,14 +102,19 @@ assumptions = {
 };
 
 plugins = [
```
augustjk, line 104: Not exactly relevant to the PR but since we're already here, I think there should be a colon rather than an equal sign here and for the `assumptions` line above.

arthurevans: Hmm... Yeah, this is a good point. These things were copied from different sources and each of them uses a different format. I think to your point below, standardizing on something that would work in either JS or JSON is a solid choice....

augustjk (approved):
Nice!

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -10,7 +10,7 @@ versionLinks:
 
 Decorators are special functions that can modify the behavior of classes, class methods, and class fields. Lit uses decorators to provide declarative APIs for things like registering elements, reactive properties, and queries.
 
-Decorators are a [stage 2 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard, which means they're neither finalized nor implemented in browsers yet. Compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for proposed features like decorators by compiling them into standard JavaScript a browser can run.
+Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard, which means they're neither finalized nor implemented in browsers yet. Compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for proposed features like decorators by compiling them into standard JavaScript a browser can run.
```
rictic, line 13: I think a stage 3 proposal is considered finalized, save for any last minute issues that come up as browsers and compilers implement it.

augustjk: https://tc39.es/process-document/
At stage 3, the spec is considered "complete" but at stage 4 is when it's considered "final" so I think our wording here of "neither finalized nor implemented" is still valid.

justinfagnani: TC39's definition of "complete" and "final" are a bit different in practice than common understanding though. Stage 3 means that, barring drastic implementor feedback, there will not be any changes. It's basically final in 99% of cases, which is why this is the stage where implementors implement and ship.

sorvell (approved):
Confirmed that we do not support '2021-12'. Due to the churn on this and the difficulty of supporting multiple versions, we're electing to wait for browsers to move forward a bit on implementation before supporting this latest version.

packages/lit-dev-content/site/docs/components/decorators.md
justinfagnani, line 13: I think it might be worth noting that not only do browsers not implement decorators, the compilers implement an older version of the spec, but that we support both Babel and TypeScript and will support the final version when browsers do.

arthurevans: Done.

arthurevans: Updated text, added an aside to describe the process a little bit, because having all of this packed into one paragraph was getting clumsy.

packages/lit-dev-content/site/docs/components/decorators.md
```diff
@@ -10,10 +10,20 @@ versionLinks:
 
 Decorators are special functions that can modify the behavior of classes, class methods, and class fields. Lit uses decorators to provide declarative APIs for things like registering elements, reactive properties, and queries.
 
-Decorators are a [stage 2 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard, which means they're neither finalized nor implemented in browsers yet. Compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for proposed features like decorators by compiling them into standard JavaScript a browser can run.
+Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard,  Currently no browsers implement decorators, but compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for an earlier version of the decorators proposal. Lit decorators work with Babel and TypeScript, and will be updated to work with the final specification when it's implemented in browsers.
```
augustjk, line 13:
```suggestion
Decorators are a [stage 3 proposal](https://github.com/tc39/proposal-decorators) for addition to the ECMAScript standard. Currently no browsers implement decorators, but compilers like [Babel](https://babeljs.io/) and [TypeScript](https://www.typescriptlang.org/) provide support for an earlier version of the decorators proposal. Lit decorators work with Babel and TypeScript, and will be updated to work with the final specification when it's implemented in browsers.
```

arthurevans: Oops... Thanks for the fix!

justinfagnani (approved):

PR #858, opened 2022-06-16
update to playground-elements 0.16.2

taylor-vann:
Bump the `playground-element` dependency up a knotch!

This update supports `jsx` and `tsx` files.


aomarks:
The package locks need to be updated too, so you need to re-bootstrap.

PR #862, opened 2022-06-17
Convert lit.dev to npm workspaces

aomarks:
Replaces Lerna with npm workspaces.

### Issues encountered

- npm workspaces hoists shared dependencies, so there were a few spots where we were assuming that `playground-elements` would get installed to `packages/lit-dev-content/node_modules`. Updated these spots.

### Also

- Upgraded to the latest playground-elements
- Upgraded to the latest MWC.
- Upgraded to the latest Wireit.

- Moved the playground-elements dependency to the top-level, since it gets installed there anyway, and now there will only be one place to change the version constraint, instead of 3.

- Added some logging to the static playground renderer, so that we get more information when it times out (e.g. in this case there was a 404 that wasn't getting displayed in the build logs).

- Constrained our typescript version with a `~`, because there is a new breaking change I didn't want to address in this PR, and `~` is a good idea anyway because TypeScript doesn't follow semver.

- There seems a problem in the integration tests where running it a second time causes an error because the server is already running. This seems to be because playwright sends a `SIGKILL` instead of a `SIGTERM` or `SIGINT` to shut down the server, which I think causes the server to not shut down correctly (perhaps because it is a child of `npm` and `run-p`), so it keeps running after an integration test (https://github.com/google/wireit/issues/33). This means that we can't get updated goldens after a screenshot failure, because that works by running the integration tests again with a special option enabled. For now I have enabled the `reuseExistingServer` option as a workaround. Wireit services will also solve this: https://github.com/google/wireit/issues/33.

-  Split apart check links workflow from integration tests workflow. Just seems nicer to view them separately.


package.json
```diff
@@ -5,22 +5,24 @@
...
-    "bootstrap": "lerna bootstrap --ci",
     "build": "wireit",
     "start": "run-p -r start:main start:playground start:fake-github",
-    "start:main": "LITDEV_ENV=local MODE=main lerna run start --scope=lit-dev-server --stream",
-    "start:playground": "LITDEV_ENV=local MODE=playground lerna run start --scope=lit-dev-server --stream",
+    "start:main": "LITDEV_ENV=local MODE=main npm run start -w lit-dev-server",
+    "start:playground": "LITDEV_ENV=local MODE=playground npm start -w lit-dev-server",
     "start:fake-github": "wireit",
-    "dev": "cd packages/lit-dev-server && npm run build && cd ../lit-dev-content && (npm run dev:build:assets watch & npm run dev:build:eleventy:watch & node ../lit-dev-tools-esm/lib/dev-server.js --open)",
+    "dev": "cd lit-dev-server && npm run build && cd ../lit-dev-content && (npm run dev:build:assets --watch & npm run dev:build:eleventy:watch & node ../lit-dev-tools-esm/lib/dev-server.js --open)",
```
augustjk, line 17: The `cd` here doesn't work.
```suggestion
    "dev": "cd packages/lit-dev-server && npm run build && cd ../lit-dev-content && (npm run dev:build:assets --watch & npm run dev:build:eleventy:watch & node ../lit-dev-tools-esm/lib/dev-server.js --open)",
```

augustjk (approved):
Very nice! One minor thing for the `dev` script.

package.json
aomarks: Oops, fixed

AndrewJakubowicz (commented):
Amazing!

Preview URL testing looks great!
In the preview url checked home page, a tutorial, playground examples, and blog.

Having some trouble with local tests.

  - `npm run dev` playground preview is not loading.

aomarks:
> Having some trouble with local tests.
> 
> * `npm run dev` playground preview is not loading.

Should be fixed with latest push

AndrewJakubowicz (approved):
Working great! Thank you!

package.json
```diff
@@ -5,22 +5,24 @@
...
     "format": "prettier . --write",
     "format:check": "prettier . --check",
-    "nuke": "rm -rf node_modules && npm ci && lerna exec 'rm -rf node_modules' && lerna bootstrap --ci",
-    "upgrade": "rm -rf node_modules package-lock.json && npm install && lerna exec 'rm -rf node_modules package-lock.json' && lerna bootstrap",
-    "test:unit": "lerna run test --scope lit-dev-tools-cjs --scope lit-dev-tools-esm --stream",
-    "test:integration": "lerna run test:integration --stream",
-    "test:integration:update-golden-screenshots": "lerna run test:integration:update-golden-screenshots --stream",
-    "test:links": "lerna run test:links --stream"
+    "nuke": "rm -rf node_modules packages/*/node_modules && npm ci",
```
justinfagnani, line 20: Is this command necessary anymore? I thought `npm ci` was supposed to put node_modules in exactly the state described by package-lock.json

packages/lit-dev-tools-cjs/src/playground-plugin/renderer.ts
```diff
@@ -168,7 +185,9 @@ class RendererServer {
       process.stdout.write = (() => {}) as any;
       const wds = await startDevServer({
         config: {
-          rootDir: pathlib.resolve(__dirname, '..', '..'),
+          // playground-elements is hoisted to the root node_modules by
```
justinfagnani, line 188: I think ideally we should be doing a node resolve here in case the location changes again.

packages/lit-dev-tools-esm/src/dev-server.ts
```diff
@@ -16,7 +16,8 @@ import {createGitHubTokenExchangeMiddleware} from 'lit-dev-server/lib/middleware
 import {dev as ENV} from 'lit-dev-tools-cjs/lib/lit-dev-environments.js';
 
 const THIS_DIR = pathlib.dirname(fileURLToPath(import.meta.url));
-const CONTENT_PKG = pathlib.resolve(THIS_DIR, '..', '..', 'lit-dev-content');
+const REPO_ROOT = pathlib.resolve(THIS_DIR, '..', '..', '..');
+const CONTENT_PKG = pathlib.resolve(REPO_ROOT, 'packages', 'lit-dev-content');
```
justinfagnani, line 20: node resolve would be good here too

justinfagnani (commented):
Just a couple of post-merge comments...

packages/lit-dev-tools-cjs/src/playground-plugin/renderer.ts
aomarks: Can't do it here actually, because we're specifying the root directory to serve from. Then we hard code a path into the node_modules folder. Maybe could be done smarter, but requires more refactoring.

packages/lit-dev-tools-esm/src/dev-server.ts
aomarks: Done in https://github.com/lit/lit.dev/pull/863

package.json
aomarks: Done in https://github.com/lit/lit.dev/pull/863

PR #876, opened 2022-07-03
fix a link  error

novenn:
i think this is not a right link.


arthurevans (approved):
Right you are. Thanks for the fix!

PR #879, opened 2022-07-11
Fix snippet in Conditionals docs

deanhaleem:
Added missing closing template string and parenthesis.


arthurevans (approved):
Thanks for the fix!

PR #882, opened 2022-07-15
Lit for polymer users article

arthurevans:
Please review for **content**.

We're waiting on an actual article format, this includes minimal changes to add an article type with TOC (top only ATM) and the param list format from the docs pages. This is just for review purposes, and it leaves a number of weird visual glitches (like permalink anchors that are on a separate line, some weird formatting on the top TOC). So please ignore CSS glitches as we'll be throwing that stuff out. (Markdown formatting errors are fair game, though.)

### DIFF
--- a/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
+++ b/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
@@ -89,7 +89,7 @@ If your application doesn't use two-way binding, you can skip this section.
 
 For two-way bindings, Polymer uses its own protocol, which has three main components:
 
-*   A one-way binding from host to child.
+*   A binding from host to child.
 
 *   An event listener that handles property change events from the child element.
 
@@ -101,7 +101,7 @@ Ideally an event is a discrete signal sent to communicate an explicit change tha
 
 To summarize the [Custom Element Best Practices](https://web.dev/custom-elements-best-practices/#events) guidelines, a component should* fire events:
 
-*   When a property changes as a result of user interaction—like clicking a button or editing a text field inside the component.
+*   When the state of the element changes as a result of user interaction—like clicking a button or editing a text field inside the component.
 *   When something internal changes inside the component—like a timer going off or an animation completing.
 
 Ideally, a component should fire _semantic_ events, which describe what changed rather than letting low-level UI events bubble out. For example, a form that lets the user update profile data might fire a `profile-updated` event when the user clicks the **Done** button. The `profile-updated` event is relevant to the parent: the click event isn't.
@@ -139,7 +139,7 @@ static get template() {
 }
 ```
 
-This component uses two-way binding to the `input` element, and has a notifying property, so a parent element can use two-way binding with the `name` property. The two-way binding to the input isn't terrible: it's only updated when the `input` event fires. However, by adding an event listener for the `input` event, you can eliminate the notifying property:
+This component uses two-way binding to the `input` element, and has a notifying property, so a parent element can use two-way binding with the `name` property. The two-way binding to the input makes sense since the binding is set only when the `input` event fires. However, by adding an event listener for the `input` event, you can eliminate the notifying property:
 
 ```js
 static properties = {
@@ -166,7 +166,7 @@ inputChanged(e) {
 
 This code explicitly listens for input events and fires `name-changed` events. This code will work with a parent that expects the Polymer two-way binding protocol, so you can update components one at a time.
 
-This could won't fire a property change event when the parent sets the `name` property—which is a good thing. And you can migrate this code fairly directly to Lit.
+This code won't fire a property change event when the parent sets the `name` property—which is a good thing. And you can migrate this code fairly directly to Lit.
 
 Using an event handler like this, you could also add logic to the child component—such as only firing the `name-changed` event when the user stops typing for a certain interval.
 
--- a/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
+++ b/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
@@ -109,9 +109,7 @@ Ideally, a component should fire _semantic_ events, which describe what changed 
 
 #### Replacing two-way bindings
 
-There are many ways to replace two-way bindings. On the more complex side, you can use a state container like Redux. Individual components can subscribe for updates to the state, and dispatch actions to update the state. This is a great way to go, but it may require a lot of work.
-
-If you're just looking to migrate an existing Polymer application to Lit, you may just want to replace the two-way bindings with code that serves a similar function. There are two steps to this:
+There are many ways to replace two-way bindings. If you're just looking to migrate an existing Polymer application to Lit, you may just want to replace the two-way bindings with code that serves a similar function. There are two steps to this:
 
 *   Replace the automatic property change event with a more intentional change event.
 
@@ -169,6 +167,14 @@ This code explicitly listens for input events and fires `name-changed` events. T
 This code won't fire a property change event when the parent sets the `name` property—which is a good thing. And you can migrate this code fairly directly to Lit.
 
 Using an event handler like this, you could also add logic to the child component—such as only firing the `name-changed` event when the user stops typing for a certain interval.
+
+{% aside "info" %}
+
+One-way data flow.
+
+Another alternative is to replace two-way bindings with a unidirectional data flow pattern, using a state container like Redux. Individual components can subscribe for updates to the state, and dispatch actions to update the state. We recommend this for new development, but it may require more work if you already have an application based on two-way bindings.
+
+{% endaside %}
 
 ## Polymer to Lit cookbook
 
--- a/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
+++ b/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
@@ -14,17 +14,17 @@ This document provides a quick overview of how Lit relates to Polymer, and provi
 
 ## How does Lit relate to Polymer?
 
-The Polymer library was one of the first libraries for building web components. Lit is a successor to Polymer, built by the same team and with many of the same aims. The projects share many goals, but Lit takes advantage of lessons learned during the development of the Polymer library.
+Polymer was one of the first libraries for building web components. Lit is a successor to Polymer, built by the same team and with many of the same aims. The projects share many goals, but Lit takes advantage of lessons learned during the development of Polymer.
 
-Because the  Polymer library is the predecessor to Lit, there are a lot of similarities between the two. Both libraries make it easy to build components that work like built-in HTML elements, and both feature declarative HTML templates.
+Because Polymer is the predecessor to Lit, there are a lot of similarities between the two. Both libraries make it easy to build components that work like built-in HTML elements, and both feature declarative HTML templates.
 
-Lit differs from the Polymer library in several ways:
+Lit differs from Polymer in several ways:
 
 *   **Lit's rendering is asynchronous and batched by default.** With a few exceptions, all Polymer updates are synchronous.
 
 *   **Lit exposes an update lifecycle that provides a powerful mechanism for observing changes to properties and computing derived values from them.** Polymer has declarative observers and computed properties, but it can be hard to predict the order in which observers will run.
 
-*   **Lit focuses on JavaScript-first authoring, using native JavaScript modules.** The Polymer library _originally_ focused on HTML-first authoring, made possible by the HTML Imports specification, which has since been removed from the web platform.
+*   **Lit focuses on JavaScript-first authoring, using native JavaScript modules.** Polymer  _originally_ focused on HTML-first authoring, made possible by the HTML Imports specification, which has since been removed from the web platform.
 
 *   **Lit expressions use standard JavaScript.**  Polymer uses a limited domain-specific language in its bindings. Because Lit uses standard JavaScript, you can also use JavaScript for control flow inside expressions (conditional templates and repeated templates), where Polymer uses specialized helper elements.
 
@@ -585,7 +585,7 @@ import {cache} from 'lit/directives/cache.js';
    )}`;
 ```
 
-Since this won't render anything when the condition is falsy, you can use it to avoid creating a complex piece of DOM on initial page load. 
+Since this won't render anything when the condition is falsy, you can use it to avoid creating a complex piece of DOM on initial page load.
 
 #### Repeating templates
 
--- a/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
+++ b/packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
@@ -292,7 +292,7 @@ static get template() {
 }
 ```
 
-In Lit, you provide styles in a static `styles` field using the `css` tag function.
+In Lit, you _typically_ provide styles in a static `styles` field using the `css` tag function.
 
 ```js
 import {LitElement, css, html} from 'lit';
@@ -300,6 +300,22 @@ import {LitElement, css, html} from 'lit';
 
   static styles = css`.fancy { color: blue; }`;
 ```
+
+Adding a style tag directly in the template, like you would in Polymer, is also supported:
+
+```js
+import {LitElement, html} from 'lit';
+  ...
+  render() {
+    return html`
+      <style>
+        .fancy { color: blue; }
+      </style>
+      ...
+  }
+```
+
+Using a style tag may be slightly less performant than the static `styles` field, because the styles are evaluated once per instance instead of once per class.
 
 For more information, see [Styles](/docs/components/styles/).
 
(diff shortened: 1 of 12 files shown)


packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+tags: blogPosts
+layout: article.html
+title: "Lit for Polymer users"
+summary: "Migrating Polymer apps to Lit, or learning Lit as a Polymer user"
+date: 2022-07-19
+author:
+  - arthur-evans
+---
+
+Lit is a successor to the Polymer library. If you have a project built with Polymer and want to migrate it to Lit, or if you're familiar with Polymer and want to know how it compares to Lit, this is the right document for you.
```
aomarks, line 11: Since we're no longer using the "Polymer project" brand, and have disassociated Lit from it, would it make sense to start referring to Polymer simply as "Polymer" again, instead of "the Polymer library"? It doesn't really feel like the "Polymer library" name caught on in our community.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+
+For two-way bindings, Polymer uses its own protocol, which has three main components:
+
+*   A one-way binding from host to child.
+
+*   An event listener that handles property change events from the child element.
+
+*   A facility for the child to automatically fire change events when properties change (`notify: true`).
+
+This last item is the most problematic. Components fire change events for any change to a property, and each change event is handled synchronously.  Widespread use of two-way binding across an entire application can make it hard to reason about data flow and the order in which components update themselves. A change can propagate both up and down the tree, potentially leading to components being updated in an unexpected order.
```
sorvell, line 98: Maybe add something like...
Ideally an event is a discrete signal intentionally sent to communicate an explicit change that isn't otherwise easily observable. Sending an event as a side-effect of setting a property as Polymer does makes the communication potentially redundant and implicit. This implicit behavior, in partcular, can make data flow hard to understand.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+  }
+}
+```
+
+
+#### Array & object type properties
+
+Like Polymer, Lit does dirty checking when properties change to avoid performing unnecessary work. This can lead to issues if you have a property that holds an object or array. If you mutate the object or array, Lit won't detect a change.
+
+In most cases, the best way to avoid these issues is to use immutable data patterns, so that you always assign a new object or array value instead of mutating an existing object or array.
```
sorvell, line 741: The rules for Polymer and Lit are the same if you don't include `this.set` so I think that needs to be mentioned here.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+  return html`Text: ${cyphertext}`;
+}
+```
+
+Also, because Lit allows you to use any JavaScript expression in a template, Polymer's [computed bindings](https://polymer-library.polymer-project.org/3.0/docs/devguide/data-binding#annotated-computed) can be replaced with inline expressions, including function calls.
+
+Polymer:
+
+```js
+  static get template() {
```
sorvell, line 864: Include:

```
_formatName(given, family) {
  return `${given} ${family}`;
}
```

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+  }
+
+```
+
+Lit:
+
+```js
+render() {
+  html`
+      My name is <span>${this._formatName(given, family)}</span>
```
sorvell, line 877: Then this is sooo much better.

```suggestion
      My name is <span>${this.given} ${this.family}</span>
```

sorvell (changes requested):
This is really comprehensive. Great job! Suggestions are mostly minor.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
arthurevans: Added this with a few edits. Cut some text that was less on-point.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+*   When something internal changes inside the component—like a timer going off or an animation completing.
+
+Ideally, a component should fire _semantic_ events, which describe what changed rather than letting low-level UI events bubble out. For example, a form that lets the user update profile data might fire a `profile-updated` event when the user clicks the **Done** button. The `profile-updated` event is relevant to the parent: the click event isn't.
+
+
+#### Replacing two-way bindings
+
+There are many ways to replace two-way bindings. On the more complex side, you can use a state container like Redux. Individual components can subscribe for updates to the state, and dispatch actions to update the state. This is a great way to go, but it may require a lot of work.
+
+If you're just looking to migrate an existing Polymer application to Lit, you may just want to replace the two-way bindings with code that serves a similar function. There are two steps to this:
```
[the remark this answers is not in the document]
arthurevans: Right, that's what we're doing in this procedure. I'm not clear on whether you want to make the intro text more explicit, or if you're commenting on the procedure below.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+
+*   When a property changes as a result of user interaction—like clicking a button or editing a text field inside the component.
+*   When something internal changes inside the component—like a timer going off or an animation completing.
+
+Ideally, a component should fire _semantic_ events, which describe what changed rather than letting low-level UI events bubble out. For example, a form that lets the user update profile data might fire a `profile-updated` event when the user clicks the **Done** button. The `profile-updated` event is relevant to the parent: the click event isn't.
+
+
+#### Replacing two-way bindings
+
+There are many ways to replace two-way bindings. On the more complex side, you can use a state container like Redux. Individual components can subscribe for updates to the state, and dispatch actions to update the state. This is a great way to go, but it may require a lot of work.
```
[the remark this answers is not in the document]
arthurevans: Hmm... If I just put the redux comment at the end it's going to be totally lost, and it's not enough text to be its own section. Maybe I can add it as an aside near the end.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+
+*   A one-way binding from host to child.
+
+*   An event listener that handles property change events from the child element.
+
+*   A facility for the child to automatically fire change events when properties change (`notify: true`).
+
+This last item is the most problematic. Components fire change events for any change to a property, and each change event is handled synchronously.  Widespread use of two-way binding across an entire application can make it hard to reason about data flow and the order in which components update themselves. A change can propagate both up and down the tree, potentially leading to components being updated in an unexpected order.
+
+In general the Lit team recommends using top-down data flow using immutable data patterns, which make it easier to reason about how a data change propagates. To avoid issues, components should *not* generate an event when a property is set from above (that is, set programmatically from outside the component). A component should generate events:
```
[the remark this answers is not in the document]
arthurevans: Done. Reordered this as an intro to the bulleted list, to make it clear that we're basically summarizing the best practices.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+<target-element attribute-binding$=${value}></target-element>
+
+<!-- boolean attribute binding adds a ? prefix -->
+<target-element ?hidden=${value}></target-element>
+
+<!-- event handler uses @event-name syntax -->
+<target-element @click=${handler}></target-element>
+```
+
+In Lit, the handler can be either an unbound method, like `${this.clickHandler}` or an arrow function. Using an arrow function, you can close over other data or call a function with a different signature. For example:
```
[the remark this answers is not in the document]
arthurevans: Removed "unbound" (also changed "an" => "a" so doing the change manually, not committing this suggestion).

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+
+#### Configuring properties
+
+Like Polymer, Lit lets you configure properties using a static `properties` field.
+
+Lit also supports using the `@property` decorator to declare a reactive property. For more information, see [About decorators](#about-decorators).
+
+Polymer:
+
+```js
```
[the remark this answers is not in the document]
arthurevans: Fixed missing code fence.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
arthurevans: Mention `this.set`? Or mention that the rules are the same?

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
arthurevans: In this case both Polymer and Lit could simply insert two strings. Calling a function is more interesting because, for example, formatting the name might be locale-specific. And it's a very easy pattern to translate.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+  return html`
+    <style>
+      .fancy { color: blue; }
+    </style>
+    ...
+  `;
+}
+```
+
+In Lit, you provide styles in a static `styles` field using the `css` tag function.
```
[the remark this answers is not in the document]
arthurevans: Noted.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+render() {
+  let count = this.getMatchingItems(this.pattern);
+
+  return html`You have ${count} matches.`
+}
+```
+
+Like Polymer, Lit supports setting properties, attributes, and event handlers using expressions. Lit uses slightly different syntax, with prefixes instead of suffixes.
+
+Polymer:
```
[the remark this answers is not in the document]
arthurevans: Done. Ideally we'd make the table horizontally scrollable on small screens, because it's a little awkward there, but it does work.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+
+  static styles = css`.fancy { color: blue; }`;
+```
+
+For more information, see [Styles](/docs/components/styles/).
+
+
+### Data binding
+
+Where Polymer has data binding, Lit has _expressions_ in its templates. Lit expressions are standard JavaScript expressions that are bound to a particular location in the template. As such, Lit expressions can do almost everything you can do with Polymer data bindings, and many things that you can't easily do in Polymer.
```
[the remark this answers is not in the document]
arthurevans: Going to leave this one for now. Felt like the alternative was a little bit less clear than the original.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+```html
+<template is="dom-if" if="{{condition}}">
+  <div>condition is true</div>
+</template>
+```
+
+In Lit, you can use a JavaScript conditional expressions. The conditional operator (or ternary) works well:
+
+```js
+html`${this.condition
```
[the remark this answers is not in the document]
arthurevans: I thought we'd decided we should use `map` by default from now on, because it has some advantages over array.map (namely, being able to use other, non-array iterables).

Whereas `when` doesn't seem to add anything other than another import statement.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
arthurevans: Mentioned rules are basically the same, added a paragraph mentioning `this.set` and friends (which will probably need to be replaced with immutable patterns when moving to Lit).

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,935 @@
...
+Unlike the Polymer `dom-if`, the conditional operator lets you supply content for both true and false conditions, although you can also return the empty string to render nothing, as in the example.
+
+For more information, see [Conditionals](https://lit.dev/docs/templates/conditionals/).
+
+
+#### The restamp property
+
+By default, Polymer's `dom-if` behaves a little differently from a Lit conditional. When the condition goes from a truthy to a falsy value, the `dom-if` simply _hides_ the conditional DOM, instead of removing it from the DOM tree. This may save some resources when the condition becomes truthy again.
+
+On the other hand, Lit removes and discards the DOM when a condition changes; `dom-if` does the same thing if you set the `restamp` property to `true`.
```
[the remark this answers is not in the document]
arthurevans: Added (and rewrote section a good bit to try and explain the differences between `hidden=${condition}`, a simple JavaScript conditional, and the `cache` directive.)

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
arthurevans: I'm not making these two changes because this section is specifically showing how to use an inline function.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
arthurevans: I'm not sure about dropping it altogether; but for the sake of brevity I use "Polymer" in most of this article. I updated the text to be more consistent, and only use the full name on the first reference.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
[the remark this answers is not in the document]
arthurevans: Done.

arthurevans:
Addressed feedback, I think.

Once this doc looks GTG, I'll close this PR and open a new one based on top of Elliott's new article PR. I've tested this out, and it looks pretty good.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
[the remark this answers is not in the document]
arthurevans: Made text more explicit.

arthurevans:
Ping @sorvell

arthurevans:
Now replaced by #909. 

Leaving this as a draft for now in case anyone wants to check the comments/diffs.

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,1013 @@
...
+
+*   A binding from host to child.
+
+*   An event listener that handles property change events from the child element.
+
+*   A facility for the child to automatically fire change events when properties change (`notify: true`).
+
+This last item is the most problematic. Components fire change events for any change to a property, and each change event is handled synchronously.  Widespread use of two-way binding across an entire application can make it hard to reason about data flow and the order in which components update themselves.
+
+Ideally an event is a discrete signal sent to communicate an explicit change that isn't otherwise easily observable. Sending an event as a side-effect of setting a property as Polymer makes the communication potentially redundant and implicit. This implicit behavior, in particular, can make data flow hard to understand.
```
sorvell, line 100:
```suggestion
Ideally an event is a discrete signal sent to communicate an explicit change that isn't otherwise easily observable. Sending an event as a side-effect of setting a property as Polymer does makes the communication potentially redundant and implicit. This implicit behavior, in particular, can make data flow hard to understand.
```

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,1013 @@
...
+
+To summarize the [Custom Element Best Practices](https://web.dev/custom-elements-best-practices/#events) guidelines, a component should* fire events:
```
sorvell, line 102:
```suggestion
To summarize the [Custom Element Best Practices](https://web.dev/custom-elements-best-practices/#events) guidelines, a component *should* fire events:
```

packages/lit-dev-content/site/blog/2022-07-19-lit-for-polymer-users.md
```diff
@@ -0,0 +1,1013 @@
...
+
+*   When the state of the element changes as a result of user interaction—like clicking a button or editing a text field inside the component.
+*   When something internal changes inside the component—like a timer going off or an animation completing.
```
sorvell, line 105:
```suggestion

*   When something internal changes inside the component—like a timer going off or an animation completing.
```

sorvell (approved):
Just a couple of very small typo fixes but otherwise looks good!

arthurevans:
Cool! I'll make those changes on #909.

arthurevans:
Fixed in https://github.com/lit/lit.dev/pull/909/commits/c3398eae897a7f0256e362fbda309474de6aa299

PR #889, opened 2022-08-02
Update playground-elements

taylor-vann:
Update playground-elements to `0.16.3` to include JSX and TSX syntax highlighting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -196,30 +196,39 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.16.7",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
-      "integrity": "sha512-iAXqUn8IIeBTNd72xsFlgaXHkMBMt6y4HJp1tIaK465CWLT/fG1aqB7ykr95gHHmlBdGbFeWWfyB4NJJ0nmeIg==",
+      "version": "7.18.6",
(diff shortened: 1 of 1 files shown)


AndrewJakubowicz:
The build and preview deploy is broken.
It looks like the `SourceFile` type between `@lit/ts-transformers` and `typescript` is incompatible:
https://github.com/lit/lit.dev/runs/7640278509?check_suite_focus=true#step:9:128

Is it possible two different incompatible typescript versions are being used?

aomarks:
Looks like this probably upgrades everything? Probably some new incompatibility introduced, could be totally unrelated to playground-elements. I would suggest reverting to the previous `package-lock.json`, and then bumping the playground-elements version in the `package.json`, and then running `npm i`. That should only upgrade playground-elements.

taylor-vann:
it is totally possible!
im gonna remove my my changes and try again @AndrewJakubowicz

aomarks:
> Looks like this probably upgrades everything? Probably some new incompatibility introduced, could be totally unrelated to playground-elements. I would suggest reverting to the previous `package-lock.json`, and then bumping the playground-elements version in the `package.json`, and then running `npm i`. That should only upgrade playground-elements.

I think you could also try `npm update playground-elements`

taylor-vann:
nice! the "just updating playground elements" seems to have worked well

aomarks (approved):
Confirmed JSX and TSX highlighting is happening in the latest preview. Nice!

AndrewJakubowicz (approved):
Nice!

taylor-vann:
Thanks @aomarks for all the help on this one!

PR #891, opened 2022-08-05
implement articles MVP

e111077:
This is a PR in a chain:

| PR 1 | PR 2 | PR 3 |
| -- | -- | --|
| #901 | #900 | This one |
| Move Authors | Filters & Styles | Articles implementation |

This PR implements the following views:

- articles
  - article-nav
  - article-feed
  - article

It also implements a LHS article navigation, and article tag view, but they are currently hidden and not click-navigable until the team decides when or if we will show them to users. (Marked in CSS with TODO comments)

## Good URLs to test

- All articles feed
  - https://pr891-33b4c5d---lit-dev-5ftespv5na-uc.a.run.app/articles/
- tag feed (for now will look the same as all articles feed)
  - https://pr891-33b4c5d---lit-dev-5ftespv5na-uc.a.run.app/articles/tags/composition/
- article view
  - https://pr891-33b4c5d---lit-dev-5ftespv5na-uc.a.run.app/articles/light-dom/

### DIFF
--- a/packages/lit-dev-content/site/_includes/articles.html
+++ b/packages/lit-dev-content/site/_includes/articles.html
@@ -1,0 +1,105 @@
+{% extends 'default.html' %}
+
+<!-- When we're rendering API docs, our "versionLinks" data doesn't come from
+     markdown frontmatter, instead it comes from the API JSON data.
+     TODO(aomarks) There must be some way to assign "versionLinks" to
+     "data.otherVersions" from api.html, but I haven't figure out how! -->
+{% if not versionLinks %}
+  {% set versionLinks = data.versionLinks %}
+{% endif %}
+
+{% block head %}
+  {% inlinecss "docs.css" %}
+  {% inlinecss "articles.css" %}
+  {% inlinejs "pages/docs.js" %}
+  <script type="module" src="{{ site.baseurl }}/js/components/playground-elements.js"></script>
+  <script type="module" src="{{ site.baseurl }}/js/components/litdev-example.js"></script>
+  <script type="module" src="{{ site.baseurl }}/js/components/litdev-switchable-sample.js"></script>
+{% endblock %}
+
+{% block content %}
+  {% set toc = content | toc %}
+
+  <div id="docsNavWrapper" class="minimalScroller">
+    <nav id="docsNav">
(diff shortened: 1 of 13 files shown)


packages/lit-dev-content/site/css/articles.css
```diff
@@ -0,0 +1,298 @@
+main {
+  /* Colors */
+  --docs-h1-underline-color: var(--color-blue);
```
aomarks, line 3: Seems like a lot is copy-pasted from docs.css.

It would also be nice to factor out the common styles.

Also, is everything here actually needed? E.g. I don't see use of the next/previous buttons in the example article, but the CSS is copied  here.

packages/lit-dev-content/site/_includes/articles.html
```diff
@@ -0,0 +1,105 @@
...
+<!-- When we're rendering API docs, our "versionLinks" data doesn't come from
+     markdown frontmatter, instead it comes from the API JSON data.
+     TODO(aomarks) There must be some way to assign "versionLinks" to
+     "data.otherVersions" from api.html, but I haven't figure out how! -->
+{% if not versionLinks %}
+  {% set versionLinks = data.versionLinks %}
+{% endif %}
+
+{% block head %}
+  {% inlinecss "docs.css" %}
```
aomarks, line 12: What about an RSS feed for articles?

Could we just include articles in the existing RSS feed for the blog?

packages/lit-dev-content/site/css/articles.css
e111077: Next and previous buttons are there but hidden. docs.css is actually loaded. articles.css is just some stuff on top of docs.css.

Though it was meant to start out as its own thing there is some cruft. Let me take another pass at it

packages/lit-dev-content/site/_includes/articles.html
e111077: Gray says we should only the blog should have RSS

aomarks (approved):

packages/lit-dev-content/site/_includes/articles.html
e111077: We've decided to table this until we get a concensus from the team

PR #895, opened 2022-08-08
Docs should refer to the minified template polyfill.

bicknellr:
The unminified version of the template polyfill is not compiled. For compatibility, users should point to the minified version in their apps.


justinfagnani (approved):

arthurevans (approved):
LGTM

PR #906, opened 2022-08-16
Pass corrected watch argument to Wireit

nkabrown:
Resolve Wireit 0.6.0 breaking change.
https://github.com/google/wireit/blob/main/CHANGELOG.md#060---2022-06-15


aomarks (approved):
Thanks!

Issue #565, opened 2021-10-15
Update documentation about what various falsy values render


nkabrown:
Documentation for [Rendering nothing](https://lit.dev/docs/templates/conditionals/#rendering-nothing) in the Conditionals section and the comments for `unsafeHTML` and `unsafeSVG` need to be updated to reflect the fact that `undefined`, `null`, and the empty string no longer render an empty text node but similar to `nothing` do not render a node.

See [issue 1559](https://github.com/lit/lit/issues/1559) in the lit repository and various comments like "Rendering the value `nothing`, `null`, or `undefined` in the `unsafeHTML` or `unsafeSVG` directive will now result in no content being rendered (previously it would render `'[object Object]'`, `'null'`, or `'undefined'`, respectively)." from the upgrade guide and "In child expressions, `undefined`, `null`, `''`, and `nothing` all behave the same and render no nodes." from [lit-html lines 351-352](https://github.com/lit/lit/blob/main/packages/lit-html/src/lit-html.ts#L351-L352).

sorvell:
Quick update on this:
* We're planning to clarify these docs and add info about using `nothing` either in a child or attribute expression.
* We're electing to explain `nothing` more explicitly in favor of offering more directives like `ifDefined`. In particular, there's a lot of flexibility in just doing: `value ?? nothing` or `value || nothing`.

PR #908, opened 2022-08-17
Clarify docs for rendering nothing in child and attribute expressions

sorvell:
Fixes https://github.com/lit/lit.dev/issues/565

### DIFF
--- a/packages/lit-dev-content/site/docs/components/shadow-dom.md
+++ b/packages/lit-dev-content/site/docs/components/shadow-dom.md
@@ -159,7 +159,7 @@ To assign a child to a specific slot, ensure that the child's `slot` attribute m
 
 {% playground-ide "docs/components/shadowdom/namedslots/" %}
 
-### Specifying slot fallback content
+### Specifying slot fallback content {#fallback}
 
 You can specify fallback content for a slot. The fallback content is shown when no child is assigned to the slot.
 
@@ -169,7 +169,7 @@ You can specify fallback content for a slot. The fallback content is shown when 
 
 <div class="alert alert-info">
 
-**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`.
+**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`. When using a Lit expression as a child of an element with Shadow DOM and a slot with fallback content, make sure to use a non-rendering value when appropriate so that fallback content is rendered. See [non-rendering values](/docs/templates/expressions/#non-rendering) for more information.
 
 </div>
 
--- a/packages/lit-dev-content/site/docs/components/shadow-dom.md
+++ b/packages/lit-dev-content/site/docs/components/shadow-dom.md
@@ -169,7 +169,7 @@ You can specify fallback content for a slot. The fallback content is shown when 
 
 <div class="alert alert-info">
 
-**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`. When using a Lit expression as a child of an element with Shadow DOM and a slot with fallback content, make sure to use a non-rendering value when appropriate so that fallback content is rendered. See [non-rendering values](/docs/templates/expressions/#non-rendering) for more information.
+**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`. When using a Lit expression as a child of a custom element, make sure to use a non-rendering value when appropriate so that any slot fallback content is rendered. See [non-rendering values](/docs/templates/expressions/#non-rendering) for more information.
 
 </div>
 
--- a/packages/lit-dev-content/site/docs/templates/expressions.md
+++ b/packages/lit-dev-content/site/docs/templates/expressions.md
@@ -138,24 +138,30 @@ html`<main>${bodyText}</main>`
 Expressions in the child position can take many kinds of values:
 
 * Primitive values
-* Non-rendering values
+* Sentinel values
 * `TemplateResult` objects created with the `html` function.
 * DOM nodes
 * Arrays or iterables of any of the supported types
 
 ### Primitive values
 
-Strings, numbers, and booleans are all converted to strings when interpolated into text content.
+Lit can render almost all (primitive values)[https://developer.mozilla.org/en-US/docs/Glossary/Primitive] and converts them to strings when interpolated into text content.
 
-A value of `5` will render the string `'5'`. The empty string `''` is specially treated as a [non-rendering value](#non-rendering).
+Numbers values like `5` will render the string `'5'`. Bigints are treated similarly.
 
-A boolean value `true` will render `'true'`, and `false` will render `'false'`. Rendering a boolean like this is uncommon. Instead booleans are typically used in conditionals to render other appropriate values. For more on conditionals, see [Conditionals](/docs/templates/conditionals/).
+A boolean value `true` will render `'true'`, and `false` will render `'false'`, but rendering a boolean like this is uncommon. Instead booleans are typically used in conditionals to render other appropriate values. For more on conditionals, see [Conditionals](/docs/templates/conditionals/).
 
-### Non-rendering values {#non-rendering}
+The empty string `''`, `null`, and `undefined` are specially treated and render nothing. See [Removing child content](#removing-child) for more information.
 
-The values `null`, `undefined`, the empty string `''`, and Lit's [nothing](/docs/api/templates/#nothing) sentinel value will all render no node when in a child expression.
+Symbol values cannot be converted to strings and throw when placed in child expressions.
 
-Rendering no node can be important when an expression is a child of an element with Shadow DOM that includes a `slot` with fallback content. Rendering no node ensures the fallback content is rendered. See [fallback content](/docs/components/shadow-dom/#fallback) for more information.
+### Sentinel values
+
+Lit supplies a couple of special sentinel values that can be used in child expressions.
+
+The `noChange` sentinel value does not change the expression's existing value. It is typically used in custom directives. See [Signaling no change](/docs/templates/custom-directives/#signaling-no-change) for more information.
+
+The `nothing` sentinel renders nothing. See [Removing child content](#removing-child) for more information.
 
 ### Templates
 
@@ -196,9 +202,17 @@ const page = html`
 `;
 ```
 
-### Arrays and iterables
+### Arrays or iterables of any of the supported types
 
 An expression can also return an array or iterable of any of the supported types, in any combination. You can use this feature along with standard JavaScript like the Array `map` method to create repeating templates and lists. For examples, see [Lists](/docs/templates/lists/).
+
+### Removing child content {#removing-child}
+
+The values `null`, `undefined`, the empty string `''`, and Lit's [nothing](/docs/api/templates/#nothing) sentinel value remove any previously rendered content and render no node.
+
+Setting or removing child content is often done based on a condition. See [Conditionally rendering nothing](docs/template/conditionals#conditionally-rendering-nothing) for more information.
+
+Rendering no node can be important when an expression is a child of an element with Shadow DOM that includes a `slot` with fallback content. Rendering no node ensures the fallback content is rendered. See [fallback content](/docs/components/shadow-dom/#fallback) for more information.
 
 ## Attribute expressions {#attribute-expressions }
 
@@ -218,7 +232,17 @@ If the expression makes up the entire attribute value, you can leave off the quo
 html`<img src="/images/${this.image}">`;
 ```
 
-### Removing an attribute under certain conditions { #removing-attribute }
+Note, some primitive values are handled specially in attributes. Boolean values are converted to strings so, for example, `false` renders `'false'`. Both `undefined` and `null` render to an attribute as an empty string.
+
+### Boolean attributes {#boolean-attribute-expressions }
+
+To set a boolean attribute, use the `?` prefix with the attribute name. The attribute is added if the expression evaluates to a truthy value, removed if it evaluates to a falsy value:
+
+```js
+html`<div ?hidden=${!this.showAdditional}>This text may be hidden.</div>`;
+```
+
+### Removing an attribute { #removing-attribute }
 
 Sometimes you want to set an attribute only if the value or set of values is available, and otherwise remove the attribute. If the attribute value is just an empty string, use a [boolean attribute](#boolean-attribute-expressions). Sometimes, however, an attribute should be set to a specific value.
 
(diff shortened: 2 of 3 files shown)


packages/lit-dev-content/site/docs/components/shadow-dom.md
```diff
@@ -169,7 +169,7 @@ You can specify fallback content for a slot. The fallback content is shown when
 
 <div class="alert alert-info">
 
-**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`.
+**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`. When using a Lit expression as a child of an element with Shadow DOM and a slot with fallback content, make sure to use a non-rendering value when appropriate so that fallback content is rendered. See [non-rendering values](/docs/templates/expressions/#non-rendering) for more information.
```
kevinpschaaf, line 172: The sentence is pretty dense, I think you can elide a bit of detail because "so that any slot fallback content" implies a CE with a shadow DOM with a slot with default content...
```suggestion
**Rendering fallback content.** If any child nodes are assigned to a slot, its fallback content doesn't render. A default slot with no name accepts any child nodes. It won't render fallback content even if the only assigned nodes are text nodes containing whitespace, for example `<example-element> </example-element>`. When using a Lit expression as a child of a custom element, make sure to use a non-rendering value when appropriate so that any slot fallback content is rendered. See [non-rendering values](/docs/templates/expressions/#non-rendering) for more information.
```

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,14 +137,21 @@ html`<main>${bodyText}</main>`
...
 
-* Primitive values likes strings, numbers, and booleans
+* Primitive rendering values
+* Non-rendering values
 * `TemplateResult` objects created with the `html` function.
 * DOM nodes
 * Arrays or iterables of any of the supported types
 
-### Primitive values
+### Primitive rendering values
```
kevinpschaaf, line 146: My initial gut is that trying to coin "Primitive rendering values" as a thing with a specific meaning is more confusing than just leaving the heading "Primitive values", and saying  "With the exception of empty string, all strings numbers and booleans are converted to strings..." in the prose.

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,14 +137,21 @@ html`<main>${bodyText}</main>`
...
 
-Primitives values like strings, numbers, booleans, null, and undefined are converted to strings when interpolated into text content or attribute values. They are checked for equality with the previous value so the DOM is not updated if the value hasn't changed.
+Strings, numbers, and booleans are all converted to strings when interpolated into text content. This means that a value of `5` will render the string `'5'`, but likewise a value of `true` will render `'true'` and a value of `false` will render `'false'`. Rendering a boolean like this is uncommon, and these values are typically rendered using conditionals. For more on conditionals, see [Conditionals](/docs/templates/conditionals/).
```
kevinpschaaf, line 148: > Rendering a boolean like this is uncommon, and these values are typically rendered using conditionals.

Don't totally follow this. What is "these values" referring to, booleans? You don't _render a boolean_ using a conditional. Do you mean to say something like "booleans are typically used when rendering conditionals or in boolean attribute expressions"?

kevinpschaaf: > They are checked for equality with the previous value so the DOM is not updated if the value hasn't changed.

Was this part intentionally deleted?

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,14 +137,21 @@ html`<main>${bodyText}</main>`
...
+
+### Non-rendering values {#non-rendering}
+
+The values `null`, `undefined`, the empty string `''`, and Lit's [nothing](/docs/api/templates/#nothing) sentinel value will all render no node when in a child expression.
+
+Rendering no node can be important when an expression is a child of an element with Shadow DOM that includes a `slot` with fallback content. Render no node ensures the fallback content is rendered. See [fallback content](/docs/components/shadow-dom/#fallback) for more information.
```
kevinpschaaf, line 154:
```suggestion
Rendering no node can be important when an expression is a child of an element with Shadow DOM that includes a `slot` with fallback content. Rendering no node ensures the fallback content is rendered. See [fallback content](/docs/components/shadow-dom/#fallback) for more information.
```

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -207,9 +214,11 @@ If the expression makes up the entire attribute value, you can leave off the quo
 html`<img src="/images/${this.image}">`;
 ```
 
-### Setting attributes if data is defined { #ifDefined }
+### Setting attributes if data is available { #attribute-available }
```
kevinpschaaf, line 217: I think this section might be better re-conceptualized as "Removing an attribute under certain conditions", and I think we should also demote `ifDefined` and instead start by describing that `nothing` removes attributes (this is the key primitive). 

And then maybe add a footnote about `ifDefined` that says "because typically `undefined` and `null` shouldn't ever be rendered as strings, we provide an `ifDefined()` helper that returns `value ?? nothing`, and can be used like this... [example].

packages/lit-dev-content/site/docs/templates/expressions.md
sorvell: Yes, it was the only place where the update behavior was addressed. It's an odd place to put that imo.

packages/lit-dev-content/site/docs/templates/expressions.md
sorvell: Removed.

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -214,7 +218,7 @@ If the expression makes up the entire attribute value, you can leave off the quo
 html`<img src="/images/${this.image}">`;
 ```
 
-### Setting attributes if data is available { #attribute-available }
+### Removing an attribute under certain conditions { #removing-attribute }
 
 Sometimes you want to set an attribute only if the value or set of values is available, and otherwise remove the attribute. If the attribute value is just an empty string, use a [boolean attribute](#boolean-attribute-expressions). Sometimes, however, an attribute should be set to a specific value.
```
kevinpschaaf, line 223: Just a little bit more clarification:
```suggestion
Sometimes you want to set an attribute only if the part's value(s) meet certain conditions, and otherwise remove the attribute. For common "boolean attributes" like `disabled` and `hidden` where you want to set the attribute to an empty string for a truthy value and remove it otherwise, use a [boolean attribute](#boolean-attribute-expressions). Sometimes, however, an attribute should be set to a specific value in some cases, and removed in others.
```

kevinpschaaf (approved):
One more minor clarification, otherwise LGTM

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,14 +137,25 @@ html`<main>${bodyText}</main>`
...
 ### Primitive values
 
-Primitives values like strings, numbers, booleans, null, and undefined are converted to strings when interpolated into text content or attribute values. They are checked for equality with the previous value so the DOM is not updated if the value hasn't changed.
+Strings, numbers, and booleans are all converted to strings when interpolated into text content.
+
+A value of `5` will render the string `'5'`. The empty string `''` is specially treated as a [non-rendering value](#non-rendering).
+
+A boolean value `true` will render `'true'`, and `false` will render `'false'`. Rendering a boolean like this is uncommon. Instead booleans are typically used in conditionals to render other appropriate values. For more on conditionals, see [Conditionals](/docs/templates/conditionals/).
+
+### Non-rendering values {#non-rendering}
```
justinfagnani, line 154: "Non-rendering" sounds to my ear like these have no action, or it describes a set of values that could include `noChange`. These values do render, they just render "nothing".

I would consider a phrase like "Empty values" or "Clearing values". React uses the word "skip" to describe how it handles `false`, but I don't think that quite right.

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,14 +137,25 @@ html`<main>${bodyText}</main>`
...
+* Primitive values
+* Non-rendering values
 * `TemplateResult` objects created with the `html` function.
 * DOM nodes
 * Arrays or iterables of any of the supported types
 
 ### Primitive values
 
-Primitives values like strings, numbers, booleans, null, and undefined are converted to strings when interpolated into text content or attribute values. They are checked for equality with the previous value so the DOM is not updated if the value hasn't changed.
+Strings, numbers, and booleans are all converted to strings when interpolated into text content.
```
justinfagnani, line 148: I think this might be clear if it mentions all primitives. So add bigint, mention that symbol throws, and mention right away that `null`, `undefined`, and `''` render nothing, with a link to the next section. Possibly link to https://developer.mozilla.org/en-US/docs/Glossary/Primitive

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,14 +137,25 @@ html`<main>${bodyText}</main>`
 
 Expressions in the child position can take many kinds of values:
 
-* Primitive values likes strings, numbers, and booleans
+* Primitive values
+* Non-rendering values
```
justinfagnani, line 141: I think this organization is mixing the _type_ of value with the _result_ of certain values. I would probably organize this list purely by type:

- Primitives
- `TemplateResult` objects created with the `html` and `svg` functions
- DOM Nodes
- The special sentinels `nothing` and `noChange`
- Arrays or iterables of any of the above

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -207,23 +218,39 @@ If the expression makes up the entire attribute value, you can leave off the quo
 html`<img src="/images/${this.image}">`;
 ```
 
-### Setting attributes if data is defined { #ifDefined }
+### Removing an attribute under certain conditions { #removing-attribute }
```
justinfagnani, line 221: I think it's not great that talking about removing at attribute is here, but the general talk about removing child expressions is either in the conditionals page or under a type-oriented hierarchy above. I think I'd move most of the discussion about rendering nothing in child expressions to a section outside of the type-based organization, that's analogous to this section on attributes, then link to conditionals from that section. And keep the part organized by type just organized by type.

packages/lit-dev-content/site/docs/templates/expressions.md
arthurevans: I think that non-rendering is clearer than "empty values" or "clearing values."  "Values that render nothing" is maybe? clearer but also longer. Next choice maybe "Empty values"?

I think for clarity it might be good to mention `noChange` . Maybe something like:

If you write custom directives, you'll encounter another sentinel value, `noChange`. The `noChange` sentinel is used to leave the previously-rendered value in place. For more information, see [Signaling no change](/docs/templates/custom-directives/#signaling-no-change).

packages/lit-dev-content/site/docs/templates/expressions.md
arthurevans: This makes sense to me. To reconcile the two orgs, I might add a sentence after the list that says something like:

A few values get special treatment. These include `null`, `undefined`, the empty string, and the sentinel values `nothing` and `noChange`. For more information, see [Non-rendering values](link) and [Signaling no change](link?).

packages/lit-dev-content/site/docs/templates/expressions.md
arthurevans: So, the type hierarchy is specifically about the types of values for child expressions (because there are so many and there's some complexity). It makes sense to me that the information about removing (or clearing) child expressions is in the child expression section and the information about removing attributes is in the attribute expression section.

I think the only really confusing thing here is that we treat attributes and boolean attributes as separate things. And as a side effect, we end up describing removing an attribute using `nothing` (arguably the advanced use case) before we describe boolean attribute expressions (arguably the simple use case). 

We could make that a little more logical by making the "boolean attribute expressions" section a child of the "Attribute expressions" section, and moving this "Removing an attribute..." section _after_ the boolean attr section.  It makes sense from a user point of view ("I want to ... X") even if it slightly less "correct" in some sense.

packages/lit-dev-content/site/docs/templates/expressions.md
arthurevans: Actually, stepping back a bit, adding a bullet for "the special sentinels `nothing` and `noChange`" means we have to add another section in here just to explain that there are two sentinel values that do different things and are discussed elsewhere. That seems not great.

I think I'd lean towards either dropping this bullet & adding a paragraph like I suggested above, or keeping Steve's original list and just resigning ourselves that this list buckets values by treatment, rather than by type.

packages/lit-dev-content/site/docs/templates/expressions.md
kevinpschaaf: Or we could just add a "(see below)" parenthetical to the boolean attribute syntax callout; doesn't seem like the end of the world to have a use-before-define in this case, but I think what you said works too.

At this point, both `ifDefined` and arguably boolean attribute `?attr=` syntax are both vestiges of not initially having a sentinel to remove attributes; if we had that, I'm not sure we'd have either (maybe arguable).

As for placement, agree with Arthur that pulling it out into a "rendering nothing" section that covers both child and attr expressions isn't clearly better than describing `nothing`'s behavior per-expression type, since it does do different things per type that needs to be explained.

justinfagnani: I'm not suggesting a "rendering noting" section that covers both attributes and children. I'm suggesting that both child expressions and attribute expressions sections had "rendering nothing" subsections, because one right now is here and the other is in the conditionals page.

packages/lit-dev-content/site/docs/templates/expressions.md
sorvell: Attempted to address feedback, ptal.

packages/lit-dev-content/site/docs/templates/expressions.md
sorvell: Incorporated.

packages/lit-dev-content/site/docs/templates/expressions.md
sorvell: Attempted to address feedback, ptal.

packages/lit-dev-content/site/docs/templates/expressions.md
sorvell: Attempted to address feedback, ptal.

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -207,32 +232,52 @@ If the expression makes up the entire attribute value, you can leave off the quo
...
+For example, consider:
 
 ```js
 html`<img src="/images/${this.imagePath}/${this.imageFile}">`;
 ```
 
 If `this.imagePath` or `this.imageFile` is not defined, the `src` attribute should not be set or an invalid network request will occur.
 
-You can use the [ifDefined](/docs/api/directives/#ifDefined) directive to avoid this issue:
+Lit's [nothing](/docs/api/templates/#nothing) sentinel value addresses this by removing the attribute when any part of it contains `nothing`.
```
arthurevans, line 257:
```suggestion
Lit's [nothing](/docs/api/templates/#nothing) sentinel value addresses this by removing the attribute when any expression in the attribute value evaluates to `nothing`.
```

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -207,32 +232,52 @@ If the expression makes up the entire attribute value, you can leave off the quo
...
+
+To set a boolean attribute, use the `?` prefix with the attribute name. The attribute is added if the expression evaluates to a truthy value, removed if it evaluates to a falsy value:
+
+```js
+html`<div ?hidden=${!this.showAdditional}>This text may be hidden.</div>`;
+```
+
+### Removing an attribute { #removing-attribute }
+
+Sometimes you want to set an attribute only if the part's value(s) meet certain conditions, and otherwise remove the attribute. For common "boolean attributes" like `disabled` and `hidden` where you want to set the attribute to an empty string for a truthy value and remove it otherwise, use a [boolean attribute](#boolean-attribute-expressions). Sometimes, however, an attribute should be set to a specific value in some cases, and removed in others.
```
arthurevans, line 247: I think "the parts values" is kind of confusing, because we don't really talk about parts until we get into custom directives. So maybe focus instead on the desired outcome?

```suggestion
Sometimes you want to set an attribute only under certain conditions, and otherwise remove the attribute. For common "boolean attributes" like `disabled` and `hidden` where you want to set the attribute to an empty string for a truthy value and remove it otherwise, use a [boolean attribute](#boolean-attribute-expressions). Sometimes, however, you might require a different condition for adding or removing an attribute. 
```

arthurevans (approved):
Looks good to me. Made a few small suggestions on wording.

K780420 (commented):
##task update

PR #911, opened 2022-08-23
Add articles to site search

### LINKED ISSUE #912: Add the articles section to the search index
currently not indexed by search

e111077:
Fixes #912

Add articles to site search:

Since there are tags pages and because we wanted the articles URLs to be in `/articles/...` rather than `/articles/article/...`, I had to add a feature to the site crawler that skips indexing those pages.

I also had to move the `<h1>` out of the article header because the indexer was not associating the following content with the h1

### DIFF
--- a/packages/lit-dev-tools-cjs/src/search/plugin.ts
+++ b/packages/lit-dev-tools-cjs/src/search/plugin.ts
@@ -41,6 +41,16 @@ export async function createSearchIndex(outputDir: '_dev' | '_site') {
   );
   const relativeDocUrlsToHtmlFile: UrlToFile = walkDir(DOCS_PATH, new Map());
 
+  const ARTICLES_PATH = path.resolve(
+    __dirname,
+    // Load the article content itself not the tags pages.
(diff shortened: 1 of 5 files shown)


e111077:
cc @arthurevans

packages/lit-dev-tools-cjs/src/search/plugin.ts
```diff
@@ -41,6 +41,40 @@ export async function createSearchIndex(outputDir: '_dev' | '_site') {
...
   const relativeDocUrlsToHtmlFile: UrlToFile = walkDir(DOCS_PATH, new Map());
 
+  const ARTICLES_PATH = path.resolve(
+    __dirname,
+    // Load the article content itself not the tags pages.
+    `../../../lit-dev-content/${outputDir}/articles`
+  );
+
+  const skipFiles = (filepath: string) => {
+    // These pages have are tag feeds and are not structured in a way the
```
aomarks, line 51:
```suggestion
    // These pages are tag feeds and are not structured in a way the
```

aomarks (approved):
Nice! Just 1 nit.

PR #919, opened 2022-08-27
Introduce an "Edit this page" link

### LINKED ISSUE #450: [docs] Docs feedback mechanism
Another general design request for post-1.0: I'd like to provide a feedback mechanism. We can easily put one at the bottom of the page, but since some of our pages are quite long, having a "was this page helpful" button at the very end isn't perfect.

Old Polymer site had "Edit on GitHub" buttons, which were nice, but much higher friction, since users had to have a GitHub account, be willing to click through and create a PR, instead of just commenting anonymously.

Options include: 

- A sticky footer (see the [Sentry docs](https://docs.sentry.io/workflow/releases/?platform=javascript).
- Something attached to the header.
- A small feedback tab on the side that opens an overlay. I'm not saying this well, but I've seen that implemented somewhere, too.

Information to collect:
Was this page helpful to you? [Yes] {No]
[If no] Comment [  ] 
Could also offer "Edit on GitHub" or "Open an Issue" as options.

e111077:
Fixes #450

There is now a link at the bottom of every doc and article Links to test:

https://pr919-f93c133---lit-dev-5ftespv5na-uc.a.run.app//docs/
https://pr919-f93c133---lit-dev-5ftespv5na-uc.a.run.app//articles/lit-for-polymer-users


justinfagnani:
I feel like the edit button is a bit too prominent. Given that our pages are mostly flat, and this button looks like a raised FAB, in the Material paradigm this would be the primary action of a page.

Do we have some other examples of an edit link that's more subtle? Could maybe also combine this with a "lasted updated on..." text? (I know that may be harder because we need to look at file write times, but maybe 11ty already pipes that through?)

augustjk:
Most other call to edits I've seen are a bit more subtle, and require scrolling all the way to the bottom of the page. Last updated probably needs to grab the last git commit date for the file which seems extra.

e111077:
Can create a link at the bottom of the page. Also "last edited x" we added it to the articles but I think only because that info can actually get stale which I don't think is necessarily the case for the docs.

justinfagnani:
> Can create a link at the bottom of the page.

The bottom of docs pages looks a little weird to me currently, with the page content background just running right into the footer. Can we fix this up a little at the same time? ie, add a bit of margin so that the bottom of the content area looks like the top, then put the edit link below:

Screen Shot 2022-08-29 at 12 05 30 PM

cc @aomarks @arthurevans

e111077:
Updated to a link at the bottom of the page. LMK what you think. Esp about the text e.g. should it be `Suggest an edit to this page`

aomarks:
> The bottom of docs pages looks a little weird to me currently, with the page content background just running right into the footer. Can we fix this up a little at the same time? ie, add a bit of margin so that the bottom of the content area looks like the top, then put the edit link below:

How about putting the link inside the white part of the page, above the table of contents? That's how Vue does it: https://vuejs.org/guide/reusability/custom-directives.html#directive-hooks I think the left aligned looks a bit better too.

aomarks:
> > The bottom of docs pages looks a little weird to me currently, with the page content background just running right into the footer. Can we fix this up a little at the same time? ie, add a bit of margin so that the bottom of the content area looks like the top, then put the edit link below:
> 
> How about putting the link inside the white part of the page, above the table of contents? That's how Vue does it: https://vuejs.org/guide/reusability/custom-directives.html#directive-hooks I think the left aligned looks a bit better too.

The "Edit Square" icon (square with pencil in it) (https://fonts.google.com/icons) also seems better than the "Open in New" icon used in the latest commit, since that suggests editing. The current icon just reads as "external link" to me.

justinfagnani:
> How about putting the link inside the white part of the page, above the table of contents?

That looks pretty good there. I like the left alignment and "Edit this page on GitHub" text too.

e111077:
Did the changes. Also added justin's suggestion for the padding on the docs

aomarks (approved):

PR #920, opened 2022-08-30
Switch from minisearch to algolia

### LINKED ISSUE #926: Implement remote search
Consider switching to algolia.

- It has better search results
- we can make our searches finer-grained with little impact
- results in fewer bytes over the wire
- easier to integrate with third-party libraries

e111077:
fixes #926 

This PR:

- switches from minisearch to algolia
- indexes all the way down to `<h4>`s
- introduces a GH action that publishes the latest search index to algolia
- adds a bash script to copy the public keys into content dir and turn them into a TS default export file

Conclusions:

- algolia library is about 5k minzipped
- minisearch downloaded a 350k (~50k minzipped) index on the client
- I increased our index's granuarity to include `h3`s and `h4s` and that increased index by 1-200k (30k minzipped)
- algolia queries scale per search and minisearch scales per content
- 15% byte xfer decrease
- 12% byte xfer decrease when entered 3 search terms

Regressions:

Introducing `@lit-labs/task` into the search component has made rollup split up task into its own package (likely because used in both this and in tutorials). That means that we now have 2 network requests for just hydrating the UI of the searchbox which eats into the algolia implementation's perf.

Future work:

- create a modal / fullscreen UI that groups results per title


e111077:
Would super appreciate feedback on the GH action

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -240,8 +257,30 @@ export class LitDevSearch extends LitElement {
...
+
+  private get _isExpanded() {
+    return (
+      this._isFocused &&
+      this._currentSuggestions.length > 0 &&
+      !this._closedByEscape
+    );
+  }
+
+  update(changed: Map<string, unknown>) {
```
augustjk, line 279: Minor comment since we don't need `changed` other than to pass into `super.update()`, but was the type change here intentional?

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -22,17 +22,22 @@ import {
...
 import {live} from 'lit/directives/live.js';
-import Minisearch from 'minisearch';
 import type {Drawer} from '@material/mwc-drawer';
 import {animate, Options as AnimationOptions} from '@lit-labs/motion';
+import algoliasearch, {
+  SearchClient,
+  SearchIndex,
+} from 'algoliasearch/dist/algoliasearch-lite.esm.browser.js';
+import {Task, TaskStatus} from '@lit-labs/task';
+import vars from '../public-vars.js';
```
augustjk, line 32: I see `"resolveJsonModule": true` was added to the tsconfig. Is this converting to ts file and importing necessary then?

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -22,17 +22,22 @@ import {
 import {repeat} from 'lit/directives/repeat.js';
 import {styleMap} from 'lit/directives/style-map.js';
 import {live} from 'lit/directives/live.js';
-import Minisearch from 'minisearch';
 import type {Drawer} from '@material/mwc-drawer';
 import {animate, Options as AnimationOptions} from '@lit-labs/motion';
+import algoliasearch, {
+  SearchClient,
+  SearchIndex,
+} from 'algoliasearch/dist/algoliasearch-lite.esm.browser.js';
```
augustjk, line 30: We do our own bundling with rollup, why do we need to reach into the dist directory for the `algoliasearch` package? Hmm.. it looks like their default export from the package might be cjs so maybe that's why. 😞 I don't know the answer to this but is TS unable to handle that with `esModuleInterop` and/or `allowSyntheticDefaultImports`?

packages/lit-dev-tools-esm/src/upload-algolia-index.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+import algolia from 'algoliasearch';
+import fs from 'fs';
+
+const varsRaw = fs.readFileSync('../../public-vars.json', 'utf8');
```
augustjk, line 10: doesn't `resolveJsonModule` let us import json files directly?

.github/workflows/build-and-publish-search-index.yml
```diff
@@ -0,0 +1,27 @@
+name: Build And Publish Search Index
+
+on:
+  push:
+    branches:
+      - main
```
augustjk, line 6: If this is published on only push to main, our PR previews's search won't be based on that content but the main content.

Perhaps this is acceptable.. is there a way to have multiple indexes?

packages/lit-dev-content/src/components/litdev-search.ts
e111077: So midway I was experimenting with accessing a private property that was changed, and (known issue) `PropertyValues` can't see private properties due to TS limitations. I've since reverted that, but didn't change this back. Do you think it's worth changing back?

packages/lit-dev-content/src/components/litdev-search.ts
e111077: oops forgot to clean up that flag!

packages/lit-dev-content/src/components/litdev-search.ts
e111077: We need to reach into dist folder because this is the only proper ESM browser build. And our dev build intentionally does no bundling just serve TSC. That means we can't do weird cjs imports without lots of infra which I didn't want to touch

packages/lit-dev-tools-esm/src/upload-algolia-index.ts
e111077: ooOOOOooo

.github/workflows/build-and-publish-search-index.yml
e111077: We can have multiple indicies, but you can't just create them on the fly. So if we had 2 indicies, then if person A creates PR 1 and person B creates PR 2, then they both upsert into the same dev index, then PR 1's preview will be querying an index that's not representative of their current preview anyway

packages/lit-dev-tools-esm/src/upload-algolia-index.ts
e111077: oops nope don't work

packages/lit-dev-content/src/components/litdev-search.ts
augustjk: Eh, I say not a big deal.

packages/lit-dev-content/src/components/litdev-search.ts
augustjk: That sounds like a good reason to me!

.github/workflows/build-and-publish-search-index.yml
augustjk: In that case, it seems reasonable for now to just only use and update the index for the prod version of our site.

augustjk: The other potential problem I see with this is that since the publishing of index is separate from our site cloud build, they can fail independently causing the site and index to be out of sync. The danger is quite minimal since I wouldn't expect the site content to have such a significant change that the off sync would be much noticeable..

We could potentially move the index upload into the `cloudbuild-main.yaml` as an extra step so it happens together with the site build instead of within github actions. It's possible to add a step like this https://github.com/lit/lit.dev/blob/de6e81d70a123a6ceb76471ed67b43d8b714c4c2/cloudbuild-cleanup.yaml#L6-L12
Maybe to run after the site build is successful? Then if the site build/deploy fails, we don't upload the new index. The algolia write key would need to be added to GCP then.

e111077: I believe in you

augustjk: I will not fail you!

justinfagnani: fwiw, I think it's fine to have the index only be for the live site.

@augustjk being part of the build sounds probably a bit better, but given that Cloud Run lets us deploy separately from builds, there's a chance that's not quite right either. I think it's an edge case of an edge case, but it'd be cool if the index was updated on _deployment_ of an image.

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -22,17 +22,22 @@ import {
...
+import {Task, TaskStatus} from '@lit-labs/task';
+import {publicVars} from 'lit-dev-tools-esm/lib/configs.js';
 
 /**
  * Representation of each document indexed by Minisearch.
```
justinfagnani, line 35: Update "Minisearch"?

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -68,6 +73,20 @@ function isApiLink(url: string) {
   return url.includes('docs/api/');
 }
 
+/**
+ * Used to mark a search page suggestion with the DOC chip.
+ */
+function isDocsLink(url: string) {
+  return url.includes('docs/') && !isApiLink(url);
```
justinfagnani, line 80: Can we do a more specific condition, like startsWith, or `new URL(url).pathname.startsWith('docs/')`?

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -68,6 +73,20 @@ function isApiLink(url: string) {
...
+}
+
+/**
+ * Used to mark a search page suggestion with the ARTICLE chip.
+ */
+function isArticleLink(url: string) {
+  return url.includes('articles/');
```
justinfagnani, line 87: Same comment as above

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -229,6 +240,12 @@ export class LitDevSearch extends LitElement {
   @state()
   private _popupSpaceRight = false;
 
+  /**
+   * Whether or not the user has closed the menu with the escape key.
```
justinfagnani, line 244: Can you add why we care here?

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -424,24 +425,28 @@ export class LitDevSearch extends LitElement {
...
-        this._suggestions = [];
+        this._closedByEscape = true;
         this._selectedIndex = -1;
         // prevent the input from closing drawer if there was text
         if (oldText.trim()) {
           e.stopPropagation();
         }
         break;
+      default:
+        this._closedByEscape = false;
```
justinfagnani, line 448: Could we do this in a `blur` handler?

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -322,69 +361,28 @@ export class LitDevSearch extends LitElement {
...
    * Load the search index.
    */
   async firstUpdated() {
     // required for popping up on hydration
     this._isFocused = !!this.shadowRoot?.activeElement;
-    await this._loadSearchIndex();
-    // If a search query has already been written, fill suggestions.
-    this._querySearch(this._searchText);
+    // Initialize the search task only on the client and not the server.
+    this._searchTask = new Task(
```
justinfagnani, line 371: I think we should usually avoid dynamically creating controllers like this...  The Task constructor doesn't look like it does any server-incompatible.

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -240,8 +257,30 @@ export class LitDevSearch extends LitElement {
...
+
+    this._lastSuggestions = this._searchTask.value!;
+    return this._searchTask.value!;
+  }
+
+  private get _isExpanded() {
+    return (
+      this._isFocused &&
+      this._currentSuggestions.length > 0 &&
+      !this._closedByEscape
```
justinfagnani, line 275: I think another way to handle this, that seems to be consistent with some search interface like GitHub's, would be to just blur on escape. Then you can elide the `_closedByEscape` field.

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -240,8 +257,30 @@ export class LitDevSearch extends LitElement {
...
+    );
+  }
+
+  update(changed: Map<string, unknown>) {
+    const textDoesntMatch = this._searchText !== this._inputEl.value;
+    const isSSRHydrate = this._inputEl && textDoesntMatch && !this.hasUpdated;
```
justinfagnani, line 281: Can you add a little more comment about this code? This is the first I'm hearing about SSR + hydration issues with inputs. Do we need to file an issue and track this?

packages/lit-dev-content/src/components/litdev-search.ts
justinfagnani: `PropertyValues` without a generic should be the same

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -393,28 +391,31 @@ export class LitDevSearch extends LitElement {
...
+    this._selectedIndex = -1;
   }
 
   /**
    * Populate suggestion dropdown from query.
    *
    * An empty query clears suggestions.
    */
-  private _querySearch(query: string) {
+  private async _querySearch(query: string): Promise<Suggestion[]> {
```
justinfagnani, line 402: Can this be inlined into the Task definition?

packages/lit-dev-content/src/components/litdev-search.ts
justinfagnani, line 371: Very, very much optional, but as an exercise in composing controllers and Task, it might be nice to see an AlgoliaSearchController that composes task and builds-in the search API function and _algoliaClient

packages/lit-dev-content/tsconfig.json
```diff
@@ -8,6 +8,6 @@
     "tsBuildInfoFile": "lib/.tsbuildinfo",
     "skipLibCheck": true
   },
-  "include": ["src/**/*.ts"],
+  "include": ["src/**/*.ts", "src/**/*.json"],
```
justinfagnani, line 11: Are you importing JSON files somewhere?

justinfagnani (commented):
Looks great!

packages/lit-dev-content/src/components/litdev-search.ts
e111077: a11y model for combobox is that the input is never un-focused even as you navigate up and down. And if you have a native input that you hit `escape` it will not blur the input. Though, the menu still closes if the input is blurred.

packages/lit-dev-content/src/components/litdev-search.ts
e111077: i forgot that `autoRun: false` was a thing until I started writing the outline for the task talk lol

packages/lit-dev-content/tsconfig.json
e111077: i _was_ >.> nice catch!

packages/lit-dev-content/src/components/litdev-search.ts
e111077: This is specifically because if you type in there and then it hydrates, because `@input` did not update `this._searchText` the `live()` directive will overwrite what is inside the `<input>` with `''` and the `live()` directive is necessary because IIRC safari on ios will make the cursor jump to the end if you set the value needlessly while editing it

packages/lit-dev-content/src/components/litdev-search.ts
e111077: they are not full URLs but rather `/docs/category/doc-name/index.html` I switched to using `startsWith` LMK what you think

packages/lit-dev-content/src/components/litdev-search.ts
e111077: did blur anyway

packages/lit-dev-content/src/components/litdev-search.ts
e111077: comment added LMK what you think

packages/lit-dev-content/src/components/litdev-search.ts
e111077: not applicable anymore

packages/lit-dev-content/src/components/litdev-search.ts
e111077: I did it. LMK what you think. It could not be a true controller because it does not need any of the interface

packages/lit-dev-content/src/components/litdev-search.ts
justinfagnani: ooh, can you file an SSR issue describing this? live() might not be the exact right thing for this case, but not sure yet.

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -0,0 +1,72 @@
...
+  searchOnlyKey: publicVars.algolia.searchOnlyKey,
+  index: publicVars.algolia.index,
+};
+
+export type AlgoliaSearchControllerOptions =
+  typeof agloliaSearchControllerDefaultOptions;
+
+export class AgloliaSearchController<T extends {}> {
+  host: ReactiveControllerHost;
+  private task;
```
justinfagnani, line 28: Name with an `_` prefix and maybe give the type with task value type param?

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -0,0 +1,72 @@
...
+  appId: publicVars.algolia.appId,
+  searchOnlyKey: publicVars.algolia.searchOnlyKey,
+  index: publicVars.algolia.index,
+};
+
+export type AlgoliaSearchControllerOptions =
+  typeof agloliaSearchControllerDefaultOptions;
+
+export class AgloliaSearchController<T extends {}> {
+  host: ReactiveControllerHost;
```
justinfagnani, line 27: private? Or I think you could remove it.

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -0,0 +1,72 @@
...
+  private task;
+
+  private _client: SearchClient;
+  private _index: SearchIndex;
+  private _lastValue: Hit<T>[] = [];
+
+  public get value() {
+    if (!this.task || this.task.status !== TaskStatus.COMPLETE) {
```
justinfagnani, line 35: `this.task` should always be defined

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -0,0 +1,72 @@
...
+      return this._lastValue;
+    }
+
+    this._lastValue = this.task.value!;
+    return this.task.value!;
+  }
+
+  constructor(
+    host: ReactiveControllerHost,
+    argsFn: ArgsFunction<[string]>,
```
justinfagnani, line 45: You could type this as `() => string` so that the element doesn't have to put the query in an array.

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Populate suggestion dropdown from query.
+   *
+   * An empty query clears suggestions.
+   */
+  private async _querySearch(query: string): Promise<Hit<T>[]> {
+    const trimmedQuery = query.trim();
+    if (!this._index || trimmedQuery === '' || trimmedQuery.length < 2) {
```
justinfagnani, line 63: `this._index` is always defined, yes?

justinfagnani, line 63: `trimmedQuery.length < 2` covers the `trimmedQuery === ''` case already

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -0,0 +1,72 @@
...
+const agloliaSearchControllerDefaultOptions = {
+  appId: publicVars.algolia.appId,
+  searchOnlyKey: publicVars.algolia.searchOnlyKey,
+  index: publicVars.algolia.index,
+};
+
+export type AlgoliaSearchControllerOptions =
+  typeof agloliaSearchControllerDefaultOptions;
+
+export class AgloliaSearchController<T extends {}> {
```
justinfagnani, line 26: I like it!

packages/lit-dev-content/src/components/litdev-search.ts
justinfagnani: lgtm

justinfagnani (commented):
Looks good to me, pending figuring out exactly when the search index is deployed

packages/lit-dev-content/src/components/litdev-search.ts
e111077: filed

packages/lit-dev-content/src/components/algolia-search-controller.ts
e111077: you caught a few copy-pasta issues

packages/lit-dev-content/src/components/algolia-search-controller.ts
e111077: lmao overthunk

.github/workflows/build-and-publish-search-index.yml
augustjk: The cloud build is responsible for both building and deploying so adding a step after that for the index upload would work. If we manually change which revision gets the traffic, this potentially won't work but that would be a very rare occurrence.

Now that I think about it more though, I wonder if it's better to keep this in github actions and configure the upload to be a wireit script so we can take advantage of wireit cache and skip unless the index actually changed?

We don't have a solution for utilizing wireit cache for cloud build yet. It's possible to manually configure persistence of artifacts with google storage. Would it be worth it to pursue that?

e111077:
Re-review ping!

.github/workflows/build-and-publish-search-index.yml
e111077: created wireit command

packages/lit-dev-tools-esm/package.json
```diff
@@ -31,13 +32,27 @@
       ],
       "command": "../../node_modules/.bin/tsc --build --pretty",
       "clean": "if-file-deleted"
+    },
+    "publish-search-index": {
+      "dependencies": [
+        "build"
+      ],
+      "files": [
+        "lib/**"
```
augustjk, line 41: not sure if possible to reach out of the project but `../lit-dev-content/_site/searchIndex.json` should also be included, which means `lit-dev-content:build` or whichever command that generates it probably should be a dependency too. if dependency can't be outside the project which i suspect is the case, it might need to be specified in the top level wireit command.

e111077: muy excelente

packages/lit-dev-tools-esm/package.json
```diff
@@ -31,13 +32,24 @@
       ],
       "command": "../../node_modules/.bin/tsc --build --pretty",
       "clean": "if-file-deleted"
+    },
+    "publish-search-index": {
+      "dependencies": [
+        "build"
+      ],
```
augustjk, line 39: Need to build content first so it generates `searchIndex.json`.
```suggestion
      "dependencies": [
        "../lit-dev-content:build",
        "build"
      ],
```

augustjk (changes requested):

augustjk (approved):

PR #921, opened 2022-08-30
Locally host fonts

e111077:


e111077:
Since Andrew is out, @augustjk, you're getting my lit.dev reviews in his place

aomarks:
Did you look at https://github.com/lit/lit.dev/pull/529? The blocker there was just that the Manrope from Google Fonts is kind of broken on some people's screens, so we needed to update it to fetch Manrope from GitHub instead.

An advantage of the other PR is that we can auto-update the fonts. It has some handling for licenses too.

e111077:
What is the status of that PR? What can we do to move that forward?

packages/lit-dev-content/site/css/fonts/open-sans.css
```diff
@@ -5,7 +5,7 @@
   font-style: normal;
   font-weight: 300;
   font-display: swap;
-  src: url(https://fonts.gstatic.com/s/opensans/v18/mem5YaGs126MiZpBA-UN_r8OUuhp.woff2) format('woff2');
+  src: url(/fonts/opensans/OpenSans-300.woff2) format('woff2');
```
augustjk, line 8: should the filename have some indication that these are for latin chars only?

augustjk (commented):
Is having a way to update the open sans and roboto fonts important like we currently have for manrope? I can see this being a stop gap until we move forward with the other PR. But if the other PR is close enough maybe it's worth pursuing that directly.

aomarks:
> Is having a way to update the open sans and roboto fonts important like we currently have for manrope? I can see this being a stop gap until we move forward with the other PR. But if the other PR is close enough maybe it's worth pursuing that directly.

Yeah, I'm fine with landing this. Updating fonts is something we probably want to do only very rarely.

You may want to look into a LICENSE file for the fonts folder, just for GitHub. I think that may be required. The script in the other PR created a combined one, I believe.

packages/lit-dev-content/site/css/fonts/open-sans.css
e111077: There wasn't really an indication before and we don't even know how we are going to handle localization. I say we hold off on this for now unless you think it's something we should push for?

augustjk: I'm guessing the original gstatic url came from something like this https://fonts.googleapis.com/css2?family=Roboto+Mono based on src and unicode-range. If the woff2 files being added here were downloaded by following the src url, I assume it contains only latin characters of the font and could potentially be misleading down the line if we need more than latin chars.

Having said that, the `unicode-range` is still present here which does indicate this is a subset, and we don't have plans for additional localization anytime soon. By then we may have also moved forward with the font update situation making this moot. So not a big deal if we don't want to make any changes on this.

e111077: I'll include it. I just can't read mangled urls quite as good as you 😄

augustjk: I thought everyone could encode/decode all forms of mangling in their head 🤖

augustjk (approved):

e111077:
suggestions in. @aomarks what should do about max-age cache headers? Should I emulate what you did in the other?

I just looked at what other framework sites set for their cache headers and copied the most-common ones

augustjk (changes requested):
Oh, can we also remove this line in this PR?
https://github.com/lit/lit.dev/blob/4d395c5034758f5895e183a0a7a40402a95aadba/packages/lit-dev-content/site/_includes/default.html#L44

e111077:
@augustjk also removed from CSP as asked in DM

augustjk (approved):
Thank you!

aomarks (approved):
Yay

PR #935, opened 2022-09-06
docs: anchor for renderable

bennypowers:
Makes https://lit.dev/docs/components/rendering/#:~:text=typically%2C%20the%20component's the canonical docs location for `renderable`

Alternatively, A new section on rendering, or a heading on the [templates](https://lit.dev/docs/templates/overview/) page could be appropriate

## Y THO?

So I'm writing the sentence fragment "Icons are JavaScript modules which export a lit renderable" in my docs. Is there a canonical docs URL for "lit renderable" asides from the list mentioned off-hand at https://lit.dev/docs/components/rendering/

Renderable feels a bit opaque at times, with its `unknown` type. I'd like to be able to provide some more comprehensive and focused guidance.

So this issue is about **docs** and explicitly not about **types.** I'm fine with `unknown` if it's strictly true, but I believe that in the absence of clear type info, "thou shalt fully document thine API's intentions"

### DIFF
--- a/packages/lit-dev-content/site/docs/components/rendering.md
+++ b/packages/lit-dev-content/site/docs/components/rendering.md
@@ -18,14 +18,7 @@ Write your template in HTML inside a JavaScript [tagged template literal](https:
 
 Lit templates can include JavaScript _expressions_. You can use expressions to set text content, attributes, properties, and event listeners. The `render()` method can also include any JavaScript—for example, you can create local variables for use in expressions.
 
-### Renderable
-
-Typically, the component's `render()` method returns a single `TemplateResult` object (the same type returned by the `html` tag function). However, it can return anything that Lit can render:
-
-*   Primitive values like string, number, or boolean.
-*   `TemplateResult` objects created by the `html` function.
-*   DOM Nodes.
-*   Arrays or iterables of any of the supported types.
+Typically, the component's `render()` method returns a single `TemplateResult` object (the same type returned by the `html` tag function). However, it can return [anything that Lit can render](/docs/templates/overview/#renderable).
 
 For more information about writing templates, see [Templates](/docs/templates/overview/).
 
--- a/packages/lit-dev-content/site/docs/components/rendering.md
+++ b/packages/lit-dev-content/site/docs/components/rendering.md
@@ -18,6 +18,8 @@ Write your template in HTML inside a JavaScript [tagged template literal](https:
 
 Lit templates can include JavaScript _expressions_. You can use expressions to set text content, attributes, properties, and event listeners. The `render()` method can also include any JavaScript—for example, you can create local variables for use in expressions.
 
+### Renderable
+
 Typically, the component's `render()` method returns a single `TemplateResult` object (the same type returned by the `html` tag function). However, it can return anything that Lit can render:
 
 *   Primitive values like string, number, or boolean.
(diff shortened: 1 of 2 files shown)


aomarks:
Preview: https://pr935-2f38eaf---lit-dev-5ftespv5na-uc.a.run.app/docs/components/rendering/#renderable

packages/lit-dev-content/site/docs/components/rendering.md
```diff
@@ -18,6 +18,8 @@ Write your template in HTML inside a JavaScript [tagged template literal](https:
 
 Lit templates can include JavaScript _expressions_. You can use expressions to set text content, attributes, properties, and event listeners. The `render()` method can also include any JavaScript—for example, you can create local variables for use in expressions.
 
+### Renderable
```
justinfagnani, line 21: I think this is fine, but the canonical location for this info should be in the template docs, since different expression positions take different types. We mention the types in a similar list in *Child expressions* for example: https://lit.dev/docs/templates/expressions/#child-expressions

Putting headings there would let you link to the specific renderable you're talking about, ie, child vs attribute expression.

bennypowers: makes sense. i'll update shortly

bennypowers: moved the list to templating overview and linked to it from rendering

justinfagnani: I think we should link to https://lit.dev/docs/templates/expressions/#child-expressions specifically, because what `LitElement.render()` returns is used in a child-expression. The types are the same.

bennypowers: My goal here was to create a canonical docs link for the name `renderable`, regardless of the type. IMO linking off to `child-expressions` doesn't do the trick, because it doesn't express "the thing you pass as the first arg to `render()`" as much as "a thing you can put in child position in a template"

arthurevans: I think the previous setup, having Renderables (although I might call it Renderable values) in the Rendering chapter is a good thin. I fear this change reverts to the earlier state of confusion. It's hard to _find_ a list of what. you can render, and when you get to it, it's not super clear where it applies.

The rendering chapter should have a brief section that 1) clearly identifies that the render() method can return anything that can be rendered by a child expression. 2) links to a section on what you can render in a child expression for more detailed information.

My intuition is that 90% of the time people are wondering about what Lit can render, they're thinking about the return of the render() method, or what goes into a child expression. For most of the other expressions, the list is smaller and fairly intuitive.

arthurevans: Rather than try and make suggestions here, I'm going to make a separate PR to compare & contrast.

justinfagnani: I think @arthurevans's PR looks good. The one thing it leaves out, that @bennypowers is focusing on here, is describing the first argument to lit-html's `render()`, which is slightly different from LitElement's `render()` return type in that it can include `svg` results if the container is an svg element.

Can we express that in the standalone lit-html docs?

arthurevans: I'll open a ticket for that. Seems like the Standalone lit-html intro needs a face lift anyway.

arthurevans: Opened as https://github.com/lit/lit.dev/issues/985

arthurevans:
Fixed in #953.

PR #943, opened 2022-09-12
Decrease algolia response size

e111077:
Expose some features from algolia search to narrow our response attributes. This decreases the algolia response size from 15.7kb (3.4kb compressed) to 1.9kb (1.1kb)

| before | after |
| - | - |
| image | image |

### DIFF
--- a/packages/lit-dev-content/src/components/algolia-search-controller.ts
+++ b/packages/lit-dev-content/src/components/algolia-search-controller.ts
@@ -18,6 +18,12 @@ const agloliaSearchControllerDefaultOptions = {
   appId: publicVars.algolia.appId,
   searchOnlyKey: publicVars.algolia.searchOnlyKey,
   index: publicVars.algolia.index,
+  attributesToHighlight: undefined as string[] | undefined,
+  attributesToRetrieve: undefined as string[] | undefined,
+};
+
+type Writeable<T extends {[x: string]: any}, K extends string | number> = {
+  [P in K]: T[P];
 };
 
 export type AlgoliaSearchControllerOptions =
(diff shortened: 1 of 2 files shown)


packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -63,10 +75,27 @@ export class AgloliaSearchController<T extends {}> {
...
       return [];
     }
-    const results = await this._index.search<T>(trimmedQuery, {
+    type SearchOptionsReadonly = NonNullable<
+      Parameters<typeof this._index.search>[1]
+    >;
+    type SearchOptions = Writeable<
+      SearchOptionsReadonly,
+      keyof SearchOptionsReadonly
+    >;
```
AndrewJakubowicz, line 84: I am not following the types and why they are required? Is it for the `searchOpts.attributesToHighlight` mutation happening beneath?

Would it be possible to pass `this._attributesToHighlight` explicitly in the search options literal?

I.e.:

```
{ page: 0, hitsPerPage: 10, attributesToHighlight: this._attributesToHighlight }
```

Although that assumes an explicit `undefined` value is treated the same as not providing the option at all.

e111077: set some sensible defaults

AndrewJakubowicz (approved):
Nice!

packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -63,10 +71,15 @@ export class AgloliaSearchController<T extends {}> {
     if (trimmedQuery.length < 2) {
       return [];
     }
-    const results = await this._index.search<T>(trimmedQuery, {
+    type SearchOptions = Parameters<typeof this._index.search>[1];
```
AndrewJakubowicz, line 74: Do we need this type? Can the literal just be inlined or the type inferred?

e111077: Sorry I merged before this comment, but without the type, it won't cause a build failure

e111077: Actually it won't cause a build failure with the type either, but you get autocomplete to help you out there

PR #944, opened 2022-09-16
Make the sidebar nav collapsible on desktop

e111077:
Check it out!


aomarks (approved):
Tested on Chrome, Firefox, Safari. I think it looks great!

PR #946, opened 2022-09-17
implement a search modal

e111077:
Changes:

- litdev-search-modal: The search button and dialog
  - A new modal that uses native dialog
  - The SSR'd part is now just a button rather than an input
  - Can be opened with cmd+k (this unfortunately lives in the hydrate-common-components.js file)
  - icons in mobile nav are now an icon bar
  - Hydration is a bit more wonky now that it needs to find the right modal and open it when clicked before hydration or else it'll be invisible
- litdev-search: The search input + options groups
  - no longer looks like an underlined material input but rather a big square
  - We now ask algolia to return the highlight result and snippet results of text title and heading
  - Now renders search options grouped by document
- litdev-search-option: the individual search result
  - Now if it is a:
    - link to document, renders only the title and a document icon
    - subsection link: renders the subsection title, and relevant text, and a hashtag icon
  - no longers renders the type of content tag inside the option because it's a part of the group title


AndrewJakubowicz (commented):
This is amazing and I love it! Some minor things from an initial manual testing pass.
Did some browser and accessibility testing - Mostly comparing to docs.astro.build and Vue that also have search modals.

1. Our search modal doesn't currently allow tabbing down the results which was the first thing I tried when keyboard navigating. Side note - I do like how Algolia's search modal has keyboard controls on the bottom. Although - our modal is a combobox so maybe tab navigation doesn't make sense?

2. I tested VoiceOver on Mac and the combobox was great to use on Chrome and Firefox. But on Safari it doesn't work at all - which surprised me. Nothing is read out, not even when using arrow keys over choices. They are not read out. Even following the VoiceOver instructions result in nothing being read.

3. If a user resizes their screen while the search modal is open into the mobile view, the page stops being responsive. I think the modal overlay is still secretly there blocking interactions. Reproduce by: Opening search modal in desktop view. Shrink window until mobile nav appears. Then notice that the page cannot be clicked on even though it looks like it should be interactive (modal overlay is gone). After pressing `ESC` the page becomes interactive.

4. You can break the desktop search by also doing the reverse as (3.). Start with a small screen on mobile nav. Open the search and type something in. Now resize the page back to desktop view. Observe that the mobile drawer is open at the same time as the top nav. After using `ESC` to close the mobile drawer everything seems fine (on Chrome). However on Firefox the desktop search icon no longer works (via click or keyboard `Enter` keypress).

Screen Shot 2022-09-22 at 8 37 55 AM

Styles on desktop work great across all the browsers.

e111077:
1. Lack of tabbing is because it's a combobox. Do you think the navigation instructions are a must?
2. Filed an issue on Webkit https://bugs.webkit.org/show_bug.cgi?id=245549
3. Fixed this by adding event listener to window
4. addressed

packages/lit-dev-content/src/global/hydrate-common-components.ts
```diff
@@ -63,6 +49,72 @@ const activateMobileDetails = async () => {
...
+// There are 2 modal buttons on the page. In the header and in the mobile nav.
+// We need to open the one in the current view or else it'll open a hidden
+// modal that makes the dom inert.
+const searchModals = document.querySelectorAll(
+  'litdev-search-modal'
+) as NodeListOf<LitDevSearchModal>;
+let modalHydrated = false;
+
+const hydrateSearchModal = async () => {
+  if (modalHydrated) return;
```
AndrewJakubowicz, line 63: Does `modalHydrated` need to be set to true below this line?

packages/lit-dev-content/src/components/litdev-search-option.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  render() {
+    return html`
+      <div class="suggestion">
+        <div aria-hidden="true">
+          ${this.isSubsection ? hashtagIcon : paperDocumentIcon}
+        </div>
+        <div class="title-and-text">
+          ${this.isSubsection
+            ? html`<span class="title">${unsafeHTML(this.heading)}</span>
```
AndrewJakubowicz, line 109: Is `unsafeHTML` required? It creates a possible security vulnerability. I think this case can be manually sanitized and have `unsafeHTML` removed if we know what tags algolia uses for the emphasis.

E.g.: https://lit.dev/playground/#gist=93d0e481de4d628db59d8edda1daf53c

AndrewJakubowicz (commented):
Sorry for the slow review. Incrementally working through this change. Some more minor comments.

packages/lit-dev-content/src/global/hydrate-common-components.ts
e111077: Excellent catch!

packages/lit-dev-content/src/components/litdev-search-option.ts
e111077: Great suggestion!

packages/lit-dev-content/src/components/litdev-search-modal.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  updated(changed: PropertyValues<this>) {
+    super.updated(changed);
+
+    if (changed.has('open')) {
+      this.showDialog(this.open);
+    }
+  }
+
+  connectedCallback(): void {
```
AndrewJakubowicz, line 98: Nit: needs cleanup in `disconnectedCallback`.

packages/lit-dev-content/src/components/litdev-search-option.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  @property({type: Boolean, attribute: true})
+  checked = false;
+
+  static styles = css`
+    :host {
+      display: block;
+    }
+
+    :host([checked]) :is(.suggestion, svg) {
```
AndrewJakubowicz, line 41: Woah! TIL! `:is` is neat.

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -41,26 +30,52 @@ interface UserFacingPageData {
...
-  isSubsection: boolean
-): TemplateResult {
-  if (isSubsection) {
-    return html`<div class="title-and-header">
-      <span class="title">${title}</span><span class="header">${heading}</span>
-    </div>`;
-  }
-  return html`<span class="header">${title}</span>`;
+interface SuggestionType {
+  tag: string;
```
AndrewJakubowicz, line 58: Can this be made a stricter type (which can also be used in the `SuggestionGroup`.

`tag: 'other' | 'api' | 'docs' | 'article'`

packages/lit-dev-content/src/global/hydrate-common-components.ts
```diff
@@ -63,6 +49,73 @@ const activateMobileDetails = async () => {
...
+  if (e.key === 'k' && (e.ctrlKey || e.metaKey)) {
+    e.preventDefault();
+    const isMobile = window.matchMedia('(max-width: 864px)').matches;
+    const navBar = isMobile ? 'mobileSiteNav' : 'desktopNav';
+
+    // We need to make sure we open the currently visible modal or else it'll
+    // open a hidden modal that makes the dom inert.
+    const searchModal = document.querySelector(
+      `#${navBar} litdev-search-modal`
+    ) as LitDevSearchModal;
```
AndrewJakubowicz, line 104: (optional) It may be worth explicitly also handling `undefined` and warning or throwing so that when developing we can quickly pinpoint the issue. Although line 112 would throw a runtime error.

packages/lit-dev-content/src/util/render-algolia-suggestions.ts
```diff
@@ -0,0 +1,32 @@
...
+  snippet: string
+): TemplateResult | string => {
+  // Algloia snippets have stringified HTML like so:
+  // 'BAz <em>Foo</em> bar <em>Foo</em>z'.
+  // It is always an <em>
+  const match = snippet.match(/<em>(.*?)<\/em>/);
+  if (!match || match.index === undefined) {
+    return snippet;
+  }
+  const fullMatch = match[0]; // Baz <em>Foo</em> bar <em>Foo</em>
```
AndrewJakubowicz, line 26: Nit `match[0]` will only return `<em>Foo</em>` (in the comment)

packages/lit-dev-content/src/util/render-algolia-suggestions.ts
```diff
@@ -0,0 +1,32 @@
...
+  const emContent = match[1]; // Foo
+  const start = snippet.slice(0, match.index); // Baz
+  const end = snippet.slice(match.index + fullMatch.length); // bar <em>Foo</em>z
+
+  return html`${start}<em>${emContent}</em>${renderAlgoliaSnippet(end)}`;
```
AndrewJakubowicz, line 31: This recursion is great!

AndrewJakubowicz (approved):
Wow! Awesome!

packages/lit-dev-content/src/components/litdev-search.ts
e111077: I plan to pull this out in an upcoming PR where I do this work on the indexer level. Can we punt for now?

packages/lit-dev-content/src/global/hydrate-common-components.ts
e111077: Personally I'd rather it throw the error so we know to remove this code if it no longer exists

packages/lit-dev-content/src/util/render-algolia-suggestions.ts
e111077: Your suggestions were great!

packages/lit-dev-content/src/components/litdev-search.ts
e111077: nvm just did it anyway

PR #947, opened 2022-09-18
Fix markdown link for "primitive values"

timpulver:
This fixes a markdown link, which used the wrong brackets and was therefore not displaying correctly.

Screenshot 2022-09-18 at 13 47 17


aomarks (approved):
Thanks for the fix!

PR #953, opened 2022-09-23
Add heading for renderable values

arthurevans:
This is inspired by the changes that @bennypowers suggested in #935 ... This is basically the same thing with some changes that seemed like too fiddly to work through in GH suggestions.


arthurevans:
Ping @justinfagnani @sorvell on this small change.

packages/lit-dev-content/site/docs/templates/expressions.md
```diff
@@ -137,11 +137,11 @@ html`<main>${bodyText}</main>`
 
 Expressions in the child position can take many kinds of values:
 
-* Primitive values
-* Sentinel values
-* `TemplateResult` objects created with the `html` function.
-* DOM nodes
-* Arrays or iterables of any of the supported types
+* Primitive values likes strings, numbers, and booleans.
+* `TemplateResult` objects created with the [`html`](/docs/api/templates/#html) or [`svg`](/docs/api/templates/#svg) functions.
```
sorvell, line 141:
```suggestion
* `TemplateResult` objects created with the [`html`](/docs/api/templates/#html) or [`svg`](/docs/api/templates/#svg) functions, when inside an `svg` element.
```

Seems most important to call this out here. I'm meh on putting it in the other spot as well.

sorvell (approved):
Just a minor suggestion, but overall looks good to me.

justinfagnani (approved):

packages/lit-dev-content/site/docs/templates/expressions.md
arthurevans: Like the idea, but this way it's not clear that "when inside an `svg` element" applies only to the `svg` function. Maybe:

```suggestion
* `TemplateResult` objects created with the [`html`](/docs/api/templates/#html) function (or the [`svg`](/docs/api/templates/#svg) function, if the expression is inside an `<svg>` element).
```

PR #956, opened 2022-09-28
Associate children with parents in search index

e111077:


packages/lit-dev-content/src/components/algolia-search-controller.ts
```diff
@@ -18,7 +18,8 @@ const agloliaSearchControllerDefaultOptions = {
   appId: publicVars.algolia.appId,
   searchOnlyKey: publicVars.algolia.searchOnlyKey,
   index: publicVars.algolia.index,
-  hitsPerPage: 20,
+  hitsPerPage: 10,
+  distinct: 4 as number | boolean,
```
AndrewJakubowicz, line 22: Is this type assertion required?

packages/lit-dev-tools-cjs/src/search/plugin.ts
```diff
@@ -15,16 +15,35 @@ import {PageSearchChunker} from './PageSearchChunker.js';
 type UrlToFile = Map<string, string>;
 
 /**
- * Data to be Indexed.
+ * Generic that describes the type of document.
+ */
+interface DocType<T extends string, U extends string> {
```
AndrewJakubowicz, line 20: A little unfortunate this type is duplicated.

AndrewJakubowicz (approved):

packages/lit-dev-content/src/components/litdev-search.ts
```diff
@@ -160,6 +119,8 @@ export class LitDevSearch extends LitElement {
     this,
     () => this._searchText,
     {
+      // DO NOT SUBMIT
+      index: 'lit_dev_test',
```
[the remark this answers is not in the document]
e111077: Deploy currently is an action that runs on commit to `main`. It then runs a GH action that upserts the entire production index. I did not update that file to upsert the `lit_dev_test`. Currently I'm just manually updating it for the sake of these demo links working. But once this is submitted, it should update algolia automatically with the new schema.

There will be issues in those minutes where algolia is updated but lit.dev cloud run has not deployed yet. And it seems like cloud run is much slower than algolia upload, but at least it's automatic

packages/lit-dev-content/src/components/algolia-search-controller.ts
e111077: yes, because distinct can be false. We will be doing this in the SERP PR, and we're inferring the type of the constructor's `options` argument from this variable

packages/lit-dev-tools-cjs/src/search/plugin.ts
e111077: Agreed. Can't think of a good way to assert that they're both the same since they're separate projects, and importing across projects seems to cause unrelated TS failures. I'm not totally sure that this is too much of an issue since I don't really see us making changes to these types too often

packages/lit-dev-content/src/components/algolia-search-controller.ts
augustjk: In the future, perhaps this whole default options object can be asserted with `satisfies`.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-9-beta/#hamilton

(also great that the hash link for `satisfies` operator section is `#hamilton`)

packages/lit-dev-tools-cjs/src/search/plugin.ts
augustjk: I think a comment that mentions these should be in sync would be nice.

augustjk (commented):
Apologies for the later review. Nothing critical.

PR #966, opened 2022-09-29
Fix localization runtime docs typo

ca-d:
The exported symbol is called `sourceLocale`, not `sourceLocales`.


aomarks (approved):
Thank you

PR #968, opened 2022-10-05
Docs on handling rejected updateComplete. Fixes #649.

### LINKED ISSUE #649: Document handling rejected `updateComplete` promise
When https://github.com/Polymer/lit-element/pull/607 is merged, the `updateComplete` promise will be able to be rejected. We need to add documentation explaining that these exceptions need to be handled in any code that awaits the `updateComplete` promise. This might go in the promises [section](https://lit-element.polymer-project.org/guide/lifecycle) of the lifecycle.

arthurevans:


packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -330,6 +330,28 @@ async _loginClickHandler() {
 
 Also, when writing tests you can await the `updateComplete` Promise before making assertions about the component’s DOM.
 
+The `updateComplete` promise rejects if there's an unhandled error during the update cycle. For more information, see [Handling errors in the update cycle](#errors-in-the-update-cycle).
+
+### Handling errors in the update cycle {#errors-in-the-update-cycle}
+
+An uncaught exception in the update cycle causes the `updateComplete` promise to reject. If you're awaiting the `updateComplete` promise, you can handle these errors inline:
```
justinfagnani, line 337: It might good to note when there might be an unhandled error - Lit code doesn't have too many places to throw, but user code in render(), update(), etc., can throw and ReactiveElement will try to route that into the updateComplete promise.

So maybe:

```suggestion
An uncaught exception in the update cycle, which can be caused by exceptions in lifecycle methods like `render()` or `update()`, causes the `updateComplete` promise to reject. If you're awaiting the `updateComplete` promise, you can handle these errors inline:
```

packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -330,6 +330,28 @@ async _loginClickHandler() {
...
+
+```js
+try {
+  await this.updateComplete;
+} catch (e) {
+  /* handle error */
+}
+```
+
+In many cases, you won't need to use the `updateComplete` promise. Rather than adding extra code to every component, you can add a handler for `window.onunhandledrejection`. This lets you log errors that you might otherwise miss.
```
justinfagnani, line 347: Those errors can usually be caught in try/catch in the lifecycle methods themselves - it's usually a good practice to make those not throw.

justinfagnani (approved):

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: OK. Should I still document the `onunhandledrejection`? Maybe as a fallback to use, for example, to report unhandled rejections back to a server for diagnostic purposes?

PR #972, opened 2022-10-07
Fix typo in docs/components/properties.md

38elements:
There is syntax error. This fixed it.


augustjk (approved):
Nice catch! Thank you for the fix.

PR #976, opened 2022-10-10
Document scheduleUpdate. Fixes #457.

### LINKED ISSUE #457: Update lifecycle docs to include `sheduleUpdate`
When https://github.com/lit/lit/issues/2091 lands.

arthurevans:
Fixes #457.

### DIFF
--- a/packages/lit-dev-content/site/docs/components/lifecycle.md
+++ b/packages/lit-dev-content/site/docs/components/lifecycle.md
@@ -383,7 +383,7 @@ async scheduleUpdate() {
 
 {% endswitchable-sample %}
 
-If you override `scheduleUpdate()`, it's your responsibility to call `super.scheduleUpdate()` to continue the update.
+If you override `scheduleUpdate()`, it's your responsibility to call `super.scheduleUpdate()` to perform the pending update.
 
 {% aside "info" %}
 
--- a/packages/lit-dev-content/site/docs/components/lifecycle.md
+++ b/packages/lit-dev-content/site/docs/components/lifecycle.md
@@ -361,7 +361,7 @@ This section covers some less-common methods for customizing the update cycle.
 
 #### scheduleUpdate() {#scheduleupdate}
 
-Override `scheduleUpdate()` to customize the timing of the update. `scheduleUpdate()` is called when an update is about to be performed, and by default it calls `performUpdate()` immediately. Override it to defer the update. 
+Override `scheduleUpdate()` to customize the timing of the update. `scheduleUpdate()` is called when an update is about to be performed, and by default it calls `performUpdate()` immediately. Override it to defer the update—this technique can be used to unblock the main rendering/event thread. 
 
 For example, the following code schedules the update to occur after the next frame paints, which can reduce jank if the update is expensive:
 
@@ -370,20 +370,20 @@ For example, the following code schedules the update to occur after the next fra
 ```ts
 protected override async scheduleUpdate(): Promise<void> {
   await new Promise((resolve) => setTimeout(resolve));
-  this.performUpdate();
+  super.scheduleUpdate();
 }
 ```
 
 ```js
 async scheduleUpdate() {
   await new Promise((resolve) => setTimeout(resolve));
-  this.performUpdate();
+  super.scheduleUpdate();
 }
 ```
 
 {% endswitchable-sample %}
 
-If you override `scheduleUpdate()`, it's your responsibility to call `performUpdate()`.
+If you override `scheduleUpdate()`, it's your responsibility to call `super.scheduleUpdate()` to continue the update.
 
 {% aside "info" %}
 
@@ -393,13 +393,12 @@ This example shows an [async function](https://developer.mozilla.org/en-US/docs/
 
 {% endaside %}
 
-This technique can be used to unblock the main rendering/event thread. See the Chrome Dev Summit talk by Justin Fagnani [The Virtue of Laziness](https://www.youtube.com/watch?v=ypPRdtjGooc) for an extended discussion. This talk shows an early version of Lit, but describes many useful concepts.
 
 #### performUpdate()  {#performupdate}
 
 Implements the reactive update cycle, calling the other methods, like `shouldUpdate()`, `update()`, and `updated()`.
 
-Call `performUpdate()` to immediately process a pending update. This should generally not be needed, but it can be done in rare cases when you need to update synchronously.
+Call `performUpdate()` to immediately process a pending update. This should generally not be needed, but it can be done in rare cases when you need to update synchronously. (If there is no update pending, you can call `requestUpdate()` followed by `performUpdate() to force a synchronous update.)
 
 {% aside "info" %}
 
--- a/packages/lit-dev-content/site/docs/components/lifecycle.md
+++ b/packages/lit-dev-content/site/docs/components/lifecycle.md
@@ -361,23 +361,22 @@ This section covers some less-common methods for customizing the update cycle.
 
 #### scheduleUpdate() {#scheduleupdate}
 
-Override `scheduleUpdate()` to customize the timing of the update cycle. `scheduleUpdate()` is called when you request an update, and by default it calls `performUpdate()` immediately. Override it to defer the update. For example, you could use it to coordinate update timing between multiple elements.
+Override `scheduleUpdate()` to customize the timing of the update. `scheduleUpdate()` is called when an update is about to be performed, and by default it calls `performUpdate()` immediately. Override it to defer the update. 
 
-The following code schedules the update to occur just before the next frame:
+For example, the following code schedules the update to occur after the next frame paints, which can reduce jank if the update is expensive:
 
 {% switchable-sample %}
 
 ```ts
 protected override async scheduleUpdate(): Promise<void> {
-  await new Promise<void>((resolve) => 
-      requestAnimationFrame(() => resolve()));
+  await new Promise((resolve) => setTimeout(resolve));
   this.performUpdate();
 }
 ```
 
 ```js
 async scheduleUpdate() {
-  await new Promise((resolve) => requestAnimationFrame(() => resolve()));
+  await new Promise((resolve) => setTimeout(resolve));
   this.performUpdate();
 }
 ```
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -357,22 +357,79 @@ window.onunhandledrejection = function(e) {
...
+        resolve()
+      )
+  );
 }
 ```
 
-In this example, the update is performed after paint. This technique can be used to unblock the main rendering/event thread. See the Chrome Dev Summit talk by Justin Fagnani [The Virtue of Laziness](https://www.youtube.com/watch?v=ypPRdtjGooc) for an extended discussion.
+{% endswitchable-sample %}
+
+This technique can be used to unblock the main rendering/event thread. See the Chrome Dev Summit talk by Justin Fagnani [The Virtue of Laziness](https://www.youtube.com/watch?v=ypPRdtjGooc) for an extended discussion. This talk shows an early version of Lit, but describes many useful concepts.
```
justinfagnani, line 418: I feel like this maybe could be merged with the intro above to shorten the whole section.

packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -357,22 +357,79 @@ window.onunhandledrejection = function(e) {
...
+Override `scheduleUpdate()` to customize the timing of the update cycle. `scheduleUpdate()` is called when you request an update, and by default it calls `performUpdate()` immediately. Override it to defer the update. For example, you could use it to coordinate update timing between multiple elements.
+
+The following code schedules the update to occur just before the next frame:
+
+{% switchable-sample %}
+
+```ts
+protected override async scheduleUpdate(): Promise<void> {
+  await new Promise<void>((resolve) => 
+      requestAnimationFrame(() => resolve()));
```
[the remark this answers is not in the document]
arthurevans: When I tried to use just `Promise` TypeScript complained: "Generic type 'Promise<T>' requires 1 type argument(s)."

Or did you just mean in the "await new Promise()" line? Looks like I can safely leave it out there.

packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -357,22 +357,79 @@ window.onunhandledrejection = function(e) {
...
 ### Implementing additional customization {#reactive-update-cycle-customizing}
 
-#### performUpdate()  {#performupdate}
+This section covers some less-common methods for customizing the update cycle.
 
-Implements the reactive update cycle, calling the other methods, like `shouldUpdate()`, `update()`, and `updated()`.
+#### scheduleUpdate() {#scheduleupdate}
 
-Call `performUpdate()` to immediately process a pending update. This should generally not be needed, but it can be done in rare cases when you need to update synchronously.
+Override `scheduleUpdate()` to customize the timing of the update cycle. `scheduleUpdate()` is called when you request an update, and by default it calls `performUpdate()` immediately. Override it to defer the update. For example, you could use it to coordinate update timing between multiple elements.
```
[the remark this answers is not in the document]
arthurevans: I think this is a little tricky, because we don't talk about enqueueUpdate() ... We only talk about `requestUpdate()` and `performUpdate()`, and `scheduleUpdate()` comes right in the middle. Nevertheless, rephrased for now. If we re-do the diagrams, might add "enqueue update" as a fuzzy step between the "request and update" and "perform an update" diagrams.

packages/lit-dev-content/site/docs/components/lifecycle.md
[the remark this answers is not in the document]
arthurevans: Replaced example.

packages/lit-dev-content/site/docs/components/lifecycle.md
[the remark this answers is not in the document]
arthurevans: Done.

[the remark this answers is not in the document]
arthurevans: Updated.

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: I played around with this, but putting it up top kind of suggests that you need to go watch the video to understand the section. The video has some good information, but it's a little confusing as an intro because it covers several topics (including shceduleUpdate) and it uses older versions of lit-html and lit-element. (I think it also mentions an add-on library.)

packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -357,22 +357,79 @@ window.onunhandledrejection = function(e) {
...
+```
+
+{% endswitchable-sample %}
+
+If you override `scheduleUpdate()`, it's your responsibility to call `performUpdate()`.
+
+In an async function, you can `await` any promises, then call `performUpdate()` to complete the current update. The *next* update can't proceed until `scheduleUpdate()` returns.
 
-Implement `performUpdate()` to customize the timing of the update cycle. This can be useful for implementing custom scheduling. Note, if `performUpdate()` returns a Promise, the `updateComplete` Promise will await it.
+You can also write `scheduleUpdate()` as a regular function. In this case, `scheduleUpdate()` should return a promise. When resolving the promise, you must call `performUpdate()` to complete the current update. The *next* update can't proceed until the promise resolves.
```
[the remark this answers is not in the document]
arthurevans: Moved this into an aside, added link.

packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -357,22 +357,57 @@ window.onunhandledrejection = function(e) {
...
+async scheduleUpdate() {
+  await new Promise((resolve) => setTimeout(resolve));
+  this.performUpdate();
 }
 ```
 
-In this example, the update is performed after paint. This technique can be used to unblock the main rendering/event thread. See the Chrome Dev Summit talk by Justin Fagnani [The Virtue of Laziness](https://www.youtube.com/watch?v=ypPRdtjGooc) for an extended discussion.
+{% endswitchable-sample %}
+
+If you override `scheduleUpdate()`, it's your responsibility to call `performUpdate()`.
```
justinfagnani, line 386: I think a more robust recommendation is to return `super.scheduleUpdate()` (which just does `this.performUpdate()`) because that would work in mixins too, or if we ever made the default scheduleUpdate do more.

So:

```js
async scheduleUpdate() {
  await new Promise((resolve) => setTimeout(resolve));
  return super.scheduleUpdate();
}
```

packages/lit-dev-content/site/docs/components/lifecycle.md
```diff
@@ -357,22 +357,57 @@ window.onunhandledrejection = function(e) {
...
+
+{% endaside %}
+
+This technique can be used to unblock the main rendering/event thread. See the Chrome Dev Summit talk by Justin Fagnani [The Virtue of Laziness](https://www.youtube.com/watch?v=ypPRdtjGooc) for an extended discussion. This talk shows an early version of Lit, but describes many useful concepts.
+
+#### performUpdate()  {#performupdate}
+
+Implements the reactive update cycle, calling the other methods, like `shouldUpdate()`, `update()`, and `updated()`.
+
+Call `performUpdate()` to immediately process a pending update. This should generally not be needed, but it can be done in rare cases when you need to update synchronously.
```
justinfagnani, line 402: There's a trick here where you need to call `this.isUpdatePending = true` first to get `performUpdate()` to do anything. cc @sorvell

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: I think it was @kevinpschaaf who suggested that we use `performUpdate()` directly because it was clearer. I could go either way.

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: Hmm. That did not seem to be the case when I tested the examples, but I may have missed something.

arthurevans: Hmm... Interesting point. It says "pending update" ... If someone's enqueued an update, but then deferred it in `scheduleUpdate()`, then `isUpdatePending` will be true and calling `performUpdate()` processes the current pending update (which should set `isPendingUpdate` to false, and cause the eventual, scheduled update to be a no-op. Is that a realistic scenario?

What are the other cases where you could have a pending update but _not_ have `isPendingUpdate()` be true?

arthurevans: Also, looking at the current docs for `isUpdatePending`, it says the value should only be read. So if we're going to document this trick here, we should probably also update the API docs.

https://lit.dev/docs/api/ReactiveElement/#ReactiveElement.isUpdatePending

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: Got it. I can do that.

arthurevans:
Waiting for feedback from @kevinpschaaf and @sorvell on the two open questions above.

packages/lit-dev-content/site/docs/components/lifecycle.md
kevinpschaaf: > because that would work in mixins too

Hmm, yeah I guess. I just think it's really unintuitive semantically that `super.scheduleUpdate()` _does_ the update.

packages/lit-dev-content/site/docs/components/lifecycle.md
kevinpschaaf: I think the current words are correct: "tim immediately process a _pending update_. Could maybe just add "If no update is pending, this method does nothing." And if users _want_ it to do something, just call `requestUpdate()` before `performUpdate()` (instead of any suggestion they should set the flag).

packages/lit-dev-content/site/docs/components/lifecycle.md
justinfagnani: I dunno... I usually think that when you override a method, by default you should call `super.thatMethod()` as it does the correct base behavior.

Maybe this case is just different? It's not a huge deal either way.

packages/lit-dev-content/site/docs/components/lifecycle.md
justinfagnani: When I tried calling `performUpdate()` directly myself it didn't do anything since there was no update pending and the first thing `performUpdate()` does is check `isUpdatePending`: https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L1283

@sorvell was there another intended way to provoke an update? `this. requestUpdate(); this.performUpdate()` should also work.

arthurevans: I think it makes sense to keep `requestUpdate()` as the canonical way to trigger a pending update. `this.requestUpdate(); this.performUpdate();` (or `this.reactiveProperty = newValue; this.performUpdate();`) can be used if you really need to force a synchronous update. But how often do you need to do this?

I looked for examples in the MWC web components and Shoelace; I didn't see anyone calling `this.performUpdate()` explicitly. On a wider search in GitHub, I saw some examples. Most of these appear to be unnecessary (for example, components that call `performUpdate()` after every property change). 

One component uses the pattern change property, `performUpdate()`, dispatch event, which might make sense... although arguably you could do the same thing by dispatching the event later, from `updated()`, say.

packages/lit-dev-content/site/docs/components/lifecycle.md
sorvell: I'm inclined to agree with @justinfagnani but similarly don't think it's a huge deal either way =)

packages/lit-dev-content/site/docs/components/lifecycle.md
sorvell: As noted, the current way to ensure that you "force an update" is with 2 calls `this.requestUpdate(); this.performUpdate()`. I believe we've previously considered this a rough spot in the API and that we could add a `forceUpdate` method to make it simpler, but my recollection is that we were holding off on that pending feedback on the need for it.

sorvell (approved):
LGTM; added my opinion on the 2 open questions, but I think neither is blocking.

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: SG. Changed back to `super.scheduleUpdate()`.

packages/lit-dev-content/site/docs/components/lifecycle.md
arthurevans: Updated to add mention of `requestUpdate()` to force an update.

justinfagnani (approved):

PR #983, opened 2022-10-25
bump playwright browser version

e111077:
This will bump browser version which are currently failing on `main` because of #982 

Chrome version is now 106. Pinned it to this version due to an elusive bug in very specific versions of Chromium 107

https://discord.com/channels/1012791295170859069/1031947694224199740


e111077:
I legit can't believe I finally got tests to pass on this 😭😭😭

PR #984, opened 2022-10-25
Note on boolean attributes. Fixes #744.

### LINKED ISSUE #744: Document that Boolean attributes should never be initialized to true if they should be configurable from markup
### Context

This is an area of confusion and we don't have clear documentation on lit.dev relating to this case.

Initializing a boolean attribute to `true` results in confusing behavior, as you cannot declaratively remove the attribute that has been defaulted to true from markup.

### Proposal

Add a warning or note to https://lit.dev/docs/templates/expressions/#boolean-attribute-expressions stating that boolean attributes should never be default initialized to true.

### References

This has been discussed in the past:

- Kevin's comment: https://github.com/Polymer/polymer/issues/1812#issuecomment-112226630
- Link to polymer documentation that explains the same issue: https://polymer-library.polymer-project.org/2.0/docs/devguide/properties#configuring-boolean-properties

Out of scope of this issue, but maybe this can be caught by the analyzer.

arthurevans:


AndrewJakubowicz (approved):
Nice!

PR #990, opened 2022-11-03
Fix docs/components/properties.md

38elements:
I think `requestUpdate()` is called, not mutated.

```js
this.myArray.splice(indexToRemove, 1);
this.requestUpdate();
```


arthurevans (approved):
That's clearer, thanks.

PR #991, opened 2022-11-08
Add section on not bundling, optimizing reusable components.

### LINKED ISSUE #898: [docs] Publishing best practice should mention not to bundle
Here: https://lit.dev/docs/tools/publishing/#publishing-best-practices

We mention not to import polyfills, but we should also mention not to bundle/minify when publishing components to NPM. Perhaps also worth noting, if wanting to support consumption via CDN, create a bundle/package for it separate from what's npm installed?

arthurevans:
Fixes #898.

### DIFF
--- a/packages/lit-dev-content/site/docs/tools/publishing.md
+++ b/packages/lit-dev-content/site/docs/tools/publishing.md
@@ -118,7 +118,7 @@ they're needed, they should only go in `devDependencies`.
 
 ### Don't bundle, minify, or optimize modules
 
-Bundling and other optimizations are application concerns. Bundling a reusable component before publishing to npm can also introduce multiple versions of Lit (and other packages) into your users' application since npm can't deduplicate the packages. This causes bloat and may cause bugs.
+Bundling and other optimizations are application concerns. Bundling a reusable component before publishing to npm can also introduce multiple versions of Lit (and other packages) into a user's application since npm can't deduplicate the packages. This causes bloat and may cause bugs.
 
 Optimizing modules before publication may also prevent application-level optimizations.
 
--- a/packages/lit-dev-content/site/docs/tools/publishing.md
+++ b/packages/lit-dev-content/site/docs/tools/publishing.md
@@ -118,11 +118,11 @@ they're needed, they should only go in `devDependencies`.
 
 ### Don't bundle, minify, or optimize modules
 
-Bundling and other optimizations are application concerns. Bundling can also introduce multiple versions of Lit (and other packages) into your application since npm can't deduplicate the packages. This causes bloat and may cause bugs.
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/tools/publishing.md
```diff
@@ -106,16 +106,26 @@ You can run Babel via a bundler plugin such as [@rollup/plugin-babel](https://ww
...
 Packages may need to depend on polyfills for tests and demos, so if
 they're needed, they should only go in `devDependencies`.
 
+### Don't bundle, minify, or optimize modules
+
+Bundling and other optimizations are application concerns. Bundling can also introduce multiple versions of Lit (and other packages) into your application since npm can't deduplicate the packages. This causes bloat and may cause bugs.
+
+Optimizing modules before publication may also prevent application-level optimizations.
+
+Bundling and other optimizations can be valuable when serving a module from a CDN. Because of the potential issues with bundling, this should usually be used for prototyping, not for production.
```
kevinpschaaf, line 125: There's something about these two sentences that almost read as contradictory. Bundling is valuable. Bundling has issues. Feels like we want some more nuance.

```suggestion
Bundling and other optimizations can be valuable when serving a module from a CDN, but since users may need to use multiple packages that depend on Lit, serving from a CDN can result in users loading more code than necessary. For these reasons we recommend performance-sensitive applications always build from npm where packages can be deduplicated, rather than loading bundled packages off of a CDN.
```

kevinpschaaf (approved):
Tried adding some clarifications, feel free to take/leave/adapt the words.

packages/lit-dev-content/site/docs/tools/publishing.md
```diff
@@ -106,16 +106,26 @@ You can run Babel via a bundler plugin such as [@rollup/plugin-babel](https://ww
...
 the browsers the application needs to support, and that choice is best left to
-the application developer using your component.
+the application developer using your component. Your component's documentation **should** clearly identify any APIs it uses that may require polyfills. 
 
 Packages may need to depend on polyfills for tests and demos, so if
 they're needed, they should only go in `devDependencies`.
 
+### Don't bundle, minify, or optimize modules
+
+Bundling and other optimizations are application concerns. Bundling a reusable component before publishing to npm can also introduce multiple versions of Lit (and other packages) into your users' application since npm can't deduplicate the packages. This causes bloat and may cause bugs.
```
arthurevans, line 121:
```suggestion
Bundling and other optimizations are application concerns. Bundling a reusable component before publishing to npm can also introduce multiple versions of Lit (and other packages) into a user's application since npm can't deduplicate the packages. This causes bloat and may cause bugs.
```

PR #995, opened 2022-11-09
Fix vertical resizing on mobile.

sorvell:
* adds a default background color to the resize bar to make it more discoverable.
* makes the vertical resizing bar a bit bigger.
* adds `touch-action: none` to prevent the pointer from being cancelled.


sorvell:
Closed in favor of https://github.com/lit/lit.dev/pull/1001.

PR #998, opened 2022-11-11
Adopt wireit services and cascade:false

aomarks:
Simplifies the build by using the new `service: true` and `cascade: false` settings in the latest preview release of Wireit.

https://github.com/google/wireit#services
https://github.com/google/wireit#execution-cascade

The top-level `dev` script should now be more reliable, and automatically starts up the gist server emulator too.

Also makes the output a little quieter by using eleventy's `--quiet` flag, and removing some logging we didn't really need in the sample generation script.

Fixes https://github.com/lit/lit.dev/issues/994

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "start:main": "LITDEV_ENV=local MODE=main npm run start -w lit-dev-server",
     "start:playground": "LITDEV_ENV=local MODE=playground npm start -w lit-dev-server",
     "start:fake-github": "wireit",
-    "dev": "cd packages/lit-dev-server && npm run build && cd ../lit-dev-content && (npm run dev:build:assets --watch & npm run dev:build:eleventy:watch & node ../lit-dev-tools-esm/lib/dev-server.js --open)",
+    "dev": "npm run dev:serve --workspace lit-dev-content --watch",
     "format": "prettier . --write",
     "format:check": "prettier . --check",
     "nuke": "rm -rf node_modules packages/*/node_modules && npm ci",
@@ -54,6 +54,6 @@
     "playground-elements": "^0.17.0",
     "prettier": "^2.1.2",
     "typescript": "~4.6.2",
-    "wireit": "^0.6.1"
+    "wireit": "^0.7.3-pre.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -37,10 +37,10 @@
       ]
     },
     "start:fake-github": {
+      "command": "LITDEV_ENV=local node packages/lit-dev-tools-esm/lib/fake-github-server.js",
       "dependencies": [
         "./packages/lit-dev-tools-esm:build"
       ],
-      "command": "LITDEV_ENV=local node packages/lit-dev-tools-esm/lib/fake-github-server.js",
       "output": []
     },
     "publish-search-index": {
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
       "dependencies": [
         "./packages/lit-dev-tools-esm:build"
       ],
-      "output": []
+      "files": [],
+      "service": true
     },
     "publish-search-index": {
       "dependencies": [
--- a/packages/lit-dev-content/package.json
+++ b/packages/lit-dev-content/package.json
@@ -16,7 +16,6 @@
     "build:rollup": "wireit",
     "build:rollup:watch": "npm run build:rollup -- --watch",
     "build:samples": "wireit",
-    "build:samples:watch": "npm run build:samples -- --watch",
     "dev:build": "wireit",
     "dev:build:assets": "wireit",
     "dev:build:eleventy": "wireit",
--- a/packages/lit-dev-content/package.json
+++ b/packages/lit-dev-content/package.json
@@ -138,7 +138,7 @@
       "dependencies": [
         "../lit-dev-tools-esm:build:ts"
       ],
-      "command": "../../node_modules/.bin/tsc --build --pretty",
+      "command": "tsc --build --pretty",
       "clean": "if-file-deleted"
     },
     "build:rollup": {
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "start:main": "LITDEV_ENV=local MODE=main npm run start -w lit-dev-server",
     "start:playground": "LITDEV_ENV=local MODE=playground npm start -w lit-dev-server",
     "start:fake-github": "wireit",
-    "dev": "cd packages/lit-dev-server && npm run build && cd ../lit-dev-content && (npm run dev:build:assets --watch & npm run dev:build:eleventy:watch & node ../lit-dev-tools-esm/lib/dev-server.js --open)",
+    "dev": "npm run dev:serve --workspace lit-dev-content --watch",
```
e111077, line 17: 😍😍😍😍

packages/lit-dev-content/package.json
```diff
@@ -9,124 +9,88 @@
...
+      "service": true,
       "dependencies": [
-        "build:ts",
-        "build:samples",
-        "fonts:manrope"
+        "../lit-dev-tools-esm:build",
+        "../../:start:fake-github",
+        {
+          "script": "dev:build",
+          "cascade": false
```
e111077, line 70: Wonder if we'll ever have need for cascade to only cascade on a single dep or a subset of deps.

aomarks: Yeah like if we did SSR on the server live (as we are doing for webcomponents.org), then the JS for the server renderable components would need to restart the server potentially. In this case we don't need to.

e111077: What i'm thinking in that case would be it'd be cool if cascade could be of type `boolean|string[]` where the array would be `"cascade": ["build:ssr"]`

aomarks: It already works like that, but with a different syntax. In this case, "../../:start:fake-github" has default cascade:true, but "dev:build" has cascade:false.

e111077 (approved):
Awesome! Do you have any performance numbers here? Would be cool to track as a "Testimonial"

packages/lit-dev-content/package.json
aomarks: (Which made me realize start:fake-github can also be cascade, since it doesn't require a restart either)

e111077: So i'm imagining something like

```json
"server": {
  "command": "node server.js",
  "dependencies": [
    "build:ssr",
    "build:ts"
  ],
  cascade: [
    "build:ssr"
  ]
}
```

so if ts rebuilds the client-side TS it shouldn't _have_ to reload. Or unless I'm misunderstanding how cascade works across rules

e111077: also maybe duplicating rules across two fields like cascade and deps is bad...

aomarks: By default a dependency triggers reload (`cascade:true`). If you want a dependency to not trigger reload, you set `cascade:false` (on that specific dependency). Does that help? I think your example is an alternative syntax to what's already supported? Or maybe I am misunderstanding.

aomarks:
> Awesome! Do you have any performance numbers here? Would be cool to track as a "Testimonial"

I don't think performance will be much different, it's just a simpler and more reliable way to configure the same thing.

aomarks:
> > Awesome! Do you have any performance numbers here? Would be cool to track as a "Testimonial"
> 
> I don't think performance will be much different, it's just a simpler and more reliable way to configure the same thing.

Filed https://github.com/google/wireit/issues/525 to track an easy way to get this timing data for each wireit run, though.

packages/lit-dev-content/package.json
e111077: oh! okay I did misunderstand. So any services above it will not trigger?

e.g.

```
┌───────────┐         ┌─────────┐
│ Service 1 ├────────►│Service 2│
└───────────┘         └─┬───────┘
                        │
                        │
                        ├─────────────┐
                        │             │
                        │             │
                        │             │
         ┌──────────────▼──┐   ┌──────▼────────┐
         │       Rule 1    │   │      Rule 2   │
         │                 │   │               │
         │  cascade: false │   │ cascade: true │
         └─────────────────┘   └───────────────┘
```

if rule 1 changes then service 1 and 2 do not change, but if rule 2 changes they both do?

aomarks: Yep that's right!

And if you wanted for some reason e.g. rule 1 changes to cause service 1 to restart, but not service 2, then you'd also add a dependency between service 1 and rule 1, with cascade:false.

aomarks: (Edited above comment, had a typo)

e111077: uhh like this?

```
              ┌───────────┐         ┌─────────┐
              │ Service 1 ├────────►│Service 2│
              └────┬──────┘         └─┬───────┘
                   │                  │
                   │                  │
                   │                  ├─────────────┐
       ┌───────────┘                  │             │
       │                              │             │
       │                              │             │
┌──────▼────────┐      ┌──────────────▼──┐   ┌──────▼────────┐
│    Rule 1.5   │      │     Rule 1      │   │    Rule 2     │
│               ├─────►│                 │   │               │
│ cascade: true │      │  cascade: false │   │ cascade: true │
└───────────────┘      └─────────────────┘   └───────────────┘
```

aomarks: Kinda, but `cascade` is as annotation on the edge not the script, so it's simpler to think of it like this:

```
   ┌───────────┐         ┌─────────┐
   │ Service 1 ├────────►│Service 2│
   └────┬──────┘         └─┬────┬──┘
        │                  │    │
        │                  │    └────────┐
        │                  │             │
        │           [cascade:false]      │
        │                  │             │
        │   ┌──────────────▼──┐   ┌──────▼────────┐
        └──►│     Rule 1      │   │    Rule 2     │
            └─────────────────┘   └───────────────┘
```

PR #999, opened 2022-11-14
Fix typo _he next cycle_

nikolamicic21:
Spell check in components-lifecycle section - should be _the next cycle_


AndrewJakubowicz (approved):
Thank you!

PR #1004, opened 2022-11-17
Fix typo

nikolamicic21:
Should be _is separate**d**_ instead of _is separate_.


arthurevans:
Thanks, but "is separate from" is what we intended to say here. "Separated" would suggest that either they were once one, and are now divided; or that there is a thing dividing them ("separated by a wall"). Whereas here we're emphasizing the fact that they're two separate or distinct things.

nikolamicic21:
> Thanks, but "is separate from" is what we intended to say here. "Separated" would suggest that either they were once one, and are now divided; or that there is a thing dividing them ("separated by a wall"). Whereas here we're emphasizing the fact that they're two separate or distinct things.

Aah, my bad. Thanks for the explanation.

PR #1009, opened 2022-11-24
Add description about customizing event listener options in templates

38elements:
This resolves #1008.

PR #1017, opened 2022-12-03
Revert "Add description about customizing event listener options in templates"

e111077:
Reverts lit/lit.dev#1009

### DIFF
--- a/packages/lit-dev-content/site/docs/components/events.md
+++ b/packages/lit-dev-content/site/docs/components/events.md
@@ -42,14 +42,6 @@ private _handleTouchStart(e) { console.log(e.type) }
 
 </div>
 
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/components/events.md
```diff
@@ -42,14 +42,6 @@ private _handleTouchStart(e) { console.log(e.type) }
 
 </div>
 
-If you're not using decorators, you can customize event listener options by passing an object to the event listener expression. The object must have a `handleEvent()` method and can include any the options that would normally appear in the `options` argument to `addEventListener()`.
-
-```js
```
justinfagnani, line 47: Should this have had both JS and TS versions anyway, even if they're the same?

justinfagnani (approved):

packages/lit-dev-content/site/docs/components/events.md
e111077: don't think so. There seems to be another example in this page just like this

PR #1019, opened 2022-12-03
Fix example & spell check in content

MM25Zamanian:


arthurevans (changes requested):
As noted above, narrow to type fixes.

PR #1031, opened 2023-01-11
Add script and title to the set of elements where bindings aren't allowed

aomarks:


aomarks:
Nevermind, this seems to be wrong.

PR #1036, opened 2023-01-30
Update deps with npm update

justinfagnani:
Just runs `npm update` on the repo. So this doesn't update deps to new major version.

Dependabot PRs can't run all of our workflows. I wonder if we want to keep it turned on?


e111077:
uhh why is node_modules building... will TAL

PR #1037, opened 2023-01-31
Add context doc

justinfagnani:
This adds a *Components > Context* page.

I'm not sure that's the best place in the long term, but for the next single page in component concepts it makes sense. I think I'd rather have a *Managing Data* section that also includes state managers, some conceptual information about difference ways to get state, maybe Task, etc. We could try to add that section now with just *Context* in it.

### DIFF
--- a/packages/lit-dev-content/site/docs/components/context.md
+++ b/packages/lit-dev-content/site/docs/components/context.md
@@ -137,7 +137,7 @@ This means that there are two main ways to create a context object:
 1. With a value that is globally unique, like an object (`{}`)  or symbol (`Symbol()`)
 2. With a value that is not globally unique, so that it can be equal under strict equality, like a string (`'logger'`) or _global_ symbol (`Symbol.for('logger')`).
 
-If you want two _separate_ `createContext()` calls to referrer to the same
+If you want two _separate_ `createContext()` calls to refer to the same
 context, then use a key that will by equal under strict equality like a
 string:
 ```ts
--- a/packages/lit-dev-content/site/docs/components/context.md
+++ b/packages/lit-dev-content/site/docs/components/context.md
@@ -135,7 +135,7 @@ Context objects are used by providers to match a context request event to a valu
 
 This means that there are two main ways to create a context object:
 1. With a value that is globally unique, like an object (`{}`)  or symbol (`Symbol()`)
-2. With a value that is no globally unique, so that it can be equal under strict equality, like a string (`'logger'`) or _global_ symbol (`Symbol.for('logger')`).
+2. With a value that is not globally unique, so that it can be equal under strict equality, like a string (`'logger'`) or _global_ symbol (`Symbol.for('logger')`).
 
 If you want two _separate_ `createContext()` calls to referrer to the same
 context, then use a key that will by equal under strict equality like a
--- a/packages/lit-dev-content/site/docs/components/context.md
+++ b/packages/lit-dev-content/site/docs/components/context.md
@@ -7,6 +7,8 @@ eleventyNavigation:
   labs: true
 ---
 
+{% labs-disclaimer %}
+
 Context is a way of making data available to entire component subtrees without having to manually bind properties to every component. The data is "contextually" available, such that ancestor elements in between a provider of data and consumer of data aren't event aware of it.
 
 Lit's context implementation is part of [Lit Labs](/docs/libraries/labs/) and available in the `@lit-labs/context` package:
--- a/packages/lit-dev-content/site/docs/components/context.md
+++ b/packages/lit-dev-content/site/docs/components/context.md
@@ -19,7 +19,7 @@ npm i @lit-labs/context
 
 Context is useful for data that needs to be consumed by a wide variety and large number of components - things like an app's data store, the current user, a UI theme - or when data-binding isn't an option, such as when an element needs to provide data to its light DOM children.
 
-Context is very in ways to React's Context, or to dependency injection systems like Angular's, with some important differences that enable interoperability across different web components libraries, frameworks and plain JavaScript.
+Context is very similar to React's Context, or to dependency injection systems like Angular's, with some important differences that make Context work with the dynamic nature of the DOM, and enable interoperability across different web components libraries, frameworks and plain JavaScript.
 
 ## Example
 
@@ -58,7 +58,7 @@ Consumer:
 import {LitElement, property} from 'lit';
 import {consume} from '@lit-labs/context';
 
-import {type Logger, loggerContext} from './logger.js';
+import {type Logger, loggerContext} from './logger-context.js';
 
 export class MyElement extends LitElement {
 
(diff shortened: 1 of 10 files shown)


packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+
+Lit's context implementation is part of [Lit Labs](/docs/libraries/labs/) and available in the `@lit-labs/context` package:
+
+```bash
+npm i @lit-labs/context
+```
+
+Context is useful for data that needs to be consumed by a wide variety and large number of components - things like an apps data store, the current user, a UI theme - or when data-binding isn't an option, such as when an element needs to provide data to its light DOM children.
+
+Context is very in ways to React's Context, or to dependency injection systems like Angular's, with some important differences that enable interoperability across different web components libraries, frameworks and plain JavaScript.
```
graynorton, line 22: very *similar

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+  }
+}
+```
+
+Consumer:
+```ts
+import {LitElement, property} from 'lit';
+import {consume} from '@lit-labs/context';
+
+import {type Logger, loggerContext} from './logger.js';
```
graynorton, line 61: Seems like this should be importing from `'./logger-context.js`

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+
+### Themes
+
+Themes are sets of styles that apply to the entire page or entire subtrees within the page - exactly the kind of scope of data that context provides.
+
+One way of building a theme system would be to define a `Theme` type that containers can provide that holds named styles. Elements that want to apply a theme can consume the theme object and look up styles by name. Custom theme reactive controllers can wrap ContextProvider and ContextConsumer to reduce boilerplate.
+
+### HTML-based plugins
+
+Context can be used to provide data to light DOM children, which typically aren't created from the same declarative templates with databinding available.
```
kevinpschaaf, line 270: I know what you mean here, but it's hard to phrase in an understandable way.

Maybe
```suggestion
Context can be used to pass data from a parent to its light DOM children. Since the parent does not create the children, it cannot leverage template-based data-binding to pass data to its children, but can listen to and respond to `context-request` events.
```

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+For example, consider a code editor element with plugins for different language modes. You can make a plain HTML system for adding features using context:
+
+```html
+<code-editor>
+  <code-editor-javascript-mode></code-editor-javascript-mode>
+  <code-editor-python-mode></code-editor-python-mode>
+</code-editor>
+```
+
+In this case `<code-editor>` would provide an API for adding langage modes via context, and plugin elements would consume that API and add themselves to the editor.
```
kevinpschaaf, line 281: I don't super love this example; it's definitely valid, but could we choose one where both are visual (rather than the headless/utility element example)?

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+
+### Data formatters, link generators, etc.
+
+Sometimes reusable components will need to format data or URLs in an application-specific way. For example, a documentation viewer that renders a link to another item. The component will not know the URL space of the application.
```
kevinpschaaf, line 285: 😉

kevinpschaaf (approved):
Looks great. A couple nits and one-or-two stronger suggestions. Otherwise LGTM.

justinfagnani:
@kevinpschaaf thanks for the review!

This PR puts the doc at `/docs/components/context/`. I think there's still a big open question of whether it should go there or somewhere like `/docs/managing-data/context/` as I sketched out in https://github.com/lit/lit.dev/pull/1038

I'll address all the comments here, but not submit yet until we all decide on the final location. For me, the topics of Context, Task, state management integrations, and top-down data flow concepts, push towards a separate data section, fwiw.

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+
+```ts
+  private _myData = new ContextConsumer(this,
+    myContext,
+    undedined, /* callback */
+    true /* subscribe */
+  );
+```
+
+## Example Use Cases
```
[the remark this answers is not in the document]
justinfagnani: thanks!

packages/lit-dev-content/site/docs/components/context.md
justinfagnani: it's a great example!

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,497 @@
...
+**Signature**:
+
+```ts
+function createContext<ValueType, K = unknown>(key: K): Context<K, ValueType>;
+```
+
+
+Contexts are compared with with strict equality.
+
+If you want two separate `createContext()` calls to referer to the same context, then use a key that will by equal under strict equality like a string for `Symbol.for()`:
```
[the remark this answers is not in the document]
justinfagnani: I reworked this a bit. Hopefully it's a bit more clear.

packages/lit-dev-content/site/docs/components/context.md
justinfagnani: I think because we're not showing anything visual, that this might be easier to understand.

It is a case a came across a few times actually, where an editor maintainer said they didn't see the value in custom elements because you had to configure the editor in JS anyway.

packages/lit-dev-content/site/docs/components/context.md
```diff
@@ -0,0 +1,526 @@
...
+
+#### Context equality
+
+Context objects are used by providers to match a context request event to a value. Contexts are compared with with strict equality (`===`), so a provider will only handle a context request if its context key equals the context key of the request.
+
+This means that there are two main ways to create a context object:
+1. With a value that is globally unique, like an object (`{}`)  or symbol (`Symbol()`)
+2. With a value that is not globally unique, so that it can be equal under strict equality, like a string (`'logger'`) or _global_ symbol (`Symbol.for('logger')`).
+
+If you want two _separate_ `createContext()` calls to referrer to the same
```
arthurevans, line 140:
```suggestion
If you want two _separate_ `createContext()` calls to refer to the same
```

arthurevans: 👋 
Just dropping in to say hi and correct a typo for old time's sake 😂 ... Apparently I haven't turned off my GitHub notifications yet ...

arthurevans:
FWIW I like the idea of a separate data section. Given the other section titles are terse, mostly a single noun, "DATA" or "DATA FLOW" might fit better. I think putting the important word first makes it easier to scan (and we don't have any other section titles with gerunds).

justinfagnani:
I think we're going with a new section, as in https://github.com/lit/lit.dev/pull/1038

I'm not so sure about the title... I'm a little worried that just "Data" isn't descriptive enough to be inviting. "Data Flow" is a bit specific if we also talk about task and state management in there.

So I kept the "Managing Data" section label for now, but made the URL just `/docs/data/` so that we have some flexibility with the label going forward. If we wanted to change the URL too, we could with a redirect.

PR #1038, opened 2023-02-02
WIP: Sketch out a Managing Data section

justinfagnani:
This is an idea for a section that would talk about all the ways to manage data. It could include a lot of what's currently in the component composition section about data flow, context, task, and a page on state management integrations.


justinfagnani:
Created Managing Data section in #1037

PR #1041, opened 2023-02-05
Update typedoc

justinfagnani:
I'm doing this upgrade as a first step to try to get closer to API docs for labs packages like context, and because it'll likely be necessary at some point with TypeScript updates.  TypeDoc 0.23 seems to have some better monorepo support.

There were some breaking changes that I think I got through. I looked through some API pages, but not all, and they seem fine so far. Will look at more soon.

The built JSON files are quite different, so they're going to be hard to review.  I do wonder if we want stop regenerating the lit-html-1 and lit-element-2 data files so that we don't have to worry about changes to typedoc being incompatible with older jsdoc comment styles or older TypeScript source.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2713,6 +2713,7 @@
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
       "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
+      "dev": true,
       "dependencies": {
         "balanced-match": "^1.0.0",
         "concat-map": "0.0.1"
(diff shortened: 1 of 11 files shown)


justinfagnani:
So the *Check API data in sync* approach no longer works with typedoc@0.23 because the output isn't stable. There's a `target` property on a bunch of node that looks like an integer ID. If we want to make sure the file is roughly "the same" we'll have to do some custom processing. @aomarks

justinfagnani:
Somehow the `directives/` page isn't being generated.

aomarks:
> So the _Check API data in sync_ approach no longer works with typedoc@0.23 because the output isn't stable. There's a `target` property on a bunch of node that looks like an integer ID. If we want to make sure the file is roughly "the same" we'll have to do some custom processing. @aomarks

Can we just delete the `target` property?

justinfagnani:
@aomarks 
> Can we just delete the target property?

Maybe? If we don't use that to generate the docs. I'll try.

justinfagnani:
@aomarks I didn't realize we were already pruning so many properties. I added `'target'` to the list and we'll see if things render properly.

justinfagnani:
@aomarks that worked!

justinfagnani:
Right now the import locations are broken. They're rendering as: 

```ts
import { LitElement } from '';
```

fwiw, I was able to get some labs/contexts docs rendering as a follow-on.

AndrewJakubowicz:
This is superseded by https://github.com/lit/lit.dev/pull/1093 and also https://github.com/lit/lit.dev/pull/1093.

Thank you for providing a great reference to build on!

PR #1042, opened 2023-02-07
refactor(search): update search interface design

e111077:
Design wanted some changes to the search function's button and the searchbox


justinfagnani:
wdyt here @e111077 ?

I think the current search button is much more inviting to click and obviously search than a small button with jus an icon, regardless of whether it looks like an input and acts like a button.

e111077:
I think I like it overall, it's more concise, a custom icon that generally matches the sharpness of the rest of the site and a bit more professional. It also separates us a bit from the rest of the `algolia/instantsearch.js` apps out there like React, vue, and alpinejs. Meanwhile Angular and Preact look like textboxes but are actual textboxes

Also i like the searchbox much better in general.

e111077:
Also I've kind of always hated the MD search icon. It's like off to the side and makes it look like the whitespace is off or crooked to one side

augustjk:
The new icon looks more like a "Q" to me, or like a rotated power button. Took me a long time to realize it was supposed to be a magnifying glass.

e111077:
HAHA sorta looks like a record player if you mess with it a bit

search icon

e111077:
So is this a no to this PR?

e111077:
I want to put this back on the radar. I think the icon os a good design: it's sharp and in-line with the rest of the site's design. it looks like a magnifying glass or a "Q" for "question". It also has a `title="site search"` set and it is accessible. Additionally it looks more inline in mobile view:

| before | after |
| - | - |
| image | image|

justinfagnani:
@e111077 I personally don't think that the icon looks like a search icon anymore. It looks like a custom Q company logo.

I think that even though the current button looks like an input, it's at least very clear that it's for search. I might be ok with making it more button-like, and _maybe_ even removing the "Search" label (but to what end?), but I'm very dubious about innovating on the appearance of a standard and critical bit of functionality like search.

e111077:
Okay I went for a mix of the two. LMK what y'all think. I also implemented the search option design

e111077:
note the icon and the new search options
image

augustjk (approved):
I am okay with how this looks.

PR #1044, opened 2023-02-10
Add Tasks doc

justinfagnani:
Fixes: https://github.com/lit/lit.dev/issues/1191

Adds Task docs to 3.0.


packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+- Triggers a host update when the task changes status
+- Handles race conditions, ensuring that only the latest task invocation completes the task
+- Renders the correct template for the current task status
+
+This removes most of the boilerplate for correctly using async data from your code, and ensures robust handling of race conditions and other edge-cases.
+
+## Installation
+
+```bash
+npm i @lit-labs/task
```
e111077, line 70: should we put `-s`?

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+The task function takes the task arguments as an _array_ passed as the first parameter
+
+```ts
+new Task(this, {
+  task: async ([arg1, arg2]) => {
+    // do async work here
+  },
+  args: () => [this.field1, this.field2]
+})
+```
```
e111077, line 104: You never show or mention the shorthand of `new Task(this, ()=> {...}, () => [...])`. Should we mention it, or remove it from Task?

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+
+Write the `task` and `args` functions as arrow function so that the `this` reference points to the host element.
```
e111077, line 106: Maybe put this in a positive aside? https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/samples/tutorials/CONTRIBUTING.md#asides

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+
+When a task completes or otherwise changes status, it triggers a host update so the host can handle the new task status and render if needed.
+
+There are a few members on Task that represent the state of the task:
+- `value`: the current value of the task, if it has completed
+- `error`: the current error of the task, if it has errored
+- `status`: the status of the task. See [Task status](#task-status)
+- `render()`: A method that chooses a callback to run based on the current status.
+
+The simplest and most common API to use is `render()`, since it chooses the right code to run and provides it the relevant data.
```
e111077, line 118: This feels like a positive aside as well. Maybe can be rewritten a bit

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+        complete: (value) => html`<p>The task completed with: ${value}</p>`,
+        error: (error) => html`<p>Oops, something went wrong: ${error}</p>`,
+      })}
+    `;
+  }
+```
+
+### Running tasks
+
+By default, Tasks will run any time the arguments change. This is controlled by the `autoRun` option, which default to `true`.
```
e111077, line 143:
```suggestion
By default, Tasks will run any time the arguments change. This is controlled by the `autoRun` option, which defaults to `true`.
```

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+npm i @lit-labs/task
+```
+
+## Usage
+
+`Task` is a [reactive controller](../../composition/controllers/), so it can respond to and trigger updates to Lit's reactive update lifecycle.
+
+You'll generally have one Task object for each logical task that your component needs to perform. Install tasks as fields on your class:
+
+```ts
```
e111077, line 79: you use a private field. make it public or make it a switchable sample

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+      })}
+    `;
+  }
+```
+
+### Running tasks
+
+By default, Tasks will run any time the arguments change. This is controlled by the `autoRun` option, which default to `true`.
+
+In auto-run mode, the task will call the `args` function when the host has updated, compare the args to the previous args, and invoke the task function if they have changed.
```
e111077, line 145: you refer to `manual` mode in italics. Maybe mention the same for auto-run?

```suggestion
In _auto-run_ mode, the task will call the `args` function when the host has updated, compare the args to the previous args, and invoke the task function if they have changed.
```

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+
+If `autoRun` is set to false, the task will be in _manual_ mode. In manual mode you can run the task by calling the `.run()` method, possibly from an event handler:
+
+```ts
```
e111077, line 149: You use a private keyword. Can you please make this a switchable sample?

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+
+Tasks can be in one of four states:
+- `INITIAL`: The task has not been run
+- `PENDING`: The task is running and awaiting a new value
+- `COMPLETE`: The task completed successfully
+- `ERROR`: The task errored
+
+Usually a Task will proceed from INITIAL to PENDING to one of COMPLETE or ERROR, and then back to PENDING if the task is re-run.
+
+It's important to understand the status a task can be in, but it's not usually necessary to access it directly.
```
e111077, line 193: What is this line supposed to convey?

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Task status
+
+Tasks can be in one of four states:
+- `INITIAL`: The task has not been run
+- `PENDING`: The task is running and awaiting a new value
+- `COMPLETE`: The task completed successfully
+- `ERROR`: The task errored
+
+Usually a Task will proceed from INITIAL to PENDING to one of COMPLETE or ERROR, and then back to PENDING if the task is re-run.
```
e111077, line 191:
```suggestion
Usually a Task will proceed from `INITIAL` to `PENDING` to one of `COMPLETE` or `ERROR`, and then back to `PENDING` if the task is re-run.
```

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+  if (this.task.status === TaskStatus.ERROR) {
+    // ...
+  }
+```
+
+### Task chaining
+
+Sometimes you want to run one task when another task completes. To do this you can use the value of a task as an argument to the other:
+
+```ts
```
e111077, line 211: switchable

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+    args: () => [this._getDataTask.value, this.param],
+  });
+}
+```
+
+This can be useful if the tasks have different arguments so that the chained task may run without the first task running again. In this case it'll use the first task like a cache.
+
+You can also often use one task function and await intermediate results:
+
+```ts
```
e111077, line 229: switchable

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+  });
+
+  private _processDataTask = new Task(this, {
+    task: ([data, param]) => processData(data, param),
+    args: () => [this._getDataTask.value, this.param],
+  });
+}
+```
+
+This can be useful if the tasks have different arguments so that the chained task may run without the first task running again. In this case it'll use the first task like a cache.
```
e111077, line 225: This is an important point. I say move it to the top of the section

e111077 (changes requested):
sorry know it's still in draft but I have some suggestions

packages/lit-dev-content/site/docs/data/task.md
```diff
@@ -0,0 +1,243 @@
...
+
+As a class field, the task status and value are easily available:
+
+```ts
+this._task.value
+```
+
+### Arguments and the task function
+
+The most critical part of a task declaration is the _task function_. This is the function that does the actual work. The Task controller will automatically call the task function with arguments, which have to be supplied with a separate callback. Arguments are separate so they can be checked for changes and the task function only called if the arguments have changed.
```
wimdetroyer, line 93:
```suggestion
The most critical part of a task declaration is the _task function_. This is the function that does the actual work. The Task controller will automatically call the task function with arguments, which have to be supplied with a separate callback. Arguments are separate so they can be checked for changes and the task function is only called if the arguments have changed.
```

packages/lit-dev-content/site/docs/data/task.md
wimdetroyer, line 143: Perhaps specify what happens if the task is configured with _no_ arguments?

packages/lit-dev-content/site/docs/data/task.md
justinfagnani: removed

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+
+```ts
+import {Task} from '@lit-labs/task';
+
+class MyElement extends LitElement {
+  @property() productId?: string;
+
+  private _productTask = new Task(this, {
+    task: async ([productId], {signal}) => {
+      const response = await fetch(`http://example.com/product/${productId}`, {signal});
```
AndrewJakubowicz, line 33: Would it be useful if examples worked in the lit.dev playground without too many changes?

I wanted to try this out, but it required me fixing the URL to `https:` and adding `mode: "no-cors",` to the fetch options. Then the URL always seems to return 404 (which I think makes sense because it's an example URL).

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+
+## Installation
+
+```bash
+npm i @lit/task
+```
+
+## Usage
+
+`Task` is a [reactive controller](../../composition/controllers/), so it can respond to and trigger updates to Lit's reactive update lifecycle.
```
AndrewJakubowicz, line 125: I think we prefer server relative URLs. (That's the convention. I forget the precise reason).
```suggestion
`Task` is a [reactive controller](/docs/v3/composition/controllers/), so it can respond to and trigger updates to Lit's reactive update lifecycle.
```

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+}
+```
+
+{% endswitchable-sample %}
+
+
+As a class field, the task status and value are easily available:
+
+```ts
+this._task.value
```
AndrewJakubowicz, line 149: Nit: in sentence above introduce `status`, and `value`. Could this example also have `this._task.status`?

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+  },
+  args: () => [this.field1, this.field2]
+})
+```
+
+The task function's args array and the args callback should be the same length.
+
+{% aside "positive" "no-header" %}
+
+Write the `task` and `args` functions as arrow function so that the `this` reference points to the host element.
```
AndrewJakubowicz, line 173:
```suggestion
Write the `task` and `args` functions as arrow functions so that the `this` reference points to the host element.
```

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+`Task` is a controller that takes an async task function and runs it either manually or automatically when its arguments change. Task stores the result of the task function and updates the host element when the task function completes so the result can be used in rendering.
+
+### Example
+
+This is an example of using `Task` to call an HTTP API via [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). The API is called whenever the `productId` parameter changes, and the component renders a loading message when the data is being fetched.
+
+{% switchable-sample %}
+
+```ts
+import {Task} from '@lit-labs/task';
```
AndrewJakubowicz, line 26: Nit: `@lit-labs/task` -> `@lit/task`
(and for other code samples)

```suggestion
import {Task} from '@lit/task';
```

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+      // ...
+    },
+  });
+```
+
+{% endswitchable-sample %}
+
+Forwarding the signal to `fetch()` will cause the browser to cancel the network request if the signal is aborted.
+
+You can also check if a signal has been aborted in your task function. It's often useful to race a Promise against a signal:
```
AndrewJakubowicz, line 341: This says "race a signal against a promise", but then the example doesn't look like it's racing. I'd expect some kind of `Promise.race` usage.

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+      signal.throwIfAborted();
+      return secondResult;
+    },
+  });
+```
+
+{% endswitchable-sample %}
+### Task chaining
+
+Sometimes you want to run one task when another task completes. To do this you can use the value of a task as an argument to the other:
```
AndrewJakubowicz, line 372:
```suggestion
Sometimes you want to run one task when another task completes. To do this you can use the value of a task as an argument to another task:
```

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,440 @@
...
+    args: () => [this.dataId, this.param],
+  });
+}
+```
+
+{% endswitchable-sample %}
+
+### Race conditions
+
+A task function can be called while previous task calls are still pending. In this case the `AbortSignal` passed to previous runs will be aborted and the `Task` controller will ignore the result of the previous task runs and only handle the most recent run.
```
AndrewJakubowicz, line 440: Maybe this section can be removed? It seems like a duplicate of the `### Aborting tasks` section earlier in the documentation page.

packages/lit-dev-content/site/docs/v3/libraries/labs.md
```diff
@@ -65,6 +65,20 @@ A package containing controllers and decorators for using the [Context Protocol]
...
+<td>A reactive controller for handling asynchronous tasks.</td>
+<td class="labs-table-links">
+
+[📄&nbsp;Docs](/docs/v3/data/task/ "Docs")<br>[💬&nbsp;Feedback](https://github.com/lit/lit/discussions/3361 "Feedback")<br>[🐞&nbsp;Issues](https://github.com/lit/lit/issues?q=is%3Aissue+is%3Aopen+in%3Atitle+%5Blabs%2Ftask%5D "Issues")
+
+</td>
+</tr>
+
+<tr>
+<td>
```
AndrewJakubowicz, line 80: Awesome!

AndrewJakubowicz (approved):
These docs are awesome! Great stuff!
A couple minor comments, and change the example imports to use the graduated package.

e111077 (approved):
I don't have much time to review this, so don't count my review as blocking!

packages/lit-dev-content/site/docs/v3/libraries/labs.md
```diff
@@ -65,6 +65,20 @@ A package containing controllers and decorators for using the [Context Protocol]
...
+[task](https://www.npmjs.com/package/@lit-labs/task)
+
+</td>
+<td>A reactive controller for handling asynchronous tasks.</td>
+<td class="labs-table-links">
+
+[📄&nbsp;Docs](/docs/v3/data/task/ "Docs")<br>[💬&nbsp;Feedback](https://github.com/lit/lit/discussions/3361 "Feedback")<br>[🐞&nbsp;Issues](https://github.com/lit/lit/issues?q=is%3Aissue+is%3Aopen+in%3Atitle+%5Blabs%2Ftask%5D "Issues")
+
+</td>
+</tr>
```
AndrewJakubowicz, line 77: I think this can be removed? I don't think we are tracking the graduated packages here.

packages/lit-dev-content/site/docs/v3/data/task.md
justinfagnani: Promise.race is a way to race multiple promises against each other, and it's not the only way to race promises. In this case we're not racing multiple promises though, but an abort signal and promises.

justinfagnani: fixed

rictic (approved):

AndrewJakubowicz (approved):

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,444 @@
...
+
+Async data is data that's not available immediately, but may be available at some time in the future. For example, instead of a value like a string or an object that's usable synchronously, a promise provides a value in the future.
+
+Async data is usually returned from an async API, which can come in a few forms:
+- Promises or async functions, like `fetch()`
+- Functions that accept callbacks
+- Objects that emit events, such as DOM events
+- Libraries like observables and signals
+
+The Task controller deals in promises, so no matter the shape of your async API you'll adapt it to promises to use with Task.
```
augustjk, line 109:
```suggestion
The Task controller deals in promises, so no matter the shape of your async API you can adapt it to promises to use with Task.
```

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -0,0 +1,444 @@
...
+
+If arguments are not provided to `run()`, they are gathered from the `args` callback.
+
+### Aborting tasks
+
+A task function can be called while previous task runs are still pending. In these cases the result of the pending task runs will be ignored, and you should try to cancel any outstanding work or network I/O in order to save resources.
+
+You can do with with the [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) that is passed in the `signal` property of the second argument to the task function. When a pending task run is superseded by a new run, the `AbortSignal` that was passed to the pending run is aborted to signal the task run to cancel any pending work.
+
+`AbortSignal` doesn't cancel any work automatically - it is just a signal. To cancel some work you must either do it yourself by checking the signal, or forward the signal to another API that accepts `AbortSignal`s like [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/fetch) or [`addEventListner()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener).
```
augustjk, line 320:
```suggestion
`AbortSignal` doesn't cancel any work automatically - it is just a signal. To cancel some work you must either do it yourself by checking the signal, or forward the signal to another API that accepts `AbortSignal`s like [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/fetch) or [`addEventListener()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener).
```

augustjk (approved):
Some typo

PR #1048, opened 2023-02-22
Update suggested webpack loader for minifying literals

steverep:
This just updates the suggested webpack loader to match the suggested rollup plugin so that they are both using the same minifier under the hood.  This package is also more mature than the current suggestion.


e111077 (approved):

e111077:
thanks for the submission

PR #1050, opened 2023-02-25
fix url in docs/components/shadow-dom.md

38elements:
[#non-rendering](https://lit.dev/docs/templates/expressions/#non-rendering) does not exist. [#removing-child](https://lit.dev/docs/templates/expressions/#removing-child) exists. Therefore, this changed to `#removing-child`.


augustjk (approved):
Thank you for catching this and providing a fix! I updated the link text to match the target location heading and pushed it up. I hope you don't mind.

PR #1051, opened 2023-02-28
Fix a few typos in context docs

RobinTTY:


RobinTTY:
> Thanks a lot @RobinTTY!
> 
> Due to policy, we do need the CLA signed even for very small changes like this. If you can sign it that's great, or I could just fix these if you don't want to.

Already did ;)

augustjk:
Looking good here: https://pr1051-7a1c914---lit-dev-5ftespv5na-uc.a.run.app/docs/data/context/ :)

PR #1052, opened 2023-03-01
Add frameworks section with React

### LINKED ISSUE #884: [labs/react] Create "Frameworks Integration" section for React package
Create a "Frameworks Integration" for the React package under the `Tools` section.

Review and edits required.

Possibly belongs in a separate section than `Tools`

augustjk:
Redo of #1010 for cleaner history and so that preview builds can get generated without approval.

I incorporated all review comments from the previous PR and did a bit more clean up.

Content is based on `@lit-labs/react`'s README. I think there's room for improvement to expand on what the experience with React is like without the wrapper.

For future consideration too is whether the Next.js integration should be part of the "React" page or a separate page under "Framework Integration", leaning towards latter.

Closes #884

### DIFF
--- a/packages/lit-dev-content/site/docs/frameworks/index.md
+++ b/packages/lit-dev-content/site/docs/frameworks/index.md
@@ -1,8 +1,8 @@
 ---
-title: Framework integration
+title: Frameworks
 eleventyNavigation:
-  key: Framework integration
-  order: 12
+  key: Frameworks
+  order: 8
 ---
 
 <!-- This file exists only to create a section heading.
--- a/packages/lit-dev-content/site/docs/frameworks/index.md
+++ b/packages/lit-dev-content/site/docs/frameworks/index.md
@@ -1,0 +1,9 @@
+---
+title: Framework integration
+eleventyNavigation:
+  key: Framework integration
+  order: 12
+---
+
+<!-- This file exists only to create a section heading.
+     Its output is deleted by the Eleventy build process. -->
--- a/packages/lit-dev-content/site/docs/frameworks/react.md
+++ b/packages/lit-dev-content/site/docs/frameworks/react.md
@@ -98,7 +98,7 @@ To render the child to a specific named slot, the standard `slot` attribute can 
 </MyElementComponent>
 ```
 
-Since React components cannot have a `slot` attribute, they will need to be wrapped with another container element. Giving it a `display: contents;` style will make this container disappear.
+Since React components are not themselves HTML elements, they usually cannot directly have a `slot` attribute. To render into a named slot, the component will need to be wrapped with a container element that has a `slot` attribute. If a wrapper element interferes with styling, like for grid and flexbox layouts, giving it a `display: contents;` style ([See MDN for details](https://developer.mozilla.org/en-US/docs/Web/CSS/display#box)) will remove the container from rendering, and only render its children.
 
 ```jsx
 <MyElementComponent>
@@ -112,22 +112,22 @@ Try it out in the [React slots playground example](/playground/#sample=examples/
 
 #### Typing event handler props
 
-The parameter type for the event handler prop can be refined by type casting the event name in the `events` options with the provided `EventName` utility type. Non-casted event names will fall back to the `Event` type.
+In TypeScript, the event type can be specified by casting the event name to the `EventName` utility type. This is a good practice to do so that React users will get the most accurate types for their event callbacks.
+
+The `EventName` type is a string that takes an event interface as a type parameter. Here we cast the `'my-event'` name to an `EventName<MyEvent>` to provide the right event type:
 
 ```ts
-import type {EventName} from '@lit-labs/react';
 
 import React from 'react';
 import {createComponent} from '@lit-labs/react';
-import {MyElement} from './my-element.js';
+import {MyElement, type EventName} from './my-element.js';
 
 export const MyElementComponent = createComponent({
   tagName: 'my-element',
   elementClass: MyElement,
   react: React,
   events: {
-    onClick: 'pointerdown' as EventName<PointerEvent>,
-    onChange: 'input',
+    onMyEvent: 'my-event' as EventName<MyEvent>,
   },
 });
 ```
@@ -136,11 +136,8 @@ With the casting done above, a `PointerEvent` is expected for the parameter of t
 
 ```tsx
 <MyElementComponent
-  onClick={(e: PointerEvent) => {
-    // handle click
-  }}
-  onChange={(e: Event) => {
-    // handle change
+  onMyEvent={(e: MyEvent) => {
+    console.log(e.myEventData);
   }}
 />
 ```
--- a/packages/lit-dev-content/site/docs/frameworks/react.md
+++ b/packages/lit-dev-content/site/docs/frameworks/react.md
@@ -161,7 +161,12 @@ See the [mouse controller example](../../composition/controllers/#example:-mouse
 
 ### How it works
 
-`useController()` uses `useState` to create and store an instance of a controller and a `ReactControllerHost`. It then calls the controller's lifecycle from the hook body and `useLayoutEffect` callbacks, emulating the `ReactiveElement` lifecycle as closely as possible. `ReactControllerHost` implements `addController` so that controller composition works and nested controller lifecycles are called correctly. `ReactControllerHost` also implements `requestUpdate` by calling a `useState` setter, so that a controller with new renderable state can cause its host component to re-render.
+`useController()` creates a custom host object for the controller passed to it and drives the controller's lifecycle by using React hooks.
+
+- `useState()` is used to store an instance of a controller and a `ReactControllerHost`
+- The hook body and `useLayoutEffect()` callbacks emulate the `ReactiveElement` lifecycle as closely as possible.
+- `ReactControllerHost` implements `addController()` so that controller composition works and nested controller lifecycles are called correctly.
+- `ReactControllerHost` also implements `requestUpdate()` by calling a `useState()` setter, so that a controller can cause its host component to re-render.
 
 Controller timings are implemented as follows:
 
--- a/packages/lit-dev-content/site/docs/frameworks/react.md
+++ b/packages/lit-dev-content/site/docs/frameworks/react.md
@@ -65,6 +65,12 @@ After defining the React component, you can use it just as you would any other R
 />
 ```
 
+{% aside "positive" "no-header" %}
+
+See it in action in the [React playground examples](/playground/#sample=examples/react-basics).
+
+{% endaside %}
+
 #### Options
 
 `createComponent` takes an options object with the following properties:
@@ -73,6 +79,36 @@ After defining the React component, you can use it just as you would any other R
 - `elementClass`: The custom element class.
 - `react`: The imported `React` object. This is used to create the wrapper component with the user supplied `React`. This can also be an import of `preact-compat`.
 - `events`: An object that maps an event handler prop to an event name fired by the custom element.
+
+#### Using slots
+
+Children of component created with `createComponent()` will render to the default slot of the custom element.
+
+```jsx
+<MyElementComponent>
+  <p>This will render in the default slot.</p>
+</MyElementComponent>
+```
+
+To render the child to a specific named slot, the standard `slot` attribute can be added.
+
+```jsx
+<MyElementComponent>
+  <p slot="foo">This will render in the slot named "foo".</p>
+</MyElementComponent>
+```
+
+Since React components cannot have a `slot` attribute, they will need to be wrapped with another container element. Giving it a `display: contents;` style will make this container disappear.
+
+```jsx
+<MyElementComponent>
+  <div slot="foo" style="display: contents;">
+    <ReactComponent />
+  </div>
+</MyElementComponent>
+```
+
+Try it out in the [React slots playground example](/playground/#sample=examples/react-slots).
 
 #### Typing event handler props
 
@@ -167,15 +203,3 @@ See the [mouse controller example](../../composition/controllers/#example:-mouse
 - The hook body and `useLayoutEffect()` callbacks emulate the `ReactiveElement` lifecycle as closely as possible.
 - `ReactControllerHost` implements `addController()` so that controller composition works and nested controller lifecycles are called correctly.
 - `ReactControllerHost` also implements `requestUpdate()` by calling a `useState()` setter, so that a controller can cause its host component to re-render.
-
-Controller timings are implemented as follows:
-
-| Controller API   | React hook equivalent               |
-| ---------------- | ----------------------------------- |
-| constructor      | useState initial value              |
-| hostConnected    | useState initial value              |
-| hostDisconnected | useLayoutEffect cleanup, empty deps |
-| hostUpdate       | hook body                           |
-| hostUpdated      | useLayoutEffect                     |
-| requestUpdate    | useState setter                     |
-| updateComplete   | useLayoutEffect                     |
--- a/packages/lit-dev-content/site/docs/frameworks/react.md
+++ b/packages/lit-dev-content/site/docs/frameworks/react.md
@@ -128,9 +128,7 @@ together state and behavior related to a feature. They are similar to React
 hooks in the user cases and capabilities, but are plain JavaScript objects
 instead of functions with hidden state.
 
-`useController()` is a React hook that creates and stores a reactive controller
-and drives its lifecycle using React hooks like `useState` and
-`useLayoutEffect`.
+`useController()` lets you make React hooks out of reactive controllers allowing for the sharing of state and behaviors across web components and React.
 
 ### Usage
 
@@ -139,13 +137,12 @@ import React from 'react';
 import {useController} from '@lit-labs/react/use-controller.js';
 import {MouseController} from '@example/mouse-controller';
 
-// Write a React hook function:
+// Write a custom React hook function:
 const useMouse = () => {
   // Use useController to create and store a controller instance:
   const controller = useController(React, (host) => new MouseController(host));
-  // Return the controller itself
-  // or return a custom object for a more React-idiomatic API:
-  return controller.position;
+  // Return relevant data for consumption by the component:
+  return controller.pos;
 };
 
 // Now use the new hook in a React component:
@@ -159,6 +156,8 @@ const Component = (props) => {
   );
 };
 ```
+
+See the [mouse controller example](../../composition/controllers/#example:-mousemovecontroller) in the reactive controller docs for its implementation.
 
 ### How it works
 
--- a/packages/lit-dev-content/site/docs/frameworks/react.md
+++ b/packages/lit-dev-content/site/docs/frameworks/react.md
@@ -9,9 +9,23 @@ eleventyNavigation:
 
 {% labs-disclaimer %}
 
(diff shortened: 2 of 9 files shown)


packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,153 @@
...
+
+-   `useController()` lets you use a Lit reactive controller as a React hook.  
+
+## createComponent
+
+The `createComponent()` function makes a React component wrapper for a custom element class. The wrapper correctly passes React `props` to properties accepted by the custom element and listens for events dispatched by the custom element.
+
+### How it works
+
+For properties, the wrapper interrogates the web component class to discover its available properties. Then any React props passed with property names are set on the custom element as properties and not attributes.
```
justinfagnani, line 26:
```suggestion
During a render, the wrapper receives props from React and based on the options and the custom element class, changes the behavior of some of the props:

* If a prop name is a property on the custom element, as determined with an `in` check, the wrapper sets that property on the element to the prop value
* If a prop name is an event name passed to the `events` option, the prop value is passed to `addEventListener()` with the name of the event.
* Otherwise the prop is passed to React's `createElement()` to be rendered as an attribute.

Both properties and events are added in `componentDidMount()` and `componentDidUpdate()` callbacks, because the element must have already been instantiated by React in order to access it.
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,153 @@
...
+After defining the React component, you can use it just as you would any other React component.
+
+```jsx
+<MyElementComponent
+  active={isActive}
+  onActivate={(e) => setIsActive(e.active)}
+/>
+```
+
+#### Typescript
```
justinfagnani, line 59: I'm not sure a separate TypeScript section makes as much sense as having details on properties and events. This whole section is basically on events.

A properties section could talk about the difference between properties and attributes when you have both?

I also think it might be worth covering each option: tagName (must be the name the element was registered with), elementClass (hopefully self-explanatory?), react (for compatibility with CJS and so you could pass in Preact).

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,153 @@
...
+Reactive controllers allow developers to hook in to a component's lifecycle to bundle
+together state and behavior related to a feature. They are similar to React
+hooks in the user cases and capabilities, but are plain JavaScript objects
+instead of functions with hidden state.
+
+`useController()` is a React hook that creates and stores a reactive controller
+and drives its lifecycle using React hooks like `useState` and
+`useLayoutEffect`.
+
+### How it works
```
justinfagnani, line 111: Move this down as well

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,153 @@
...
+{% endaside %}
+
+## useController
+
+Reactive controllers allow developers to hook in to a component's lifecycle to bundle
+together state and behavior related to a feature. They are similar to React
+hooks in the user cases and capabilities, but are plain JavaScript objects
+instead of functions with hidden state.
+
+`useController()` is a React hook that creates and stores a reactive controller
```
justinfagnani, line 107: This is a bit more how it works that why you'd use it. I think saying that `useController()` let you make hooks out of controllers so you can share "headless" state and behavior across web components and frameworks is key.

justinfagnani (changes requested):

packages/lit-dev-content/site/docs/frameworks/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Framework integration
```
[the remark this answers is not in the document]
augustjk: I did have that thought. It's currently the longest one there. "Frameworks" seems okay to me. Also +1 for moving polymer article there too.

packages/lit-dev-content/site/docs/frameworks/index.md
```diff
@@ -0,0 +1,9 @@
...
+eleventyNavigation:
+  key: Framework integration
+  order: 12
```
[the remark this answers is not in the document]
augustjk: Renamed and reordered.

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,153 @@
...
+
+-   `createComponent()` creates a React component that _wraps_ an existing web component. The wrapper allows you to set props on the component and add event listeners to the component like you would any other React component.
+
+-   `useController()` lets you use a Lit reactive controller as a React hook.  
+
+## createComponent
+
+The `createComponent()` function makes a React component wrapper for a custom element class. The wrapper correctly passes React `props` to properties accepted by the custom element and listens for events dispatched by the custom element.
+
+### How it works
```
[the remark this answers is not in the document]
augustjk: Done

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: I added a section for the options preceding this. I renamed the "Typescript" heading to "Typing event handler props" as a usage example of the `EventName` utility type we vend. We go into properties/attributes in the "how it works" section.

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Done

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,176 @@
...
+    </pre>
+  );
+};
+```
+
+See the [mouse controller example](../../composition/controllers/#example:-mousemovecontroller) in the reactive controller docs for its implementation.
+
+### How it works
+
+`useController()` uses `useState` to create and store an instance of a controller and a `ReactControllerHost`. It then calls the controller's lifecycle from the hook body and `useLayoutEffect` callbacks, emulating the `ReactiveElement` lifecycle as closely as possible. `ReactControllerHost` implements `addController` so that controller composition works and nested controller lifecycles are called correctly. `ReactControllerHost` also implements `requestUpdate` by calling a `useState` setter, so that a controller with new renderable state can cause its host component to re-render.
```
justinfagnani, line 164: This section needs some help here. It's a little too blunt.

```suggestion
`useController()` creates a custom host object for the controller passed to it and drives the controller's lifecycle by using React hooks.

- `useState()` is used to store an instance of a controller and a `ReactControllerHost`
- The hook body and `useLayoutEffect()` callbacks emulate the `ReactiveElement` lifecycle as closely as possible.
- `ReactControllerHost` implements `addController()` so that controller composition works and nested controller lifecycles are called correctly.
- `ReactControllerHost` also implements `requestUpdate()` by calling a `useState()` setter, so that a controller can cause its host component to re-render.
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,176 @@
...
+
+Controller timings are implemented as follows:
```
justinfagnani, line 166: I think we should drop this section

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,176 @@
...
+#### Options
+
+`createComponent` takes an options object with the following properties:
+
+- `tagName`: The custom element's tag name.
+- `elementClass`: The custom element class.
+- `react`: The imported `React` object. This is used to create the wrapper component with the user supplied `React`. This can also be an import of `preact-compat`.
+- `events`: An object that maps an event handler prop to an event name fired by the custom element.
+
+#### Typing event handler props
```
justinfagnani, line 77: I think we should add a section on slots, just pointing out that React children render to the default slot, and if you want to render to a different slot you need to use the standard `slot` attribute, and that means if you want to slot a React component to a named slot you need to wrap it in an element like `<div slot="foo">` and that you can use `display: contents` to hide that element...

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,176 @@
...
+After defining the React component, you can use it just as you would any other React component.
+
+```jsx
+<MyElementComponent
+  active={isActive}
+  onActivate={(e) => setIsActive(e.active)}
+  onChange={handleChange}
+/>
+```
+
```
justinfagnani, line 67: We have some nice examples in the playground we should link to from here. We should also probably link to the docs from comments (or links in the previews) of some of the playgrounds.

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Done

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Good idea. Added the section above the typing event one as it seems more relevant/important. And linked to the slot example in the playground too.

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Good idea. Done!

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+{% aside "positive" "no-header" %}
```
justinfagnani, line 68: I like this aside!

I wonder, can we have a standard "Try it" phrasing and some kind of different icon for playground links? @aomarks @e111077 if you have opinions there.

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+
+To render the child to a specific named slot, the standard `slot` attribute can be added.
+
+```jsx
+<MyElementComponent>
+  <p slot="foo">This will render in the slot named "foo".</p>
+</MyElementComponent>
+```
+
+Since React components cannot have a `slot` attribute, they will need to be wrapped with another container element. Giving it a `display: contents;` style will make this container disappear.
```
justinfagnani, line 101:
```suggestion
Since React components are not themselves HTML elements, they usually cannot directly have a `slot` attribute. To render into a named slot, the component will need to be wrapped with a container element that has a `slot` attribute. If a wrapper element interferes with styling, like for grid and flexbox layouts, giving it a `display: contents;` style ([See MDN for details](https://developer.mozilla.org/en-US/docs/Web/CSS/display#box)) will remove the container from rendering, and only render its children.
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+
+```jsx
+<MyElementComponent>
+  <div slot="foo" style="display: contents;">
+    <ReactComponent />
+  </div>
+</MyElementComponent>
+```
+
+Try it out in the [React slots playground example](/playground/#sample=examples/react-slots).
```
justinfagnani, line 111: Use the aside here too?

packages/lit-dev-content/site/docs/frameworks/react.md
justinfagnani: I do think that this could still be just "Events" with an initial paragraph describing the event prop to event name mapping in more detail. Maybe saying what naming conventions are (while the name can be whatever you want, they should start with `on`, etc), and that you need to make sure that the event name doesn't collide with other properties on the element.

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+
+#### Typing event handler props
+
+The parameter type for the event handler prop can be refined by type casting the event name in the `events` options with the provided `EventName` utility type. Non-casted event names will fall back to the `Event` type.
```
justinfagnani, line 115:
```suggestion
In TypeScript, the event type can be specified by casting the event name to the `EventName` utility type. This is a good practice to do so that React users will get the most accurate types for their event callbacks.

The `EventName` type is a string that takes an event interface as a type parameter. Here we cast the `'my-event'` name to an `EventName<MyEvent>` to provide the right event type:
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+
+```ts
+import type {EventName} from '@lit-labs/react';
+
+import React from 'react';
+import {createComponent} from '@lit-labs/react';
+import {MyElement} from './my-element.js';
```
justinfagnani, line 122:
```suggestion
import {MyElement, type EventName} from './my-element.js';
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+Try it out in the [React slots playground example](/playground/#sample=examples/react-slots).
+
+#### Typing event handler props
+
+The parameter type for the event handler prop can be refined by type casting the event name in the `events` options with the provided `EventName` utility type. Non-casted event names will fall back to the `Event` type.
+
+```ts
+import type {EventName} from '@lit-labs/react';
```
justinfagnani, line 118:
```suggestion
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+import {createComponent} from '@lit-labs/react';
+import {MyElement} from './my-element.js';
+
+export const MyElementComponent = createComponent({
+  tagName: 'my-element',
+  elementClass: MyElement,
+  react: React,
+  events: {
+    onClick: 'pointerdown' as EventName<PointerEvent>,
+    onChange: 'input',
```
justinfagnani, line 130:
```suggestion
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+import React from 'react';
+import {createComponent} from '@lit-labs/react';
+import {MyElement} from './my-element.js';
+
+export const MyElementComponent = createComponent({
+  tagName: 'my-element',
+  elementClass: MyElement,
+  react: React,
+  events: {
+    onClick: 'pointerdown' as EventName<PointerEvent>,
```
justinfagnani, line 129: We shouldn't use a built-in event here, since the types are already correct.

"MyEvent" might be too generic, but I think it's ok here:

```suggestion
    onMyEvent: 'my-event' as EventName<MyEvent>,
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+With the casting done above, a `PointerEvent` is expected for the parameter of the function provided to the `onClick` prop.
+
+```tsx
+<MyElementComponent
+  onClick={(e: PointerEvent) => {
+    // handle click
+  }}
+  onChange={(e: Event) => {
+    // handle change
+  }}
```
justinfagnani, line 144:
```suggestion
  onMyEvent={(e: MyEvent) => {
    console.log(e.myEventData);
  }}
```

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,205 @@
...
+  elementClass: MyElement,
+  react: React,
+  events: {
+    onClick: 'pointerdown' as EventName<PointerEvent>,
+    onChange: 'input',
+  },
+});
+```
+
+With the casting done above, a `PointerEvent` is expected for the parameter of the function provided to the `onClick` prop.
```
justinfagnani, line 135: Casting the event name t `EventName<MyEvent>` causes the React component to have an `onMyEvent` callback prop that accepts a `MyEvent` parameter instead of a plain `Event`:

justinfagnani (changes requested):
One last round (I think). Thanks for all the work!

packages/lit-dev-content/site/docs/frameworks/react.md
e111077: Why not just do:

`{% playground-ide "examples/react-basics/" %}` and embed it directly

augustjk: I had not considered that possibility! Is it fine to reference the same example in the doc while leaving it standalone in the playgrounds section too? I don't know if we have any precedence for that.

e111077: /shrug it should also stay up to date. Though there's a thing that goes through and essentially server renders the playground previews in the docs section. Not sure if that is limited to the `packages/lit-dev-content/samples/docs` folder

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Regarding prop naming convention, future React support looks like it'll straight up take anything after `on` as the event name. 

Implementation: https://github.com/facebook/react/blob/161f6ae42cf570d3daec355abcf3b006c58315da/packages/react-dom-bindings/src/client/DOMPropertyOperations.js#L163-L171

Tests:
https://github.com/facebook/react/blob/161f6ae42cf570d3daec355abcf3b006c58315da/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js#L173-L213

I'm thinking our example follow this and do `onmy-event` for the prop name that gets called for `my-event`.

justinfagnani: yeah, that's a great idea. Mention that re recommendation is to be compatible with React on this front.

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Done

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: Done

packages/lit-dev-content/site/docs/frameworks/react.md
augustjk: I'll keep them as links to playground in an aside for now. Embedding the existing example is adding too much noise in my eyes when there are already code blocks right above it. I think it would be good future improvement to replace the code blocks with a tailored example.

packages/lit-dev-content/site/docs/frameworks/react.md
```diff
@@ -0,0 +1,208 @@
...
+
+{% aside "positive" "no-header" %}
+
+Try it out in the [React slots playground example](/playground/#sample=examples/react-slots).
+
+{% endaside %}
+
+#### Events
+
+The `events` option takes an object whose key will be the React prop name and value will be the event name. The function passed to the created React component as a prop that matches the key name will be called when the specified event is fired from the custom element. While the the React prop name can be whatever you want, the recommended convention is to add `on` in front of the event name. This is in line with how React is planning to implement event support for custom element properties. You should also make sure this prop name does not collide with any existing properties on the element.
```
justinfagnani, line 119: This paragraph is a bit long and feels like it combines a few topics. Maybe something like this, or tweak as best you can:

```suggestion
The `events` option takes an object maps React prop names to event names. When a component user passes a callback prop with one of the event prop names, the wrapper will add it as an event handler for the corresponding event.

While the the React prop name can be whatever you want, the recommended convention is to add `on` in front of the event name. This matches how React is planning to implement event support for custom elements. You should also make sure this prop name does not collide with any existing properties on the element.
```

justinfagnani (approved):
🎉  awesome work!

Issue #1053, opened 2023-03-05
Incorrect input type on .ts files 


georgeiliadis91:
On the rendering section of the docs. 

https://lit.dev/docs/components/rendering/#composing-templates 

On the typescript example there is a miss match of the types of the files.

<img width="745" alt="Screenshot 2023-03-05 at 8 01 37 PM" src="https://user-images.githubusercontent.com/43796963/222977748-31b4afa2-02c7-407c-b9bb-d994d21bbf3c.png">

e111077:
This is valid TS syntax. Funnily enough, TS doesn't allow you to import .ts files, and unfortunately due to how the playground works, imports without `.js` will not work.

So closing this as WAI

PR #1055, opened 2023-03-05
fix-type-imports-on-docs

georgeiliadis91:
Desciption
==========
On the docs site there is an issue with not missmatching types on the ts examples, where they user .js instead of .ts

Changes
=======
- Removed file types from imports in order not to confuse the user. 

Addresses #1053


e111077:
closing due to explanation on #1053

PR #1059, opened 2023-03-25
fix(list-tutorial): square colors

dhaigh:
This was around the wrong way. White's bottom right square, h1, is always white.

with these changes you get a board looking like this:

for reference:


e111077:
Do you mind just switching `black` and `white` strings rather than inverting the boolean?

e111077 (approved):

e111077:
Thanks for the submission!

PR #1064, opened 2023-04-04
update the banner

e111077:
our banner has old text. let's update it. A-OK to workshop it


augustjk (approved):
Sounds like a good call to action to me.

e111077:
oops updated wrong cl with screenshots
removed

PR #1079, opened 2023-04-27
[infra] Fix dev command

augustjk:
Update `@lit-labs/ssr` to latest to bring in https://github.com/lit/lit/pull/3849 and switch Lit eleventy plugin to VM mode.

`npm run dev` should no longer error.


AndrewJakubowicz (approved):
Thank you! This works great!

Unrelated to this change - I did notice while testing that every single change to a markdown file results in a full ~26 seconds eleventy re-run - but at least it works!

augustjk:
> Thank you! This works great!
> 
> Unrelated to this change - I did notice while testing that every single change to a markdown file results in a full ~26 seconds eleventy re-run - but at least it works!

This is indeed my experience too :(
At least we get to do https://xkcd.com/303/

Issue #1074, opened 2023-04-25
Documentation for Lit 3.0


AndrewJakubowicz:
Tracking issue for Lit.dev Lit 3.0 documentation.

Possible changes:
 - Largest change is browser support page. Fork at the page level using a one off url for the new page.
 - API documentation, think about a tag for API that are in 3.0 but not in 2.0. This could be useful even in patch releases.

PR #1083, opened 2023-04-28
Lit.dev v2 documentation fork (undiscoverable)

AndrewJakubowicz:
Part of https://github.com/lit/lit.dev/issues/1074

## First part of Lit.dev v3

This change has no user facing impact on the site, but instead sets us up for the future Lit.dev 3 user facing changes.
This change archives all the current V2 pages into a `/docs/v2/**` subdirectory, preserving cross linking. I also annotated all the redirects between the page versions so after a `v3` switch is added, the pages nicely navigate between the v1, v2, and v3 versions.

### Testing

This was tested manually. To test user facing behaviors use the second last preview URL before I removed the v3 selector:
 - https://pr1083-4bac296---lit-dev-5ftespv5na-uc.a.run.app/
      - Test by navigating to a page such as styles: https://pr1083-4bac296---lit-dev-5ftespv5na-uc.a.run.app/docs/components/styles/
      - Note that when switching between v1, v2, v3, the pages remain related to styles.

On the most recent preview I tested that the dropdown doesn't show v3, and that everything works like it currently works in production.

Thank you!

Checklist of Lit.dev 3 work

 - [x] V2 documentation forked
 - [x] Redirects between v1, v2 and v3 added.
 - [x] v2 cross links work correctly

These tasks will be moved into a separate PR:
 - [ ] Upgrade section is re-written
 - [ ] Browser Requirements section is re-written
 - [ ] v3 API works
 - [ ] Add back in the v3/v2 user facing UI.


justinfagnani:
@AndrewJakubowicz it'll be a little more work, but I think we should stage this as two separate PRs: one that makes v3 a new "next" version that's not the default, then another PR later that makes v3 the default. This would imply a `/v3/` URL space for now, and maybe a banner there that this is pre-relase.

That way we can publish this upgrade guide along with the prerelease blog post while we're still in prerelease mode.

wdyt?

AndrewJakubowicz:
I've created this PR as an archive of the v2 documentation (undiscoverable) which we can later make discoverable. See this preview link (https://pr1083-4bac296---lit-dev-5ftespv5na-uc.a.run.app/) for the v1/v2/v3 functionality - although note that the default content is still v2 and not updated to v3. This was to test that everything works & make a change that can be landed.

Regarding your pre-release strategy - I love it! Maybe to reduce amount of work, wdyt about doing the following for the "next" version PR:
  - Make a banner announcing the pre-release of Lit 3. Link to a /docs/v3/upgrade guide or blog post that specifically addresses how to try out Lit 3.0.

Then when Lit 3.0 drops, unhide the v2 documentation & land the v3 docs(upgrade & requirements)/api?

augustjk:
I was a bit confused about what hiding v2 meant but that is hiding the v2 fork, and keeping the "v3" (which is really our existing v2 doc) as the default (and v2 dropdown)?

I agree with Justin that while more work, setting up the /v3 namespace right now (and hidden) seems better, and make our current default go to /v2.

augustjk (approved):

justinfagnani:
Can we do this without copying, but _moving_ the current docs to `/docs/v2/` and updating the config to generate `/docs/*` and `/docs/v2/*` output?

I think when we add v3 we might want to have `/docs/v1/`, `/docs/v2/` and `/docs/v3/` folders, and then a config option for what the current docs version is.

AndrewJakubowicz:
Closing this PR. Thank you @justinfagnani for your awesome guidance here. Will pursue the path you suggested in a separate PR.

Issue #1086, opened 2023-04-28
[infra] Update lit.dev Lit v2 api doc generation past the Lerna -> Wireit migration commit


AndrewJakubowicz:
Lit.dev generated API documentation is 10 months old & is pinned to a commit where the Lit project was built with Lerna.

Before attempting to get the Lit v3 branch to work, I need to update the Lit v2 documentation to build with Wireit on a newer commit.

Also blocked by TypeScript version of lit.dev:
 - TS 4.6.2 does not support `module` of `NodeNext` which the Lit repo uses. This breaks generation.

AndrewJakubowicz:
Fixed by https://github.com/lit/lit.dev/pull/1088

PR #1087, opened 2023-04-28
Archive Lit 1 API content so we can upgrade generator for Lit 2 & 3

AndrewJakubowicz:
This PR unblocks updating our Lit v2 API generator (so it's not 10 months out of date) & moves towards unblocking the generation of Lit v3 docs.

Why is documentation generation blocked?
 - To generate documentation we, pull in the lit-html-v1, lit-element-v1 repos and the Lit 2 monorepo.
 - Then we build them (currently by running `lerna` because we pull in an old version of Lit 2).
 - Finally we use typedoc on the built repos to generate the API docs.

Issues:
 - The version of TypeScript matters in order to build what we need in the repos.
 - Updating typedoc currently breaks all generation because it crashes on the old v1 repos.

### How this PR helps?

 - Instead of always rebuilding v1 repos - this PR archives the `pages.json` and `symbols.json` built data for the v1 repos.

This means we no longer build the old documentation and it is essentially frozen. To allow us to continue updating the data of the newer repos I've split out the `api` template to `api-v1` so it can also remain frozen.

### Testing plan

Tested by manually opening up the preview URL and checking that the Lit 1 api docs look good. They should look good because they use a perfect fork of the layout & the data is unchanged... and may never change again :P

Part of https://github.com/lit/lit.dev/issues/1086


AndrewJakubowicz:
Note: This change has caused the API documentation to vanish from the Lit 1 navigation. Otherwise it seems to be working.

packages/lit-dev-content/site/_includes/api-v1.html
```diff
@@ -0,0 +1,319 @@
+---
```
AndrewJakubowicz, line 1: This is a copy of the `api.html` layout (so v1 docs that never change again stay in sync with their layout that never changes). And we can still update newer docs with newer layouts.

packages/lit-dev-tools-cjs/src/api-docs/generate.ts
```diff
@@ -11,14 +11,13 @@ import {execFile} from 'child_process';
...
-import {litElement2Config} from './configs/lit-element-2.js';
-import {litHtml1Config} from './configs/lit-html-1.js';
 
 import type {ApiDocsConfig} from './types.js';
 
 const execFileAsync = promisify(execFile);
 
-const configs = [lit2Config, litElement2Config, litHtml1Config];
+// Only generate documentation for most recent Lit versions.
+const configs = [lit2Config];
```
augustjk, line 20: If we're dropping `litElement2Config` and `litHtml1Config`, should there be some artifact that we check in that gets used for future site builds? Maybe we also want to keep around the config srcs in case we do need to regenerate those at some point?

augustjk: Aha, right! So then my remaining question is whether we might want to keep around the config files without deleting in case we ever need to regenerate them. Though if we start upgrading the generator, maybe the configs might not be compatible?

I guess the point of archiving is that we won't be updating these docs anyway.

augustjk (approved):

AndrewJakubowicz:
I do not think it's worth keeping the config around. They are available via source control if needed. Almost immediately after updating TypeScript and TypeDoc the old configurations no longer work at all. It also appears I may need to change the generator further to work for Lit v2 & v3. Especially because Lit.dev will need to be able to build Lit v3 which uses TypeScript 5.0

packages/lit-dev-tools-cjs/src/api-docs/generate.ts
justinfagnani: I do think it'd be nice to have some way of reconstructing the config that generated those JSON files. Maybe it's as simple as a README in the folder that points to this PR and/or instructions on which commit to check out and which commands to run.

justinfagnani:
Shouldn't this PR include `pages.json` and `symbols.json`?

AndrewJakubowicz:
No - because they are unchanged. They haven't been changed in 2 years: https://github.com/lit/lit.dev/tree/main/packages/lit-dev-api/api-data

packages/lit-dev-tools-cjs/src/api-docs/generate.ts
AndrewJakubowicz: Good idea. I can add a section to the README.

justinfagnani:
@AndrewJakubowicz thanks! I forgot they were already checked-in 😊

justinfagnani (approved):

AndrewJakubowicz:
Added README - PTAL @justinfagnani 
Thanks for the excellent suggestion.

PR #1092, opened 2023-05-02
Add labs/context package to generated API docs

AndrewJakubowicz:
Recommend reviewing 1 PR at a time. Builds on https://github.com/lit/lit.dev/pull/1090

This is a WIP - but I've made the PR early to show that it generates.

PR #1090, opened 2023-05-02
Update TypeDoc

AndrewJakubowicz:
Git is a bit confused about the commits in the change. The only changes not in main are those post the merge commit.

There are certainly improvements that can be made with the newer TypeDoc version. However this PR aims to change the `pages.json` data as little as possible.

To achieve this I've added a glue function which takes the new `TypeDoc.Comment` node and basically replaces it with the legacy fields (`shortText` and `text`) which is what we expect when generating the HTML for the docs.

Because of subtle changes in generation, I've also updated the template to fix the `Parameters` table. For example, compare the preview url with production at:
  - /docs/api/ReactiveElement/#ReactiveElement.requestUpdate

This version of TypeDoc supports TS 5.0.

### Testing

I went page by page in the v2 docs and diffed them. The upgraded TypeDoc pages seem to add some additional "View Source" links.

E.g.,
 - https://pr1090-bfd76a6---lit-dev-5ftespv5na-uc.a.run.app/docs/api/ReactiveElement/#ComplexAttributeConverter.fromAttribute which previously did not have View Source.
 - This PR also fixes some documentation that got lost. E.g. https://pr1090-dd5ea09---lit-dev-5ftespv5na-uc.a.run.app/docs/api/ReactiveElement/#ReactiveElement.updateComplete

Part of https://github.com/lit/lit.dev/issues/1074

### Screenshots

With this PR:
Screenshot 2023-05-02 at 10 24 10 AM

On prod (notice lack of descriptions in the Parameters):
Screenshot 2023-05-02 at 10 24 58 AM

PR #1093, opened 2023-05-03
Update TypeDoc to latest - 0.24.6

AndrewJakubowicz:
This bumps TypeDoc to the latest version. Only one minor patch from the last PR: https://github.com/lit/lit.dev/pull/1090

My hope is that this gives us some more tools for the labs documentation. For example a new `project root path` option is now required & may fix the strange file paths seen in https://github.com/lit/lit.dev/pull/1092.

Regardless - while I have some context in this area - I'm trying to get us as updated as possible.

### Changes

 - TypeDoc removed `kindString`. We use this extensively in our `api.html` template so I needed to add it back using a full tree traversal.
 - TypeDoc now uses `target` to reference the id of the declaration for reflected nodes. This initially broke all linking between symbols. Fixed once I figured out the id was on `target`.
 - Small API compatibility changes.

### Testing

Very manual process again.

This was tested by once again going page by page in the docs and diff checking the content.
I also checked the View Source links continued to work & generated pages look the same between prod and this PR.

### Issues found and fixed list

These issues occurred because initially I did not fully traverse the TypeDoc tree to add `kindString` fields back. Issue resolved by a full traversal.

 - [x] `enableWarning` lost its type: https://pr1093-b007374---lit-dev-5ftespv5na-uc.a.run.app/docs/api/LitElement/#LitElement.enableWarning
 - [x] `creationScope` became a TODO type: https://pr1093-b007374---lit-dev-5ftespv5na-uc.a.run.app/docs/api/LitElement/#RenderOptions.creationScope
 - [x] Lost links to MDN. E.g. https://pr1093-a1f740f---lit-dev-5ftespv5na-uc.a.run.app/docs/api/styles/#CSSResult.styleSheet
 - [x] Lost generated docs cross links. TypeDoc is linking to the exported module & not the symbol declaration!!!!

### DIFF
--- a/packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
+++ b/packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
@@ -106,9 +106,7 @@ export class ApiDocsTransformer {
     symbolMap: SymbolMap;
     pages: Pages;
   }> {
-    this.addKindStringsBackToAllNodes(
-      this.project as unknown as Record<string, unknown>
-    );
+    this.addKindStringsBackToAllNodes(this.project);
     // In the first pass, determine the page/anchor where each node should
     // appear in our layout, and index all nodes by TypeDoc numeric ID.
     for (const entrypoint of this.project.children ?? []) {
@@ -694,7 +692,7 @@ export class ApiDocsTransformer {
    * api.html. This method recursively walks the TypeDoc node and adds
    * `kindString` back to all nodes with a valid `kind` field.
    */
-  private addKindStringsBackToAllNodes(node: Record<string, unknown>) {
+  private addKindStringsBackToAllNodes(node: unknown) {
     if (typeof node !== 'object' || node == null) {
       return;
     }
@@ -707,11 +705,10 @@ export class ApiDocsTransformer {
     for (const [key, val] of Object.entries(node)) {
       if (key === 'kind' && typeof val === 'number') {
         // Add a `kindString` field to the node.
-        node['kindString'] = typedoc.ReflectionKind.singularString(
-          val as ReflectionKind
-        );
+        (node as {kindString: string})['kindString'] =
+          typedoc.ReflectionKind.singularString(val as ReflectionKind);
       }
-      this.addKindStringsBackToAllNodes(val as Record<string, unknown>);
+      this.addKindStringsBackToAllNodes(val);
     }
   }
 
--- a/packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
+++ b/packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
@@ -18,6 +18,7 @@ import {
   ExternalLocation,
   MigrationComment,
 } from './types.js';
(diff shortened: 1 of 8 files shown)


AndrewJakubowicz:
Manually tested and diff checked generated pages on the most recent preview URL and as far as I can find - we have no content changes.

External links (E.g., to MDN from native types work), and internal links, (e.g., `ReactiveElement` or `PropertyDeclaration`) are also working again.

justinfagnani (approved):

packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
```diff
@@ -480,7 +498,14 @@ export class ApiDocsTransformer {
           key === 'tags' ||
           // The "target" key is unstable causing our "Check API data is in
           // sync" approach to fail. We also do not use this key.
-          key === 'target'
+          key === 'target' ||
```
augustjk, line 501: We're now using the key of 'target' for referencing reflected node as per the PR description, and usage in line 425. Do we now want to keep this around?

AndrewJakubowicz:
Thanks for the review - did a rebase with main. Will merge after reviewing new preview url.

packages/lit-dev-tools-cjs/src/api-docs/types.ts
```diff
@@ -28,14 +28,14 @@ export interface MigrationComment
 
 export interface ExtendedDeclarationReflection extends DeclarationReflection {
   location?: Location;
-  externalLocation?: ExternalLocation;
```
augustjk, line 31: Do we need to add `externalUrl` here instead? Maybe not since we're just doing an inline assertion to add anyway, but also maybe yes for documentation purposes.

augustjk (commented):
Ohh boy. Hoping we don't need to do another breaking typedoc update anytime soon.

AndrewJakubowicz:
Haha! This is gonna be the last update for a while of this size. :P Thank you for the review.

packages/lit-dev-tools-cjs/src/api-docs/types.ts
AndrewJakubowicz: `externalUrl` is already documented natively on `DeclarationReflection`. E.g., it's a TypeDoc type that I am "correcting". However I can add it here for additional documentation.

packages/lit-dev-tools-cjs/src/api-docs/transformer.ts
AndrewJakubowicz: Correct that we use `target` and `id` for tracking declarations and references to declarations. This just filters out keys that we don't need for `pages.json` -> the rendered html. So we can remove `target` from the cached output.

packages/lit-dev-tools-cjs/src/api-docs/types.ts
augustjk: Oh if it's already there the no problem! I was just going based off the diff.

augustjk (approved):

Issue #1094, opened 2023-05-04
Remove "Edit this Page" links from generated API.


AndrewJakubowicz:
The "Edit this Page" link will always take the user to the computed `api.html` form - which is a bit confusing.

PR #1095, opened 2023-05-04
Remove editable link from generated API docs.

AndrewJakubowicz:
This link currently does not provide a good user experience (on generated API pages) - linking to a computed page.

For example, clicking: https://lit.dev/docs/api/LitElement/#:~:text=in%20the%20container.-,Edit%20this%20page,-Previous
takes the user to: https://github.com/lit/lit.dev/edit/main/packages/lit-dev-content/site/docs/api/api.html

### Fix

Remove this link from generated API pages.

Fixes: https://github.com/lit/lit.dev/issues/1094

### Testing

Tested by checking that docs pages still have the "Editable" link & generated API docs no longer have that link.

### DIFF
--- a/packages/lit-dev-content/site/_includes/docs.html
+++ b/packages/lit-dev-content/site/_includes/docs.html
@@ -58,7 +58,8 @@
 
       {{ content | safe }}
 
-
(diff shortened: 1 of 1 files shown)


augustjk (approved):
Neat!

PR #1096, opened 2023-05-04
Fix regression in unique symbol type generation

AndrewJakubowicz:
Originally we'd have the text `symbol` for the type of `noChange` or `nothing`. However a change in TypeDoc broke this rendering - and it currently renders TODO.

I fixed the check.

See `Type` in https://lit.dev/docs/api/custom-directives/#noChange or https://lit.dev/docs/api/templates/#nothing

Then compare with preview URL. There are only two instances impacted by this issue (verified by searching `pages.json` data.

### DIFF
--- a/packages/lit-dev-content/site/_includes/api-v1.html
+++ b/packages/lit-dev-content/site/_includes/api-v1.html
@@ -61,7 +61,7 @@ layout: docs
     {%- elif t.operator == 'keyof' -%}
       keyof {{ type(t.target) }}
 
-    {%- elif t.type == "typeOperator" and t.operator == "unique" -%}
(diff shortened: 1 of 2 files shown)


augustjk (commented):
Do we need to change it for v1 docs as well?
https://github.com/lit/lit.dev/blob/a79d914dc25e89acf6fd91842fb5d847995e8e1a/packages/lit-dev-content/site/_includes/api-v1.html#L64

AndrewJakubowicz:
:O Yes we do - I thought we split the docs early enough. Thanks for checking!

augustjk (approved):

AndrewJakubowicz:
Documented v1 issue: https://github.com/lit/lit.dev/issues/1097 Thank you!

PR #1098, opened 2023-05-04
Fix strange TypeDoc regression in Custom Directives API.

AndrewJakubowicz:
Somehow TypeDoc no longer recognizes the type for `ReadonlyArray`. This is a hack just for this page.

See fix on https://pr1098-7a4bf41---lit-dev-5ftespv5na-uc.a.run.app/docs/api/custom-directives/

There are no longer TODO's.
Tracking robust fix in https://github.com/lit/lit.dev/issues/1099

Note: This "fix" is a hack - but it resolves the user facing issue.

### DIFF
--- a/packages/lit-dev-content/site/_includes/api.html
+++ b/packages/lit-dev-content/site/_includes/api.html
@@ -143,6 +143,9 @@ layout: docs
     {%- elif t.name and t.name !== '__type' -%}
       {{- t.name -}}
 
+    {%- elif page.url.includes("api/custom-directives") -%}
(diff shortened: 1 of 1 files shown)


augustjk (approved):
I know we have a separate tracking issue but it would be nice to add a comment describing this 🩹 and maybe link to issue, just in case it takes a while?

AndrewJakubowicz:
Added todo comment with issue link. Great feedback!

augustjk (approved):
Thank you!

Issue #1099, opened 2023-05-04
Fix generated API for the type ReadonlyArray<string>


AndrewJakubowicz:
On https://lit.dev/docs/api/custom-directives/#AttributePart the strings property with type `ReadonlyArray<string>` ends up a `TODO`.

See 
<img width="560" alt="Screenshot 2023-05-04 at 4 47 48 PM" src="https://user-images.githubusercontent.com/15080861/236353171-2aa9d7ba-33df-4435-8985-b9d60fa9c7f2.png">

A hacky fix was introduced in https://github.com/lit/lit.dev/pull/1098 but this needs to be resolved.

PR #1100, opened 2023-05-05
Add lit-labs/context docs

AndrewJakubowicz:
Add (correctly this time), the lit-labs/context documentation.

Note - there is still a strange issue with some of the complex `types`. This issue is what made me notice https://github.com/lit/lit.dev/issues/1099


justinfagnani:
This is looking relatively good. We need a lot more jsdoc in the package, and it'd be really nice to be able to order things so that `createContext`, `consume` and `provide` should up first.

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -36,24 +38,51 @@ export const lit2Config: ApiDocsConfig = {
...
   ],
 
-  symbolOrder: ['LitElement', 'ReactiveElement'],
+  symbolOrder: [
+    'LitElement',
+    'ReactiveElement',
+    /** @lit-labs/context symbol ordering */
+    'createContext',
+    'consume',
+    'provide',
```
justinfagnani, line 84: Let's add ContextConsumer, ContextProvider, and ContextRoot here for now.

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -36,24 +38,51 @@ export const lit2Config: ApiDocsConfig = {
...
+        //   pathlib.join(srcDir, 'polyfill-support.ts'),
+        pathlib.join(srcDir, 'static-html.ts'),
+      ],
+    },
+    // @lit-labs/context documentation
+    {
+      tsConfigPath: pathlib.join(contextDir, 'tsconfig.json'),
+      entrypointModules: [
+        pathlib.join(contextSrcDir, 'index.ts'),
+        pathlib.join(contextSrcDir, 'lib/context-request-event.ts'),
```
justinfagnani, line 68: Do we need these other files as entrypoint modules? Since we want everything importing from index, can typedoc not just find them in the module graph?

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
```diff
@@ -225,6 +264,14 @@ export const lit2Config: ApiDocsConfig = {
     // fine in practice, but when we add e.g. @lit/localize we'll need to be
     // smarter here.
     let [_, pkg, pathMinusExtension] = match;
+
+    if (pkg === 'labs/context') {
+      console.log(pkg, pathMinusExtension);
```
justinfagnani, line 269: remove

packages/lit-dev-tools-cjs/src/api-docs/generate.ts
```diff
@@ -89,37 +89,42 @@ const analyze = async (config: ApiDocsConfig) => {
...
-    entryPointStrategy: typedoc.EntryPointStrategy.Expand,
-  });
-  const root = app.convert();
-  if (!root) {
-    throw new Error('TypeDoc.Application.convert() returned undefined');
-  }
+  const allPages: Pages = [];
+  const allSymbols: SymbolMap = {};
+  for (const pkg of config.packages) {
+    const app = new typedoc.Application();
```
justinfagnani, line 95: At some point I thought I read some docs that typedoc had options for a monorepo, so you didn't need to make a new app for each package. I can't remember it though. This might matter for getting crosslinks between packages. Right now `ReactiveElement` say isn't a link in the context docs.

AndrewJakubowicz: Yeah - I need to investigate this further. I feel like I read that somewhere as well.

packages/lit-dev-tools-cjs/src/api-docs/configs/lit-2.ts
AndrewJakubowicz: Agree - I have to investigate this now that I have stabilized updating all the underlying stuff & fixed many other confounding issues.

Previously when I only used the entrypoint we only got a small cross section of APIs generated. See https://pr1092-ec9e4af---lit-dev-5ftespv5na-uc.a.run.app/docs/api/context/ as an example.

But many things have been fixed since then & I understand what is going on better now. Will try this again to clean it up.
Great comment!

PR #1101, opened 2023-05-05
Add test to catch fallthrough TODO condition in generated docs

AndrewJakubowicz:
This is a test only change - so low risk. (Except I have marked up the error `TODO` so it is more distinct).

### Context

Any slight data change in TypeDoc can result in us failing to render the type of something. The fallthrough case is a `TODO` which ends up rendered in our generated documentation.

We _never_ want this case.

### Fix

An automated test which scans the generated html pages & fails if it locates the fall through html.

I added it to the unit test action because it runs very fast & I didn't think it was worth a separate action. However I am happy to change it to a separate action as well.

### Testing

I tested this by removing the hacky workaround added in https://github.com/lit/lit.dev/pull/1098. This made the test fail. Currently with the workarounds it is green.

Thank you!

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -17,3 +17,4 @@ jobs:
 
       - run: npm ci
       - run: npm run test:unit
+      - run: npm run test:generated-api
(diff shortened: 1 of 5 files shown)


augustjk (approved):
Excellent!

PR #1104, opened 2023-05-06
Adjust Discord Banner Styling

zacknoble:
At the top of the hompage, a blue banner promotes the Lit discord channel.

Looks great on Desktop dimensions.

However, on mobile dimensions where the width of the page is smaller, text spills over to the next line. The container is a forced height, so content is too close to the edges and too far apart between text lines. Also, the lack of padding makes the text bump against the side of the screen at certain dimensions.

I made small changes to all of this. Since this is front and center at the home page of the website, I think it's important there's no underpadded, overflowing content.

### DIFF
--- a/packages/lit-dev-content/site/css/home/1-splash.css
+++ b/packages/lit-dev-content/site/css/home/1-splash.css
@@ -4,12 +4,14 @@
   justify-content: center;
   background: #5865F2;
   color: white;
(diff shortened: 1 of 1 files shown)


justinfagnani:
Looks good locally:


Thanks @zacknoble !

justinfagnani (approved):

PR #1107, opened 2023-05-10
Infrastructure for building unversioned pages from the stable version.

AndrewJakubowicz:
Issue with design and background: https://github.com/lit/lit.dev/issues/1108

### Testing

This was extensively manually tested. E.g., any `/docs/v2` links were checked for the presence of a `rel=canonical` link. `v1` or unversioned pages do not have rel=canonical.

Similarly cross links were checked on `docs/v2/api` and `docs/api`.

Also added some integration tests as smoke testing for the features added.


packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  if (ext === '.md') {
+    const fileName = pathlib.basename(unversionedLocation, '.md');
+    const [frontMatterData, restOfFile] = getFrontMatterData(fileContents);
+    const existingPermalink = frontMatterData.findIndex((val) =>
+      val.includes('permalink:')
+    );
+    if (existingPermalink !== -1) {
+      throw new Error(
+        'Unhandled case: Handle this by transforming the permalink here.'
```
augustjk, line 92: We probably shouldn't get here but would be nice to include the actual filepath if we do.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+      );
+    }
+    if (fileName === 'index') {
+      frontMatterData.push(
+        `permalink: docs/${relativeChildPath.slice(0, -3)}.html`
```
augustjk, line 97: wdyt about making another variable to store the extensionless filename to reuse?
```js
const relativeChildPathWithoutExt = relativeChildPath.slice(0, -ext.length);
```

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+    ) {
+      unversionedLocation = pathlib.join(
+        pathlib.dirname(unversionedLocation),
+        'unversioned.json'
+      );
+      fileContents = JSON.stringify({
+        collection: 'docs-unversioned',
+        latestVersion: SITE_LATEST_VERSION,
+      });
+    }
```
augustjk, line 118: does this need an else block to throw in case of unexpected .json file? alternatively, if it's a given that this'll be the only json file within this directory, maybe the if condition could be dropped?

a comment that this case is to handle `v[X]/v[X].json` file would be nice. personally, it was a bit disorienting to see this between the .md and .html handling when this is more of a one off case.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+  } else if (ext === '.html') {
+    if (pathlib.basename(unversionedLocation, '.html') === 'api') {
+      const [frontMatterData, _] = getFrontMatterData(fileContents);
```
augustjk, line 121: unused trailing value from array destructuring can be dropped, though maybe you left that intentionally there to signal it is intentionally not used, please ignore this comment if so.
```suggestion
      const [frontMatterData] = getFrontMatterData(fileContents);
```

but do .html files never have content following front matter?

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+function transformFile(path: string) {
+  let fileContents = fs.readFileSync(path, {encoding: 'utf8'});
+  const relativeChildPath = pathlib.relative(LATEST_VERSION_CONTENT, path);
+  const ext = pathlib.extname(relativeChildPath);
+  let unversionedLocation = pathlib.join(
+    UNVERSIONED_VERSION_LOCATION,
+    relativeChildPath
+  );
+
+  if (ext === '.md') {
```
augustjk, line 84: nit: i personally feel `switch (ext)` here might be more readable especially as each case also has nested `if`s inside, but this is perfectly fine as is.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+      fileContents = writeFrontMatter(transformedFrontMatter);
+    } else {
+      throw new Error(`Unhandled html document: '${path}'`);
+    }
+  } else {
+    throw new Error(`Unhandled extension '${ext}' for '${path}'`);
+  }
+
+  fs.mkdirSync(pathlib.dirname(unversionedLocation), {recursive: true});
+  fs.writeFileSync(unversionedLocation, fileContents);
```
augustjk, line 134: Would it be too noisy to log something like `Writing ${unversionedLocation}...`?

augustjk (commented):
Great work! I only have very minor comments.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+const SITE_LATEST_VERSION = (
+  JSON.parse(fs.readFileSync(SITE_JSON, 'utf8')) as SiteJSON
+).latestVersion;
+const LATEST_VERSION_CONTENT = pathlib.resolve(
+  CONTENT_PKG,
+  'site',
+  'docs',
+  SITE_LATEST_VERSION
+);
+console.log(LATEST_VERSION_CONTENT);
```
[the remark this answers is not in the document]
AndrewJakubowicz: Yes! Thank you!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,154 @@
...
+    relativeChildPath
+  );
+
+  if (ext === '.md') {
+    const fileName = pathlib.basename(unversionedLocation, '.md');
+    const [frontMatterData, restOfFile] = getFrontMatterData(fileContents);
+    const existingPermalink = frontMatterData.findIndex((val) =>
+      val.includes('permalink:')
+    );
+    if (existingPermalink !== -1) {
```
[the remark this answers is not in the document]
AndrewJakubowicz: Great suggestion, changed `findIndex` to `find` removing the index.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Yes! Great!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Great question! This is scoped to `api.html` which only contains frontmatter.

See: https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/site/docs/api/api.html

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Added! Great suggestion!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Done!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Added some more docs to cover this case

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Check out new code comments in the newest commit & let me know if it sufficiently addresses this. Great suggestion!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Transform the given versioned content file and copy it to the unversioned
+ * directory.
+ */
+function transformFile(path: string) {
+  const relativeChildPath = pathlib.relative(LATEST_VERSION_CONTENT, path);
+  const ext = pathlib.extname(relativeChildPath);
+  const relativeChildPathWithoutExt = relativeChildPath.slice(0, -ext.length);
+  const fileBasename = pathlib.basename(relativeChildPath, ext);
+
+  let fileContents = fs.readFileSync(path, {encoding: 'utf8'});
```
justinfagnani, line 83: Is it possible to make this async? That'll make the Promise.all() in `buildAndTransformUnverionedDocs` more effective.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  console.log(
+    `Writing: ${pathlib.relative(
+      CONTENT_PKG,
+      unversionedLocation
+    )} from ${pathlib.relative(CONTENT_PKG, path)}`
+  );
+  fs.mkdirSync(pathlib.dirname(unversionedLocation), {recursive: true});
+  fs.writeFileSync(unversionedLocation, fileContents);
```
justinfagnani, line 153: async would be good here too

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,173 @@
...
+  const fileBasename = pathlib.basename(relativeChildPath, ext);
+
+  let fileContents = fs.readFileSync(path, {encoding: 'utf8'});
+  let unversionedLocation = pathlib.join(
+    UNVERSIONED_VERSION_LOCATION,
+    relativeChildPath
+  );
+  switch (ext) {
+    case '.md':
+      const [frontMatterData, restOfFile] = getFrontMatterData(fileContents);
```
justinfagnani, line 90: This might be easier to read if these cases were broken out into separate functions, but... this is a utility script. no big deal at all.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,173 @@
...
+  );
+  fs.mkdirSync(pathlib.dirname(unversionedLocation), {recursive: true});
+  fs.writeFileSync(unversionedLocation, fileContents);
+}
+
+/**
+ * Retrieve the 11ty frontmatter from our docs. This is not a full fledged YAML
+ * parser.
+ */
+function getFrontMatterData(
```
justinfagnani, line 160: Why not use a frontmatter parser libraray?

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
```diff
@@ -0,0 +1,173 @@
...
+  fileData: string
+): [EleventyFrontMatterData, string] {
+  const splitFile = fileData.split('---');
```
justinfagnani, line 163: I would be careful and split only on full lines, like `fileData.split(/^---$/);`

justinfagnani (approved):
Awesome work!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: I did not want to add a dependency on a YAML parser for such scoped and light work.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: 100% agree. Making more robust.

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Yes! Done!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Done!

packages/lit-dev-tools-esm/src/build-unversioned-docs.ts
AndrewJakubowicz: Punting for now - but good consideration!

Issue #1112, opened 2023-05-11
Generating Lit 3 API requires a manual change to the `ChildPart` export to work.


AndrewJakubowicz:
### Issue

Currently TypeDoc is unable to find the exported type for `ChildPart`. 

### Repro

Try and run the `lit-3` config without any changes to the built Lit v3 repo. E.g. `npm run build --workspace lit-dev-api`.
Note the following error: `Cannot find ChildPart`.

### Workaround

In `lit-html.d.ts` within `lit.dev/packages/lit-dev-api/api-data/lit-3/repo/packages/lit-html/development/lit-html.d.ts`,

change:

```ts
export type { ChildPart };
declare class ChildPart implements Disconnectable {
```

to

```ts

export declare class ChildPart implements Disconnectable {
```

### Fix

This issue is closed once we can automatically build the Lit 3.0 API docs without any manual intervention. Due to workaround this isn't blocking.

augustjk:
i wonder if somehow typedoc doesn't do the hoisting. perhaps we can move the `export type {ChildPart}; to be below the class declaration?

augustjk:
Fixed by #1119

PR #1113, opened 2023-05-11
Add Lit 3.0 pre-release branch to generated docs

AndrewJakubowicz:
This adds Lit 3.0 api docs - which are generated from https://github.com/lit/lit/pull/3894.

Manually tested by navigating to `/docs/v3/api/custom-directives/`. Note they are not publicly discoverable yet.

They will not show up in our search (which AFAIK only indexes docs at the unversioned `/api` path.

Thank you!

This can be merged prior to fixing https://github.com/lit/lit.dev/issues/1112. - as the manual workaround  produces `pages.json` that is checked in.


packages/lit-dev-content/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "wireit": {
     "build": {
-      "command": "LITDEV_ENV=local NODE_OPTIONS=--experimental-vm-modules eleventy",
+      "command": "LITDEV_ENV=local NODE_OPTIONS=\"--experimental-vm-modules --max_old_space_size=4096\" eleventy",
```
augustjk, line 23: since we're here, maybe we can move the `NODE_OPTIONS` env var setting to wireit's "env"? https://github.com/google/wireit#environment-variables

there might be other things preventing it from working, but this makes these scripts more Windows friendly.

augustjk (approved):
Looking good with a couple minor suggestions.

packages/lit-dev-content/package.json
AndrewJakubowicz: Great suggestion! TIL!

Issue #1111, opened 2023-05-10
Add Lit v3.0 prerelease generated API


AndrewJakubowicz:
### Context

Tracking task to add Lit v3.0 generated documentation to Lit.dev. Does not need to be publicly linked. Can be by URL only.

### Complexity

Naively this is not working because when building with TypeScript, it automatically walks up all parent directories and collects all `@types` encountered. This breaks because `@types/codemirror` on lit.dev is picked up and has incompatible typings with TypeScript 5 and the Lit v3 repo.

References: https://github.com/microsoft/TypeScript/issues/30124 and https://stackoverflow.com/questions/42150920/how-to-block-typescript-2-0-from-walking-up-all-parent-directories-while-resolvi

augustjk:
I found this: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64011

Updating to `@types/codemirror@5.60.7` includes above fix so no longer errors on TS 5.

PR #1119, opened 2023-05-16
Add a file patch to circumvent broken module resolution in v3 generated repo d.ts file

AndrewJakubowicz:
Fixes https://github.com/lit/lit.dev/issues/1111 and https://github.com/lit/lit.dev/issues/1112

### Context

TypeScript 5.0 creates a declaration file that typedoc doesn't resolve. This can be worked around by patching the file after `build:ts` such that it can be understood.

I also bumped the Lit 3.0 pre-release commit hash.

### Test

Tested because `pages.json` hasn't changes and "check API data in sync" has passed.


augustjk (approved):
Confirmed I am able to build the API data on my end. Good workaround!

PR #1120, opened 2023-05-17
Removing event listeners

kevinpschaaf:
Adds a short section on how to remove event listeners from EventPart

### DIFF
--- a/packages/lit-dev-content/site/docs/v2/components/events.md
+++ b/packages/lit-dev-content/site/docs/v2/components/events.md
@@ -168,6 +168,10 @@ When listening to events on repeated items, it's often convenient to use [event 
 
 {% playground-example "docs/components/events/list/" "my-element.ts" %}
 
+### Removing event listeners
+
+Passing `null`, `undefined` or `nothing` to an `@` expression will cause any existing listener to be removed, or simply no-op if no listener has been added.
+
 ## Dispatching events { #dispatching-events }
 
 All DOM nodes can dispatch events using the `dispatchEvent` method. First, create an event instance, specifying the event type and options. Then pass it to `dispatchEvent` as follows:


augustjk (approved):
Great addition!

AndrewJakubowicz (approved):
Awesome!

packages/lit-dev-content/site/docs/v2/components/events.md
```diff
@@ -168,6 +168,10 @@ When listening to events on repeated items, it's often convenient to use [event
 
 {% playground-example "docs/components/events/list/" "my-element.ts" %}
 
+### Removing event listeners
+
+Passing `null`, `undefined` or `nothing` to an `@` expression will cause any existing listener to be removed, or simply no-op if no listener has been added.
```
justinfagnani, line 173: I don't think the no-op part is really necessary

augustjk:
```suggestion
Passing `null`, `undefined` or `nothing` to an `@` expression will cause any existing listener to be removed.
```

packages/lit-dev-content/site/docs/v3/components/events.md
```diff
@@ -168,6 +168,10 @@ When listening to events on repeated items, it's often convenient to use [event
 
 {% playground-example "docs/components/events/list/" "my-element.ts" %}
 
+### Removing event listeners
+
+Passing `null`, `undefined` or `nothing` to an `@` expression will cause any existing listener to be removed, or simply no-op if no listener has been added.
```
augustjk, line 173:
```suggestion
Passing `null`, `undefined` or `nothing` to an `@` expression will cause any existing listener to be removed.
```

augustjk (commented):
I agree with Justin that no-op part seems unnecessary. Added edit suggestions if we want to go with that.

AndrewJakubowicz (approved):

packages/lit-dev-content/site/docs/v2/components/events.md
kevinpschaaf: This PR was specifically addressing a user question that was asked in a forum: "what's a value I can pass to an event part to do nothing (aka be fast)?" Saying only what values remove an event listener still requires the user to infer whether passing undefined is a good default value to avoid extra cost.

Issue #1056, opened 2023-03-08
[docs] Improvements to Frameworks > React page


augustjk:
Page should start off with more general information on using web components in React, and ease into the `@lit-labs/react` package with what it aims to solve.

It would also be good to have embedded playground examples demonstrating this.

PR #1122, opened 2023-05-18
[examples] Update React examples

### LINKED ISSUE #885: [examples/react] Investigate React and Types as modules
Currently, the react examples use the recommendation provided by React.
https://reactjs.org/docs/add-react-to-a-website.html#step-2-add-the-script-tags

After the `window.React` object is exported from a module to create a more cohesive example where demos all use modules.

Typing is lost through this method.

Other options to explore are "modularized" exports from `upkg` or others.

Importing types is difficult for similar reasons.

augustjk:
Fixes #885 

~~Overall strategy is to use playground-element project's [import map support](https://github.com/google/playground-elements#import-maps) to map `react` and `react-dom/client` to a CDN that provides them as ESM, esm.sh.~~

Edit: Using url imports instead of the hidden magic import map.

Types for these were loaded to the example project with a hidden `package.json` file that brings in `@types/react` and `@types/react-dom`. Due to the [lacking support of package exports for playground-elements](https://github.com/google/playground-elements/issues/367), an extra `@types/react-dom/client` is added to handle the `react-dom/client` import.

All React examples were updated to remove the CDN scripts loading the UMD versions of React in `index.html`, and imports were updated to use ~~the bare specifier~~ url import from esm.sh. Also did a prettier pass to the example files for consistency.

Changes to specific examples:
- React Basic
  - ~~Removed wrapper usage as it felt unnecessary for such a simple component.~~
- React events
  - ~~Also removed wrapper as `onClick` doesn't require a wrapper to listen for. React actually handles this fine.~~
- Custom events
  - Change handler functions to start with "handle" rather than "on" to avoid confusion with React's event handler prop name, as [suggested by React docs](https://react.dev/learn/responding-to-events#adding-event-handlers:~:text=By%20convention%2C%20it%20is%20common%20to%20name%20event%20handlers%20as%20handle%20followed%20by%20the%20event%20name.).
  - For the custom element's internal handler function, use private field to avoid overriding
- Slots
  - ~~Removed wrapper as slot behavior is independent of wrapper.~~
  - The only React specific thing here is the fact that React components don't handle the `slot` attribute. Not sure if this example is worth keeping around just for that.
- Refs
  - Did formatting and type fixes.
  - Dropped the `useCallback`s for handlers going on the button. They're going into built-in button elements and will have to re-render every time the component re-renders so there's really no point in memoizing and adds extra noise.
  - I'm also a bit torn about this using the wrapper. The only thing the wrapper does is do one event binding. So much of this is just using a ref to gain access to the underlying element to call methods on it and if we're getting a ref anyway the event could very well be added imperatively too. Left it there for now, but open to suggestions.

Changes to CSP middleware
- Handled TODO regarding main entrypoint CSP rules as https://crbug.com/1253267 was fixed back in 10/2021.
- ~~Allow scripts from `esm.sh` CDN.~~

Lastly, moved a module preload for `tslib.js` to non-dev only as this was 404ing in dev mode.


packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
```diff
@@ -8,7 +8,32 @@ export class DemoGreeting extends LitElement {
...
     `;
   }
 }
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'demo-greeting': DemoGreeting;
+  }
+
+  namespace JSX {
```
AndrewJakubowicz, line 29: Worth adding a comment that this is for good TypeScript types in JSX? (This is my assumption for what this does).

packages/lit-dev-content/samples/examples/react-events/app.tsx
```diff
@@ -1,24 +1,16 @@
...
-const {useCallback, useState} = React;
+import React, {useCallback, useState} from 'react';
+import {createRoot} from 'react-dom/client';
+import './counter-button.js';
 
 export const App = () => {
   const [count, setCount] = useState(0);
 
-  const onClick = useCallback(() => setCount(count + 1), [count]);
+  const handleClick = useCallback(() => setCount((c) => c + 1), []);
```
AndrewJakubowicz, line 8: Spoke offline - here are notes. Although this is a great change, and in line with [React guidance](https://react.dev/reference/react/useCallback#skipping-re-rendering-of-components), in this sample it's not required.

Both `count` and `setCount` are passed to `counter-button`, so memoizing one doesn't provide any advantages - and may provide some overhead.

Per https://react.dev/reference/react/useCallback#skipping-re-rendering-of-components, `useCallback` is not required here.

justinfagnani:
Loading from esm.sh sounds fine, but we might want to comment on the imports to point out that we're doing that. Could we also show the import map file so people can understand what's going on, and maybe educate them about import maps?

justinfagnani:
On the example changes... using the wrapper in the basic cases was there to show how the wrapper works. The Basics example was basically a hello world of wrapper usage. It's nice to show that custom elements just work too, but the point was to show how you can vend an idiomatic-looking React component from a Lit component.

Maybe we can show both uses?

justinfagnani:
How about we combine the React Events and Custom Events demos into one simply called **Events**. I think it should use the wrapper too - the goal is to show how createComponent works.

packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
justinfagnani, line 29: I don't think we should put this with the element file. We don't really expect element authors to add React JSX types too. This is one reason to use the wrapper even for simple elements - you get good types automatically.

If you want to show this I'd put it in app.tsx and try to make it handle all elements with a mapped type. I got something kind of working along these lines for Solid once: https://github.com/shoelace-style/shoelace/discussions/770#discussioncomment-2852125

I'm not 100% sure that didn't mess up types for built-in elements though.

augustjk:
@justinfagnani Thanks for the comments!

Unfortunately there's no way to surface the import map with our current playground set up. It also means we don't have a way to include an import map to playground examples that aren't premade as samples. I think this should be an enhancement for lit.dev/playground

In the meantime, I could add the comment.. or we could just change the import lines like
```ts
import React from 'https://esm.sh/react@18';
```
though we'd need to declare module for that url for TS.

Something like this: https://lit.dev/playground/#sample=examples%2Freact-basics&gist=f0fc35d9ca89a5f53a80599515d8c839
Perhaps this is more transparent than the magic hidden import map?

As for the keeping the wrappers, if the motivation for these playground examples are to showcase the labs/react package, I understand that we'd want to the wrapper for cases where it's not needed. In that case, I'd want the heading for that section of the examples to say `@lit-labs/react`, rather than just `REACT` though. I'm concerned about users thinking the **only** way to use web components in React is with the wrapper.

Showing use case without wrappers can be done instead as an update to the frameworks/react page as part of https://github.com/lit/lit.dev/issues/1056.

packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
augustjk: Thanks for the link! If we end up wanting to keep a bare custom element usage, I'll try this.

packages/lit-dev-content/samples/examples/react-basics/react.d.ts
augustjk, line 1: Only keeping one of these files so the sample TS->JS build is aware of this. Having multiple of these causes TS errors locally though, so for other examples the file is inlined into `project.json`.

augustjk (commented):
Updated PR to use url import instead of the magic hidden import map. Re-added all labs/react wrappers and updated the example section heading to say `@lit-labs/react`.

Also removed all `useCallback`s from the examples as they're unnecessary and add extra noise. (https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere) Thanks @AndrewJakubowicz for the suggestion!

packages/lit-dev-content/samples/examples/react-custom-event/secret-button.ts
```diff
@@ -1,11 +1,32 @@
-import { html, LitElement } from 'lit';
-import { customElement } from 'lit/decorators.js';
+import {html, LitElement} from 'lit';
+import {customElement} from 'lit/decorators.js';
+
+/**
+ * @fires secret-message {CustomEvent<string>} fires on button click and
+ * contains a secret message in detail.
+ */
+@customElement('secret-button')
```
AndrewJakubowicz, line 8: I really like that you moved `secret-button` to the top of the file.

packages/lit-dev-content/samples/examples/react-slots/app.tsx
```diff
@@ -1,35 +1,34 @@
...
 import {SimpleSlots as SimpleSlotsWC} from './simple-slots.js';
+import {createComponent} from '@lit-labs/react';
 
 const SimpleSlots = createComponent({
   react: React,
   tagName: 'simple-slots',
   elementClass: SimpleSlotsWC,
 });
 
-export const App = () => (
```
AndrewJakubowicz, line 11: Great cleanups!

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
```diff
@@ -89,9 +89,6 @@ export const contentSecurityPolicyMiddleware = (
     // es-module-lexer for doing an eval (see next comment for more on that).
     `script-src ${[
       `'self'`,
-      // TODO(aomarks) Remove unsafe-eval when https://crbug.com/1253267 is fixed.
-      // See comment below about playgroundWorkerCsp.
-      `'unsafe-eval'`,
```
AndrewJakubowicz, line 94: No action required - but this change could be broken into an independent PR - mostly if something strange happened it could be rolled back in isolation.

AndrewJakubowicz (approved):
LGTM

Also manually checked JS samples in preview URL

packages/lit-dev-server/src/middleware/content-security-policy-middleware.ts
augustjk: Good idea. Split this out here: https://github.com/lit/lit.dev/pull/1130

augustjk:
Also renamed `Motion` section to `@lit-labs/motion`.

AndrewJakubowicz (approved):
Nice lit-labs motion title update!

justinfagnani:
I'd still like to combine the events examples personally. I don't think the custom event example is that clear, and would be better if the custom element wasn't itself a button which you would expect to fire a click event. The secret message code distracts from the basics of there just being a non-built-in event. We could leave that to another change though.

packages/lit-dev-content/samples/examples/react-basics/app.tsx
```diff
@@ -1,14 +1,14 @@
-import {React, ReactDOM} from './react.js';
+import React from 'https://esm.sh/react@18';
+import {createRoot} from 'https://esm.sh/react-dom@18/client';
 import {createComponent} from '@lit-labs/react';
 import {DemoGreeting as DemoGreetingWC} from './demo-greeting.js';
 
 const DemoGreeting = createComponent({
```
justinfagnani, line 6: Could you add a comment here like `// This creates a React component from a Lit component`?

packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
```diff
@@ -1,5 +1,5 @@
-import { html, LitElement } from 'lit';
-import { customElement, property } from 'lit/decorators.js';
+import {html, LitElement} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
 
 @customElement('demo-greeting')
```
justinfagnani, line 4: Maybe a comment like `// A plain Lit component, which can be used in any framework`?

packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
```diff
@@ -8,7 +8,21 @@ export class DemoGreeting extends LitElement {
...
     return html`
       <p>Hello, ${this.name}!</p>
-      <p>Checkout our <a href="https://lit.dev/docs/frameworks/react/" target="_blank" rel="noopener noreferrer">React framework docs</a>!</p>
+      <p>
+        Checkout our
+        <a
+          href="https://lit.dev/docs/frameworks/react/"
+          target="_blank"
+          rel="noopener noreferrer"
+          >React framework docs</a
```
justinfagnani, line 17: I worry a little bit that any mention of React in this component might make it seem like this component was specifically written to work in React, even though that's not the case. Just thinking from the POV of someone completely unfamiliar with Lit - maybe they're a React dev told to check this out by someone who says you can use Lit in React.

If really we want a message on screen here, we could put it in index.html or app.tsx

packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
```diff
@@ -1,5 +1,5 @@
-import { html, LitElement } from 'lit';
-import { customElement, property } from 'lit/decorators.js';
+import {html, LitElement} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
 
 @customElement('demo-greeting')
 export class DemoGreeting extends LitElement {
```
justinfagnani, line 5: How about some very basic styling to make this look a touch nicer? I'd say:

```css
  display: inline-block;
  border: solid 1px gray;
  background: white;
  padding: 0 1em;
```

AndrewJakubowicz (approved):
Just found the tiniest of spacing nits. I'm stoked by the improvements to the React samples and that you're pushing this initiative forward!

Thank you so much!

packages/lit-dev-content/samples/examples/react-basics/demo-greeting.ts
```diff
@@ -1,23 +1,22 @@
...
-        Checkout our
-        <a
-          href="https://lit.dev/docs/frameworks/react/"
-          target="_blank"
-          rel="noopener noreferrer"
-          >React framework docs</a
-        >!
-      </p>
-    `;
+    return html` <p>Hello, ${this.name}!</p> `;
```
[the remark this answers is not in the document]
augustjk: oops must've been some weird formatting thing. removed!

packages/lit-dev-content/samples/examples/react-events/app.tsx
```diff
@@ -1,18 +1,37 @@
...
+    setMessage('😵');
+  };
+
+  const handleReset = () => {
+    setMessage('😬');
+  };
+
+  return (
+    <main>
+      <h1> Let's play Click Roulette!</h1>
```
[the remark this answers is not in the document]
augustjk: good find, removed!

PR #1126, opened 2023-05-19
Fix comment about ref directive

38elements:
Following code is passing a change callback to ref directive, not passing ref directive to a change callback.

```
return html`<input ${ref(this.inputChanged)}>`;
```


AndrewJakubowicz (approved):
Thank you! Agree that this is a clearer phrasing.

Issue #1124, opened 2023-05-18
Version code samples with doc versions


AndrewJakubowicz:
Related to #1123 

### Context

Currently the samples that are shown within all the docs are unversioned.

Ideally the `docs/v2` samples are using Lit 2, and `docs/v3` are using Lit 3.

PR #1127, opened 2023-05-19
Fork samples so versioned Lit 3 docs playground samples now use Lit 3 pre-release

AndrewJakubowicz:
Fixes: https://github.com/lit/lit.dev/issues/1124

Result of this change is that the docs/v3 subdirectory now uses Lit 3 pre-release for live code samples. All unversioned (and explicitly v2) tutorials, playgrounds, samples now explicitly depend on `v2-base.json`, with behavior unchanged.

### How

Renamed samples/docs to samples/v2-docs & duplicated it into v3-docs. Then forked `base.json` into `v2-base.json` and `v3-base.json`.

Then replaced all `playground-example` and `playground-ide` usages with the versioned counterpart depending on whether the docs are in `docs/v2` or `docs/v3` directories.

### Testing

Some of our integration tests check this. However I also manually went page by page and inspected our playgrounds.
Additional security is that if a file path was broken in the `playground-example` or `playground-ide` shortcode, we get a build error.

While testing this PR I found https://github.com/lit/lit.dev/issues/1128 - however it is non blocking because it's also present in prod.


packages/lit-dev-content/samples/checkable-tutorial-base.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "/samples/base.json",
+  "extends": "/samples/v2-base.json",
```
augustjk, line 2: Would it be easier for tutorials and non-doc samples to keep a `base.json` that we can update to latest version all together? We'd likely want to make those evergreen samples work with the latest version of Lit.

AndrewJakubowicz: That is a great suggestion, and that `base.json` can extend whichever version we want to be the latest. I like it!

AndrewJakubowicz: Great suggestion! Applied!

PR #1129, opened 2023-05-22
Fix Motion examples package.json formatting and trailing comma

AndrewJakubowicz:
### Issue

Motion package.json samples have:

```json
{
  "dependencies": {
    "lit": "^2.0.0",
    "@lit-labs/motion": "^1.0.1",  }
}
```

Which result in error: `Invalid package.json: SyntaxError: Expected double-quoted property name in JSON at position 78`.

Repro at: https://lit.dev/playground/#sample=examples/motion-list

### Fix

Remove trailing comma.

### Testing

This is a tidy up change as there was no user facing impact (unless they created a `package.json`). Tested manually by checking the motion samples still work locally.

### DIFF
--- a/packages/lit-dev-content/samples/examples/motion-carousel/project.json
+++ b/packages/lit-dev-content/samples/examples/motion-carousel/project.json
@@ -8,7 +8,7 @@
     "styles.ts": {},
     "index.html": {},
     "package.json": {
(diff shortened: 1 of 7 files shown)


augustjk (approved):
Thank you!

PR #1130, opened 2023-05-23
Remove CSP rules specific for playground from main

augustjk:
Addressing TODO item as https://crbug.com/1253267 has been fixed for over 1.5 years.


AndrewJakubowicz (approved):
Tested that playground works in Chrome/Firefox/Safari (latest) and it looks great! Thank you!

Issue #1128, opened 2023-05-19
TypeScript compiler error that breaks transitions in tutorials.


AndrewJakubowicz:
Found while manually testing out the whole site https://github.com/lit/lit.dev/pull/1127. Then confirmed the issue is in production. 

Impact of issue is that while navigating the Learn tutorials a certain section may break and stop working until the user refreshes the page.
I have not root caused the issue, and why it appears on only certain tutorials and certain direction navigations.

### Repros

 - Build a tooltip element: Progressing from [Step 8](https://lit.dev/tutorials/tooltip/#8) -> Step 9.
 - Build a word viewer!: Progressing from [Step 2](https://lit.dev/tutorials/word-viewer/#2) -> Step 1.

Every other tutorial does not have this issue. (Manually checked both forward and backwards navigation).

### Investigation

The error message matches https://github.com/microsoft/TypeScript/pull/49738. A possible solution may be updating TypeScript.

~I am not able to reproduce this issue on local dev, only preview environments and production.~
A repro is possible on Firefox.

AndrewJakubowicz:
Issue seems to be the addition of a new import to `lit/directives` in both cases.

AndrewJakubowicz:
Maybe related to https://github.com/microsoft/TypeScript/issues/52636?

AndrewJakubowicz:
I've manually verified that both https://github.com/google/playground-elements/pull/369 and upgrading `playground-elements` to TS version `5.0.4` fixes the issue.

Is there a preference for which should be tackled?

We'll probably want to update the bundled TypeScript within playground-elements to TS 5 eventually.

PR #1131, opened 2023-05-25
Update playground-elements to v0.18.0

AndrewJakubowicz:
Fixes https://github.com/lit/lit.dev/issues/1128

### Testing

Apart from our unit test coverage, I also manually checked the tutorials & code samples on the site.

Because the bundled version has been updated to TS 5, there is a chance that certain gists and prior code now has TS errors. However that does not seem to be avoidable unless we add a feature to bundle TypeScript versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
       ],
       "devDependencies": {
         "@types/codemirror": "^5.60.7",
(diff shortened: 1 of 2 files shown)


augustjk (approved):
Awesome! I can confirm the issue is gone for me in the preview.

PR #1134, opened 2023-05-31
Add a small note about type checking context

rictic:
I didn't want to go into the full details because it felt like it would disrupt the flow of this page, and the main information that people need to know is that we're not type checking private or protected fields.


rictic:
Yeah, the comment is there for maintainers of the site, not for visitors. Clarified into a TODO

Thanks for the review!

PR #1135, opened 2023-06-01
Fix font in examples/properties-basic sample

AndrewJakubowicz:
Fixes font not being applied in https://lit.dev/playground/#sample=examples/properties-basic, due to an errant single quote at the end of the `font-family` value.

Fix is to replace quote with semicolon. It's also technically possible to leave out the last semicolon so removing the single quote is also a possible solution.

Tested manually by checking the sample looks good now!

#### Prior
Screenshot 2023-05-31 at 5 41 53 PM

In developer tools: `Rendered font: Times—Local file(33 glyphs)`

#### With this change

Screenshot 2023-05-31 at 5 41 46 PM

### DIFF
--- a/packages/lit-dev-content/samples/examples/properties-basic/index.html
+++ b/packages/lit-dev-content/samples/examples/properties-basic/index.html
@@ -4,7 +4,7 @@
     <script type="module" src="./name-tag.js"></script>
     <style>
       body {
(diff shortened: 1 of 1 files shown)


augustjk (approved):
Good find!

PR #1136, opened 2023-06-01
[infra] Update rollup config and remove unneeded types packages

augustjk:
Removed `@types/source-map` and `@types/node-fetch` as these packages already ships with type. The following warnings were showing up on npm installs:
```
npm WARN deprecated @types/source-map@0.5.7: This is a stub types definition for source-map (https://github.com/mozilla/source-map). source-map provides its own type definitions, so you don't need @types/source-map installed!
npm WARN deprecated @types/node-fetch@3.0.3: This is a stub types definition. node-fetch provides its own type definitions, so you do not need this installed.
```

Removed the `name` fields from rollup output config that was causing the following warning:
```
(!) You have passed an unrecognized option
Unknown input options: name. Allowed options: acorn, acornInjectPlugins, cache, context, experimentalCacheExpiry, external, inlineDynamicImports, input, makeAbsoluteExternalsRelative, manualChunks, maxParallelFileOps, maxParallelFileReads, moduleContext, onwarn, perf, plugins, preserveEntrySignatures, preserveModules, preserveSymlinks, shimMissingExports, strictDeprecations, treeshake, watch
```

Added `preserveEntrySignatures: false` to rollup config as per this warning:
```
(!) To preserve the export signature of the entry module "lib/global/hydrate-common-components.js", an empty facade chunk was created. This often happens when creating a bundle for a web app where chunks are placed in script tags and exports are ignored. In this case it is recommended to set "preserveEntrySignatures: false" to avoid this and reduce the number of chunks. Otherwise if this is intentional, set "preserveEntrySignatures: 'strict'" explicitly to silence this warning.
https://rollupjs.org/guide/en/#preserveentrysignatures
```

### DIFF
--- a/packages/lit-dev-content/rollup.config.js
+++ b/packages/lit-dev-content/rollup.config.js
@@ -99,7 +99,7 @@ export default [
         showMinifiedSize: false,
       }),
     ],
-    preserveEntrySignatures: 'strict',
+    preserveEntrySignatures: false,
   },
 
   // A separate bundle is made for the server so that we do not modify the
--- a/packages/lit-dev-content/rollup.config.js
+++ b/packages/lit-dev-content/rollup.config.js
@@ -30,7 +30,6 @@ const terserOptions = {
 
 export default [
   {
-    name: 'client',
     input: [
       // lit-hydrate-support MUST be loaded first to make sure lit hydration
       // helpers are bundled before LitElement attempts to use hydration support
@@ -100,14 +99,13 @@ export default [
         showMinifiedSize: false,
       }),
     ],
+    preserveEntrySignatures: 'strict',
   },
+
+  // A separate bundle is made for the server so that we do not modify the
+  // client module graph just to SSR a component.
   {
-    name: 'SSR',
-    input: [
-      // A separate bundle is made for the server so that we do not modify the
-      // client module graph just to SSR a component.
-      'lib/components/ssr.js',
-    ],
+    input: ['lib/components/ssr.js'],
     output: {
       dir: 'rollupout/server',
       format: 'esm',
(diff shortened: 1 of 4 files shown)


packages/lit-dev-content/rollup.config.js
```diff
@@ -100,14 +99,13 @@ export default [
         showMinifiedSize: false,
       }),
     ],
+    preserveEntrySignatures: 'strict',
```
AndrewJakubowicz, line 102: In the PR comment you wrote: `preserveEntrySignatures: false` while here it is set to `strict`. I assume we need it explicitly set to generate the chunk?

augustjk: ah, good catch. I think it should be `false` as the docs and the warning message states `false` when building for a web app vs `strict` when building for a library.

augustjk: so setting it to `'strict'` which was the behavior when it wasn't set results in `js/global/hydrate-common-components.js` have `import '../hydrate-common-components.js` causing another request to happen. 
image

with `false`, it doesn't create that extra step and the needed code is in `js/global/hydrate-common-components.js`.
image

AndrewJakubowicz: Awesome! Confirmed in preview URL that we now have one less request! Great removal of a unnecessary network round trip.

Comparing 
https://pr1136-037cc06---lit-dev-5ftespv5na-uc.a.run.app/js/global/hydrate-common-components.js to 
https://lit.dev/js/global/hydrate-common-components.js

So much better!

AndrewJakubowicz (approved):
Brilliant

PR #1137, opened 2023-06-01
[examples] Pin `@types/react`

augustjk:
It seems the latest update to `@types/react` https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65135 is causing a TS error in the playground.
image

I can't reproduce the TS error in local VS Code and it only seems to happen within playground-elements. Will need to investigate this but for now pinning the version so the error isn't present in the playground seems good.

### DIFF
--- a/packages/lit-dev-content/samples/examples/react-basics/project.json
+++ b/packages/lit-dev-content/samples/examples/react-basics/project.json
@@ -15,7 +15,7 @@
     },
     "index.html": {},
     "package.json": {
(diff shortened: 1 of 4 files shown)


AndrewJakubowicz (approved):
Nice - looks great on preview URL!

nickmccurdy:
Is the playground using TypeScript 5.1?

augustjk:
> Is the playground using TypeScript 5.1?

playground is currently using TypesScript 5.0.4

nickmccurdy:
I'd recommend upgrading TypeScript, reverting this, and seeing if it solves the error.

PR #1138, opened 2023-06-06
[infra] Deflake integration test

augustjk:
This should fix test failures like the one here: https://github.com/lit/lit.dev/actions/runs/5190347793/jobs/9356646136

This was due to the changing of code before the playground preview was loaded, which somehow broke the style such that the text just became black.


AndrewJakubowicz (approved):
Brilliant!  - Thank you for figuring this out!

PR #1143, opened 2023-06-26
[infra] Use latest kaniko executor image for cloudbuild

augustjk:
The issue mentioned in the comment with the version pinning https://github.com/GoogleContainerTools/kaniko/issues/1786 has been fixed.

Our production build started failing though and seems kaniko related. The error shown in GCP is the same as https://github.com/GoogleContainerTools/kaniko/issues/1604

Not sure why we just started seeing this now, but trying to see if using the latest version fixes this.

### DIFF
--- a/cloudbuild-main.yaml
+++ b/cloudbuild-main.yaml
@@ -9,9 +9,7 @@ steps:
   # https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/docker
   # https://cloud.google.com/build/docs/kaniko-cache
   - id: build
-    # Kaniko pinned to earlier version due to
-    # https://github.com/GoogleContainerTools/kaniko/issues/1786
-    name: gcr.io/kaniko-project/executor:v1.6.0
+    name: gcr.io/kaniko-project/executor:latest
     args:
       - --dockerfile=./Dockerfile
       - --destination=$_GCR_HOSTNAME/$PROJECT_ID/$REPO_NAME/lit-dev:$COMMIT_SHA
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz:
If the preview deploy works does this verify that the change worked? Thank you!

augustjk:
> If the preview deploy works does this verify that the change worked? Thank you!

Yes the failing build step is practically identical for prod build and preview build so as it looks like it made it past the build step, I have confidence this should work for prod build as well. No risk if it somehow fails after merging as that's no worse than where we're currently at and we'll just be serving the site from an older version.

cloudbuild-main.yaml
```diff
@@ -9,9 +9,7 @@ steps:
   # https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/docker
   # https://cloud.google.com/build/docs/kaniko-cache
   - id: build
-    # Kaniko pinned to earlier version due to
```
AndrewJakubowicz, line 12: Is it worth pinning to the latest specific version so that it doesn't update out from under us?

AndrewJakubowicz (approved):
Great! Only thought is if we should still be pinning the version? Thank you!

cloudbuild-main.yaml
augustjk: I had considered that, but figured it'd be fine to get the benefit of auto update versus the risk of some version breaking us.

augustjk:
Prod build and deploy was successful with this and I can see the typo fix of the previously merged PR (with the failing build) applied in the live site.

PR #1157, opened 2023-08-03
fix typo in templates/custom-directives.md

38elements:
I think `the subtree` is unnecessary, since it does not fit context.


AndrewJakubowicz (commented):
Agree that referring to the DOM subtree that was disconnected and then reconnected is not necessary as it's described in detail in the prose above.

However, now the `is in` doesn't make sense because it was referring to the DOM subtree. Please remove the `is in` as well. Thank you!

38elements:
Thank you for review. I read your comment. I worried about why describing the same thing twice in this section. I thought the twice describing was because the author of the original sentence wanted to emphasize the cases where `reconnected` occurs in this sentence. Becouse most people may be hard to come up with a case where reconnection occurs. I am sorry.  Therefore, I think it is better not to chage.

Issue #822, opened 2022-05-10
Feature tutorials catalog and YouTube content


sorvell:
There is currently nothing on the site drawing attention to the tutorials catalog or the YouTube channel. 

Ideally, we should blog about both and probably do some re-organization to be able to more prominently feature the YT content.  

In the short term, offline consensus was to add a banner to the home page.

Some ideas for longer term:
* Have a "What's new" card on the home page
* Rename the Blog to Articles
* Change `Tutorials` to `Learn` and include prominent links to YouTube.

sorvell:
Related to https://github.com/lit/lit.dev/issues/678.

PR #1160, opened 2023-08-09
[feature] add Learn catalog page at lit.dev/learn

AndrewJakubowicz:
Part of https://github.com/lit/lit.dev/issues/822

### Why

We currently have a tutorials page. This is an extension of that page, but also including video and article content.

### How

This PR adds no new content, and instead creates a new catalog page, under `/learn`. This page collects:
 - videos
 - tutorials
 - articles

The content is shown in date order, so newest content is up top, and you can filter between the three content types. For this to work, all content now has a `date` field.
The card summaries are also manually taken from a summary field.

I don't think there is enough content to warrant the complexity of adding a search, or featured section yet.

### Testing

I tested with voice over on mac, as well as manually via preview link.

This page won't actually be made public after this PR. Instead it will only be accessible via `/learn` URL. Making this page public is a separate step.


justinfagnani:
I think for the cards we should follow what I think it a pretty standard layout these days:

1. Full-bleed image at the op
2. Title
3. Summary / snippet
4. Metadata

AndrewJakubowicz:
@justinfagnani Agree. The images currently look janky. Do you have images in mind for the non-video content?

AndrewJakubowicz:
What about something like this?

localhost_8080_learn_ (1)

AndrewJakubowicz:
Notes from conversation with @e111077 offline.

 - Consider using the small tiled Lit logos as the background instead of a plain blue.
 - Order the content reverse chronologically, with newest content up the top.
 - Add a featured section at the top for content we want to highlight.
 - Make sure it's accessible.
 - It's ok while we don't have too much content to put everything on one page.
 - Simple CSS based "Type filter chips" is a good simple solution.
 - Refine the summary text truncation algorithm.

AndrewJakubowicz:
Filter chips now work.

e111077:
sections are missing hover styles. I recommend following [general card guidelines](https://m3.material.io/components/cards/guidelines) to guide interaction.

AndrewJakubowicz:
> sections on learn page are missing hover styles. I recommend following [general card guidelines](https://m3.material.io/components/cards/guidelines) to guide interaction.

I've implemented your great suggestion you gave me offline (add underline on hover). There are now hover styles :D

justinfagnani:
The font sizes are a bit different than anywhere else on the site. For main content we use Manrope 22px title / Open Sans 16px body. For the blog list we use Manrope 24px title / Manrope 18px body. And for this we have Manrope 26px title / Manrope 20px body.

The large font sizes make things look a bit big to me. I had to check my Chrome zoom settings to see if I was zoomed in. I'd at least drop to what's on the blog list.

The "Lit 3.0 Launch Event" cover image is a bit blurry too. Is there a card size that lines up with YT preview sizes better?

packages/lit-dev-content/site/learn/learn.11tydata.js
```diff
@@ -0,0 +1,190 @@
...
+      date: "May 5 2022",
+    },
+    {
+      title: 'How to style your Lit elements',
+      summary: `We cover how the Shadow DOM works, illustrate the benefits of encapsulated CSS, and show you how to use CSS inheritance, custom properties and shadow parts to expose a flexible public styling API.`,
+      youtubeId: 'Xt7blcyuw5s',
+      date: "Oct 3 2022",
+    },
+    {
+      title: 'Introduction to Lit - Lit University (Basics)',
```
justinfagnani, line 115: I think the hyphen and the parentheses make this title a bit noisy.

The _Build With Lit_ series doesn't have the series names in the title, do the _Lit University_ videos need that?

packages/lit-dev-content/site/learn/index.html
```diff
@@ -0,0 +1,57 @@
...
+      <md-filter-chip data-value="article" label="Articles" selected></md-filter-chip>
+      <md-filter-chip data-value="video" label="Videos" selected></md-filter-chip>
+      <md-filter-chip data-value="tutorial" label="Tutorials" selected></md-filter-chip>
+    </md-chip-set>
+    <div class="card-grid">
+      {% for content in learn %}
+      <a class="card" data-content-kind="{{ content.kind }}" href="{{ content.url }}">
+        <div class="card-header">
+          {% if content.kind == "video" %}
+          <img
```
justinfagnani, line 30: I think we should have an `` tag for every card, but fall back to a default if something doesn't have one.

Right away I want to make a custom card for the Polymer article, and probably a default tutorial card.

justinfagnani: for making custom cards, can you put in a readme somewhere what the image size should be?

packages/lit-dev-content/site/learn/learn.11tydata.js
AndrewJakubowicz: Agree about noise. I've removed the series from the title. (It's visible in the thumbnail)

packages/lit-dev-content/site/learn/index.html
AndrewJakubowicz: I added it to the learn 11ty data file, near the thumbnail error message.

justinfagnani (approved):
awesome!

augustjk (approved):
Love it!

PR #1162, opened 2023-08-10
Update playground-elements to v0.18.1

AndrewJakubowicz:
Tab should now be two spaces instead of one.

Upgrade to playground-elements 0.18.1: https://github.com/google/playground-elements/releases/tag/v0.18.1

### Test plan

Tested by spot checking the playground & demos. Pressed tab and checked that 2 spaces were created.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "packages/*"
       ],
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


augustjk (approved):
can confirm tab button produces 2 spaces in preview playground!

Issue #1170, opened 2023-08-15
Update Releases section to contain notes for each release.


justinfagnani:
You should be able to see 2.0 and 3.0 release notes and upgrade guides no matter which version of the docs you have selected.

AndrewJakubowicz:
This was tested in https://github.com/lit/lit.dev/pull/1178, but I'm not sure if we want to have both.

Maybe only both Releases in the v3 version of the docs?

augustjk:
Low effort, could add link to Lit 2 upgrade guide at top of Lit 3 upgrade guide.

We should also add to the document title "Lit 3 upgrade guide" rather than "Upgrade guide".

I think it might be easy enough to do:

lit.dev v3
Releases
 - Lit 3 upgrade guide
 - Lit 2 upgrade guide (either copy or link to /v2/upgrade-guide)

lit.dev v2
Releases
 - Lit 2 upgrade guide

AndrewJakubowicz:
This issue ended up being resolved via @augustjk's suggestion of adding a link to the top.

<img width="841" alt="Screenshot 2023-10-09 at 11 10 23 AM" src="https://github.com/lit/lit.dev/assets/15080861/978ab7b7-629b-41cb-8fe6-71049f72e754">

PR #1178, opened 2023-08-25
[docs] Lit 2 and Lit 3 upgrade guides should be visible in both versions

AndrewJakubowicz:
Fixes issue: https://github.com/lit/lit.dev/issues/1170

### Context

We want both upgrade guides to be visible regardless of the dropdown selector.

Prior paths:
 - v2 selected: `/releases/upgrade/` is the Lit 2 upgrade guide.
 - v3 selected: `/releases/upgrade/` is the Lit 3 upgrade guide.

These paths are unchanged, but the following have been added.

- v2 selected: `/releases/upgrade-to-v3/` is the Lit 3 upgrade guide. It contains a pre-release warning.
- v3 selected: `/releases/upgrade-to-v2/` is the Lit 2 upgrade guide.

`versionLinks` frontmatter has been updated so that if you switch the version selector then you stay on the correct upgrade guide.

The titles have also been updated from `Upgrade Guide` to `Lit N Upgrade Guide`. This doesn't change the url, and is visual only.

### Testing

Tested manually and via:

- https://pr1178-404c018---lit-dev-5ftespv5na-uc.a.run.app/docs/v2/releases/upgrade/
- https://pr1178-404c018---lit-dev-5ftespv5na-uc.a.run.app/docs/v2/releases/upgrade-to-v3/
- https://pr1178-404c018---lit-dev-5ftespv5na-uc.a.run.app/docs/v3/releases/upgrade-to-v2/
- https://pr1178-404c018---lit-dev-5ftespv5na-uc.a.run.app/docs/v3/releases/upgrade/

### DIFF
--- a/packages/lit-dev-content/site/docs/v2/releases/upgrade-to-v3.md
+++ b/packages/lit-dev-content/site/docs/v2/releases/upgrade-to-v3.md
@@ -5,8 +5,15 @@ eleventyNavigation:
   parent: Releases
   order: 3
 versionLinks:
-  v2: releases/upgrade-to-v3/
+  v3: releases/upgrade/
 ---
+
+<div class="alert alert-warning">
+
+Lit 3 is currently in pre-release. See [Announcing Lit 3.0
+Pre-release](/blog/2023-05-15-lit-3.0-prerelease/) blog post.
+
+</div>
 
 ## Overview
 
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade-to-v2.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade-to-v2.md
@@ -5,7 +5,7 @@ eleventyNavigation:
   parent: Releases
   order: 3
 versionLinks:
-  v3: releases/upgrade-to-v2/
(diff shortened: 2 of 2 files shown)


augustjk (commented):
I had a brief thought regarding whether we might have the links actually switch the selected version and navigate to the upgrade guide for that version, but maybe that's more disorienting.

Maybe this is a discussion that should be had in the issue, but I don't know if I agree that all version should show all upgrade guides.

I'm on board with "default" version doc showing all upgrade guides, reverse chronological order. I don't particularly like that v2 docs have both v2 and v3 upgrade docs. This is kind of weird right now because v2 is the default.

I think once v3 is default, we can have v3 show v3 and v2 upgrade guide. v2 doc will only have v2 upgrade guide, but will have a banner up top stating this is an older version of the doc so doesn't include everything.

AndrewJakubowicz:
Agree. After looking at preview I also think that `v2` probably should not have `v3` in it yet. Downgrading to draft status right now.

We can decide in the future whether to implement this change, part of this change, or close the change.

PR #1182, opened 2023-09-11
Update MDN mixins link

aguingand:
MDN reference about mix-ins is now under an "extends" subsection. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/extends#mix-ins)


AndrewJakubowicz (approved):
Awesome thank you!

PR #1198, opened 2023-10-02
[infra] attempt to fix timeout on blocking render

AndrewJakubowicz:
Currently GitHub actions timeout and thus never run integration tests.

This is an attempt at a quick fix, to unblock us cheaply so working on Lit 3.0 release can go smoothly.


AndrewJakubowicz:
Failed. But ruled out some options.

Issue #1189, opened 2023-09-20
Update 3.0 upgrade guide


justinfagnani:
We have a few more changes we need to add:

- [x] Mention new language level and potential toolchain requirements
- [ ] renderRoot type change
- [x] Mention decorators

AndrewJakubowicz:
Decorators handled in: https://github.com/lit/lit.dev/pull/1199

AndrewJakubowicz:
RenderRoot has not been added to 3.0 thus far: https://github.com/lit/lit/pull/4254
Probably will be added as a feature.

And the new language level and tooling upgrade guide was added in: https://github.com/lit/lit.dev/pull/1210


Hence, this issue is fixed

PR #1199, opened 2023-10-02
v3: upgrade guide decorator section

AndrewJakubowicz:
Issue: https://github.com/lit/lit.dev/issues/1189

This PR specifically addresses the decorator section.

### Context

Since this guide was last written, we now have hybrid decorators that are compatible with both experimental and standard decorator environments.

This change updates the guidance to include the breaking changes in experimental decorators.

Other changes:
 - Add a link to the v2 upgrade guide on the top.
 - Rename the upgrade guide to `Upgrade to Lit 3.x guide`.
 - Update the custom accessor guidance. You no longer need to call `requestUpdate`.

### Release plan

Note this can be merged directly to main, as it's scoped to the `v3` documentation.

### Tested

Manually via preview link.

Thank you!

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -147,9 +147,38 @@ set myProperty (val) { ... }
 
 ### Optional: upgrade to standard decorators {#standard-decorator-migration}
 
-To use standard decorators, your decorators should add the `accessor` keyword.
-Lit 3 decorators are flexible and work as both experimental decorators, and as
-standard decorators.
+
+While Lit 3 adds support for standard decorators, we still recommend that TypeScript users stay with experimental decorators. This is because the emitted code for standard decorators from the TypeScript and Babel compilers is quite large at the moment.
+
+We will recommend standard decorators for production when browsers support them, or when we release decorator transform support in our new Lit compiler.
+
+But you can try standard decorators now, and they work in TypeScript 5.2 and above and Babel 7.23 with the `@babel/plugin-proposal-decorators` plugin.
+
+### Configuration
+
+#### TypeScript
+
+Install TypeScript 5.2 or later, and _remove_ the `"experimentalDecorators"` setting from your tsconfig if it's present.
+
+#### Babel
+
+Install Babel 7.23 or later, and [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/babel-plugin-proposal-decorators). Be sure to pass the `"version": "2023-05"` option to the plugin.
+
+### Source updates
+
+#### Add the `accessor` keyword to decorated fields.
+
+Standard decorators are not allowed to change the _kind_ of class member they decorate. Decorators that need to create getters and setters must be applied to existing getters and setters. To make this more ergonomic the decorators standard adds the `accessor` keyword which creates "auto accessors" when applied to a class field. Auto accessors look and act much like class fields, but create accessors on the prototype backed by private storage.
+
+```ts
+class A {
+  accessor foo = 42;
+}
+```ts
+
+The `@property()`, `@state()`, `@query()`, `@queryAll()`, `@queryAssignedElements()` and `@queryAssignedNode()` decorators require the `accessor` keyword.
+
+Example:
 
 ```ts
 class MyElement extends LitElement {
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -31,23 +31,28 @@ ranges to include both 2.x and 3.x, like `"^2.7.0 || ^3.0.0"`.
 
 Lit 2.x and 3.0 are _interoperable_: templates, base classes, and directives from one version of Lit will work with those from another.
 
-## Decorator behavior has been unified
+## Updates to Lit decorators
 
-In Lit 3.0 our decorators support running in three configurations:
+The Lit 3 decorators are largely backwards compatible. If you've been using them with TypeScript, **most likely no changes are needed**. We've added forward compatibility with standard decorators, but the legacy `experimentalDecorators: true` without the `accessor` keyword is currently the more performant option.
 
-1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2 users. As standard decorators are still immature, this is the most performant option that we recommend for production for the time being.
-2. With the TypeScript `experimentalDecorators` flag and with the `accessor` keyword. This aids migration to standard decorators. Code can add the `accessor` keyword one decorator at a time to migrate gradually.
+In Lit 3.0 there are three supported ways to use our decorators:
+
+1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2.
+2. With the TypeScript `experimentalDecorators` flag and using the `accessor` keyword. This aids migration to standard decorators. [Code can add the `accessor` keyword one decorator at a time to migrate gradually](#standard-decorator-migration).
 3. As standard decorators, which need the `accessor` keyword, and TypeScript 5.2 (without the `experimentalDecorators` flag), or Babel 7.23.0.
 
-Each of these configurations is self contained. An application can mix and match libraries written in each of these configurations, and it is not generally a breaking change for a library to switch from one decorator configuration to another.
+Each supported way of using decorators is self contained. An application can mix and match libraries written in each of these configurations, and it is not generally a breaking change for a library to switch from one decorator configuration to another.
 
 To make Lit decorators consistent between both decorator modes there have been some minor breaking
-changes to experimental decorators:
+changes to Lit decorator behavior:
 
-- `@property` and `@state` must be used on the setter for hand-written accessors. The getter no longer works.
-- `requestUpdate()` is called automatically for `@property` and `@state` decorated
+- `@property()` and `@state()` must be used on the setter for hand-written accessors. [Decorating the getter no longer works](#decorated-getter).
+- `requestUpdate()` is called automatically for `@property()` and `@state()` decorated
   accessors where previously that was the setters responsibility.
 - The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
+
+To use the standard decorator configuration, your decorated class fields **must** include the `accessor` keyword.
+To aid in migration from experimental decorators to standard decorators, decorated class fields with the `accessor` keyword will also work with the `experimentalDecorators` flag set to `true`. So users can start migrating code without changing the TS configuration.
 
 ## List of removed APIs
 
@@ -124,4 +129,46 @@ import {hydrate} from 'lit/experimental-hydrate.js';
 // Updated
 import '@lit-labs/ssr-client/lit-element-hydrate-support.js';
 import {hydrate} from '@lit-labs/ssr-client';
+```
+
+### Optional upgrade to standard decorators  {#standard-decorator-migration}
+
+To use standard decorators, your decorators should add the `accessor` keyword.
+Lit 3 decorators are flexible and work as both experimental decorators, and as
+standard decorators.
+
+```ts
+class MyElement extends LitElement {
+  // Lit 3.0 experimental decorators, which are backwards compatible with Lit 2.0
+  @property()
+  myProperty = "initial value"
+
+  // Lit 3.0 adds support for standard decorators.
+  // When using TypeScript:
+  //  - `accessor` keyword is optional when `experimentalDecorators: true`.
+  //  - `accessor` keyword is required when `experimentalDecorators: false`.
+  @property()
+  accessor myProperty = "initial value"
+
+...
+}
+```
+
+### Hand written getters can no longer be reactive properties {#decorated-getter}
+
+`@property()` and `@state()` must be used on the setter for hand-written accessors,
+and will throw a type error when applied on a hand written getter.
+
+```ts
+// This will throw a type error.
+@property()
+get myProperty () { ... }
+
+set myProperty (val) { ... }
+
+// Instead decorate the setter.
+get myProperty () { ... }
+
+@property()
+set myProperty (val) { ... }
 ```
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -125,26 +125,6 @@ import '@lit-labs/ssr-client/lit-element-hydrate-support.js';
 import {hydrate} from '@lit-labs/ssr-client';
 ```
 
-### Optional: To aid migration to standard decorators, decorate hand written setters {#decorated-getter}
-
-When using standard decorators, `@property()` and `@state()` must be used on the
-setter for hand-written accessors, and will throw a type error when applied on a
-hand written getter. Thus we recommend incrementally migrating.
-
-```ts
-// This will throw a type error if using standard decorators.
-@property()
-get myProperty () { ... }
-
-set myProperty (val) { ... }
-
-// Becomes:
-get myProperty () { ... }
-
-@property()
-set myProperty (val) { ... }
-```
-
 ### Optional: upgrade to standard decorators {#standard-decorator-migration}
 
 
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -47,7 +47,7 @@ Changes to Lit decorator behavior in Lit 3.0:
 - `requestUpdate()` is called automatically for `@property()` and `@state()` decorated accessors where previously that was the setters responsibility.
 - The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
 - Lit 3 decorators no longer support the `version: "2018-09"` option of `@babel/plugin-proposal-decorators`. Babel users should [migrate to standard decorators](#standard-decorator-migration).
-- [optional]: [We recommend migrating `@property()` and `@state()` to the setter for hand-written accessors to aid in migrating to standard decorators.](#decorated-getter)
+- [Optional]: [We recommend migrating `@property()` and `@state()` to the setter for hand-written accessors to aid in migrating to standard decorators.](#decorated-getter)
 
 ## List of removed APIs
 
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -42,7 +42,7 @@ While Lit 2 supported TypeScript experimental decorators and Babel's "2018-09" d
 
 The Lit 3 decorators are mostly backwards compatible with the Lit 2 TypeScript decorators - **most likely no changes are needed**.
 
-To make the Lit 2.x decorators consistent between both decorator modes there have been some minor breaking
+Some minor breaking changes were necessary to make the Lit decorators behave consistently between both experimental and standard decorator modes:
 changes to Lit decorator behavior in Lit 3.0:
 
 - `requestUpdate()` is called automatically for `@property()` and `@state()` decorated accessors where previously that was the setters responsibility.
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -40,7 +40,7 @@ changes to Lit decorator behavior in Lit 3.0:
 
 - `requestUpdate()` is called automatically for `@property()` and `@state()` decorated accessors where previously that was the setters responsibility.
 - The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
-- Lit 3 decorators do not support Babel (`@babel/plugin-proposal-decorators`) with `version: "2018-09"`.
+- Lit 3 decorators no longer support `version: "2018-09"` option of `@babel/plugin-proposal-decorators`. Babel users should instead [migrate](#standard-decorator-migration) to use `version: "2023-05"` with plugin version greater than `7.23.0` which follows the TC39 standard decorator spec.
 - [optional]: [We recommend migrating `@property()` and `@state()` the the setter for hand-written accessors to aid in migrating to standard decorators.](#decorated-getter)
 
 ## List of removed APIs
@@ -138,4 +138,26 @@ get myProperty () { ... }
 
 @property()
 set myProperty (val) { ... }
+```
+
+### Optional: upgrade to standard decorators {#standard-decorator-migration}
+
+To use standard decorators, your decorators should add the `accessor` keyword.
+Lit 3 decorators are flexible and work as both experimental decorators, and as
+standard decorators.
+
+```ts
+class MyElement extends LitElement {
+  // Lit 3.0 experimental decorators, which are backwards compatible with Lit 2.0
+  @property()
+  myProperty = "initial value"
+
+  // Lit 3.0 adds support for standard decorators.
+  // When using TypeScript:
+  //  - `accessor` keyword is optional when `experimentalDecorators: true`.
+  //  - `accessor` keyword is required when `experimentalDecorators: false`.
+  @property()
+  accessor myProperty = "initial value"
+...
+}
 ```
--- a/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
+++ b/packages/lit-dev-content/site/docs/v3/releases/upgrade.md
@@ -38,10 +38,10 @@ The Lit 3 decorators are largely backwards compatible. If you've been using them
 To make the Lit 2.x decorators consistent between both decorator modes there have been some minor breaking
 changes to Lit decorator behavior in Lit 3.0:
 
-- `@property()` and `@state()` must be used on the setter for hand-written accessors. [Decorating the getter no longer works](#decorated-getter).
 - `requestUpdate()` is called automatically for `@property()` and `@state()` decorated accessors where previously that was the setters responsibility.
 - The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
 - Lit 3 decorators do not support Babel (`@babel/plugin-proposal-decorators`) with `version: "2018-09"`.
+- [optional]: [We recommend migrating `@property()` and `@state()` the the setter for hand-written accessors to aid in migrating to standard decorators.](#decorated-getter)
 
 ## List of removed APIs
 
@@ -120,19 +120,20 @@ import '@lit-labs/ssr-client/lit-element-hydrate-support.js';
 import {hydrate} from '@lit-labs/ssr-client';
 ```
 
-### Hand written getters can no longer be reactive properties {#decorated-getter}
+### Optional: To aid migration to standard decorators, decorate hand written setters {#decorated-getter}
 
-`@property()` and `@state()` must be used on the setter for hand-written accessors,
-and will throw a type error when applied on a hand written getter.
+When using standard decorators, `@property()` and `@state()` must be used on the
+setter for hand-written accessors, and will throw a type error when applied on a
+hand written getter. Thus we recommend incrementally migrating.
 
 ```ts
-// This will throw a type error.
+// This will throw a type error if using standard decorators.
 @property()
 get myProperty () { ... }
 
 set myProperty (val) { ... }
 
-// Instead decorate the setter.
+// Becomes:
 get myProperty () { ... }
 
 @property()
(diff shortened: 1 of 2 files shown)


packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,64 @@
...
-- Is published as ES2021
-- Removes a few deprecated APIs
+Lit 3.0 has very few breaking changes from Lit 2.0:
+
+- IE11 is no longer supported.
+- Lit's npm modules are now published as ES2021.
+- APIs deprecated with the Lit 2.0 release have been removed.
+- SSR hydration support modules have moved to the `@lit-labs/ssr-client` package.
+- Decorator behavior has been unified between TypeScript experimental decorators and standard decorators.
+- Support was removed for Babel decorators version "2018-09"
```
rictic, line 26: and added support for Babel decorators starting in Babel v7.23.0

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,64 @@
...
+build with `experimentalDecorators` set to `true` or `false` without a change in
+behavior.
+
+To make Lit decorators consistent between both decorator modes there have been some minor breaking
+changes to the experimental decorators:
+
+- `@property` and `@state` must be used on the setter for hand-written accessors. The getter no longer works.
+- `requestUpdate()` is called automatically for `@property` and `@state` decorated
+  accessors where previously that was the setters responsibility.
+- The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
```
rictic, line 51: Some things that I think are missing here:

- TypeScript 5.2 is required for standard decorators
- Babel v7.23.0 is supported with standard decorators

Due to the performance cost of standard decorators at the moment, and with no browsers shipping yet, we recommend most users stick with experimental decorators for the time being.

rictic (changes requested):

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,64 @@
...
 ranges to include both 2.x and 3.x, like `"^2.7.0 || ^3.0.0"`.
 
-Lit 2.x and 3.0 are _interoperable_: templates, base classes, directives,
-decorators, etc., from one version of Lit will work with those from another.
+Lit 2.x and 3.0 are _interoperable_: templates, base classes, and directives from one version of Lit will work with those from another.
+
+## Decorator behavior has been unified
+
+To make the upgrade path from experimental decorators to standard decorators as
+smooth as possible, existing experimental decorators now support the standard
```
rictic, line 37: WDYT about these words:

> In Lit 3.0 our decorators support running in three configurations: 
>
> 1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2.0 users. As standard decorators are still immature, this the most performant option that we recommend for for production for the time being.
> 2. With the TypeScript `experimentalDecorators` flag and with the `accessor` keyword. This aids migration to standard decorators. Code can add the `accessor` keyword one property at a time to migrate gradually.
> 3. With TypeScript 5.2 and without the `experimentalDecorators` flag, or with Babel 7.23.0, and with the `accessor` keyword.
>
> Each of these configurations is self contained. An application can mix and match libraries written in each of these configurations, and it is not generally a breaking change for a library to switch from one decorator configuration to another.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: This list is specifically calling out the breaking changes to existing experimental decorators. I'm not sure those dot points should go as part of this list. Maybe elsewhere in the upgrade guide, or in a dedicated decorator section.

List is introduced with `some minor breaking
changes to the experimental decorators:`

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: I like it! Added it. I think this also includes the information that was proposed in the other comment. Namely that TS 5.2 and Babel 7.23.0 are required.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
 ---
-title: Upgrade guide
+title: Upgrade to Lit 3.x guide
```
augustjk, line 2: nit: wdyt about titling this "Lit 3 upgrade guide" ?
I think just listing the major version looks better than adding ".x"

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
   parent: Releases
   order: 3
 versionLinks:
   v2: releases/upgrade/
 ---
 
-## Overview
+<div class="alert alert-info">
+
+If you are looking to migrate from Lit 1.x to Lit 2.x, see the [Lit 2.0 upgrade guide](/docs/v2/releases/upgrade/).
```
augustjk, line 13: similar to above, can we drop the .x and call it "Lit 2 upgrade guide"?

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
-without deprecation warnings should not require any changes to upgrade to Lit 3.
+</div>
 
-Lit 3.0:
+## Overview
 
-- Drops support for IE11
-- Is published as ES2021
-- Removes a few deprecated APIs
+Lit 3.0 has very few breaking changes from Lit 2.0:
```
augustjk, line 19: I think here 2.x is more appropriate.
```suggestion
Lit 3.0 has very few breaking changes from Lit 2.x:
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
+
+- IE11 is no longer supported.
+- Lit's npm modules are now published as ES2021.
+- APIs deprecated with the Lit 2.0 release have been removed.
```
augustjk, line 23:
```suggestion
- APIs marked deprecated in Lit 2.x releases have been removed.
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
+- SSR hydration support modules have moved to the `@lit-labs/ssr-client` package.
+- Decorator behavior has been unified between TypeScript experimental decorators and standard decorators.
+- Support was removed for Babel decorators version "2018-09"
 
 Lit 3.0 should require no changes to upgrade from Lit 2.0 for the vast majority
```
augustjk, line 28: for consistency's sake, i think unifying to "2.x" in the body of the blog seems better.
```suggestion
Lit 3.0 should require no changes to upgrade from Lit 2.x for the vast majority
```

augustjk (commented):
Some initial comments

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Love it!

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Yes!

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Yes! Thanks!

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Agreed!

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
-decorators, etc., from one version of Lit will work with those from another.
+Lit 2.x and 3.0 are _interoperable_: templates, base classes, and directives from one version of Lit will work with those from another.
+
+## Decorator behavior has been unified
+
+In Lit 3.0 our decorators support running in three configurations:
+
+1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2 users. As standard decorators are still immature, this is the most performant option that we recommend for production for the time being.
+2. With the TypeScript `experimentalDecorators` flag and with the `accessor` keyword. This aids migration to standard decorators. Code can add the `accessor` keyword one decorator at a time to migrate gradually.
+3. As standard decorators, which need the `accessor` keyword, and TypeScript 5.2 (without the `experimentalDecorators` flag), or Babel 7.23.0.
```
augustjk, line 40: I think it's potentially a bit confusing to call this three configurations.

Configuration-wise, it's really 2 cases:
1. TypeScript experimental decorators (with `"experimentalDecorators": true` and `"useDefineForClassFields": false` in tsconfig)
2. TC39 standard decorators with metadata (available with TypeScript 5.2 or Babel 7.23.0)

(Then we should skip to "To make Lit decorators consistent..." with the list of breaking changes.)

To use the standard decorator configuration, `@property()` decorated class fields **must** include the `accessor` keyword. (Link to a new entry in "## Steps to Upgrade" section showing code change needed.)

To aid in migration from experimental decorators to standard decorators, decorated class fields with the `accessor` keyword will also work with the `experimentalDecorators` flag set to true. So users can start migrating code without changing the TS configuration.

(Then we should include the [snippet suggested by Peter](https://github.com/lit/lit.dev/pull/1199/files#r1344552867) about recommending users stick to experimental decorators for the time being.)

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
+
+Each of these configurations is self contained. An application can mix and match libraries written in each of these configurations, and it is not generally a breaking change for a library to switch from one decorator configuration to another.
```
augustjk, line 42: I don't think this statement is necessarily true as written.

Libraries *compiled* with different configurations are interchangeable, i.e. *internal* decorators are fine, but different decorator implementations are not mixable.

So libraries that export decorators, unless they make it hybrid, will incur breaking change when switching from one to another. If a project uses lit decorators and mobx decorators, for instance, they need to choose a config that satisfies both decorator implementations.

I would vote we remove this line.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
+
+To make Lit decorators consistent between both decorator modes there have been some minor breaking
+changes to experimental decorators:
+
+- `@property` and `@state` must be used on the setter for hand-written accessors. The getter no longer works.
```
augustjk, line 47:
```suggestion
- `@property()` and `@state()` must be used on the setter for hand-written accessors. Decorating the getter no longer works.
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
+- `requestUpdate()` is called automatically for `@property` and `@state` decorated
```
augustjk, line 48:
```suggestion
- `requestUpdate()` is called automatically for `@property()` and `@state()` decorated
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,63 @@
...
+In Lit 3.0 our decorators support running in three configurations:
+
+1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2 users. As standard decorators are still immature, this is the most performant option that we recommend for production for the time being.
+2. With the TypeScript `experimentalDecorators` flag and with the `accessor` keyword. This aids migration to standard decorators. Code can add the `accessor` keyword one decorator at a time to migrate gradually.
+3. As standard decorators, which need the `accessor` keyword, and TypeScript 5.2 (without the `experimentalDecorators` flag), or Babel 7.23.0.
+
+Each of these configurations is self contained. An application can mix and match libraries written in each of these configurations, and it is not generally a breaking change for a library to switch from one decorator configuration to another.
+
+To make Lit decorators consistent between both decorator modes there have been some minor breaking
+changes to experimental decorators:
```
augustjk, line 45: It's just a single Lit decorator we're vending (that can run in different modes) so I don't think we should call it "changes to experimental decorators"
```suggestion
changes to Lit decorator behavior:
```

augustjk (commented):
Some more suggestions. I'm happy to make the changes and push to the branch myself too.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
rictic: Yeah, configuraitons is a bit confusing, it suggests a build tool config. I think we should document these three cases though. What about:

```markdown
In Lit 3.0 there are three supported ways to use our decorators:
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
rictic: `configurations` is making this confusing I think. I think it's important that we keep this paragraph. How about this wording:

```markdown
Whether and how it uses the Lit decorators is an internal implementation detail of a Lit element. An application can mix and match libraries written in each of these styles, and it is not generally a breaking change for an element to switch from one decorator style to another.
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
rictic: Good point!

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
rictic: Also, maybe we should put the big idea up front in bold:

```markdown
The Lit 3 decorators are largely backwards compatible. If you've been using them with TypeScript, **most likely no changes are needed**. We've added forward compatibility with standard decorators, but the legacy `experimentalDecorators: true` without the `accessor` keyword is currently the more performant option.
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
augustjk: Ah, I see. You're targeting component authors with this statement. The word "libraries" is throwing me off more so than "configurations" as I think of things like lit and mobx when I hear libraries.

> An application can mix and match ~~libraries written in~~ components built with each of these ~~styles~~ configurations

However, I do think this can be misleading to application developers writing their own components (and possibly their own decorators) because switching decorator config for them is a big deal that affects the entire project.

rictic: Yeah but if you're using other decorators and you change the tsconfig, you'll get a bunch of type errors _for those decorators_. It's a self-documenting problem.

The TypeScript error could definitely be clearer though: https://www.typescriptlang.org/play?&experimentalDecorators=true#code/MYewdgzgLgBApgDwA5wE4EsC2cxQIYA2AInMDALwwAUAUDPTEqiFCAFwwgBGAVqVABo6DMHmwcACsxSooATwDScOTQCUFAHwwA3sPoB6fTCxICxzKZoBfANw0awAnggQYAMRAgdemAAFEMlg4+MSkMABmnhQwAOQAFugxdlZAA

The thing I want to convey here is that an app doesn't have to care about its libraries, and libraries don't have to care about their dependents. This makes the entire thing 10,000% more chill

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Good call. Made change

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Thank you!

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
AndrewJakubowicz: Applied.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
augustjk: Hmm.. While I believe we should definitely mention that `accessor` fields work with experimental decorator config, I don't know if I would call it out in a list of 3 "supported ways" like that. In that sense, writing separate getter setter and decorating the setter is also a "supported way" of using our decorators.

augustjk: However, a valid 3rd *configuration* is
```json
{
  "experimentalDecorators": true,
  "useDefineForClassFields": true
}
```
with decorated fields using `accessor`

AndrewJakubowicz:
I tried to thread the needle on the decorator feedback, to get all the required points across.

Also needed to rebase to get the new changes.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
augustjk: And also speaking of configuration, I think we should make clear that standard decorator support for TypeScript is currently strictly `tsc` only. `esbuild` and `swc` do not yet have all the necessary implementations. This means Vite projects can't have tsconfigs that don't set experimental decorators. Maybe this belongs in the launch blog post, rather than the upgrade guide.. but is certainly another reason to recommend sticking to experimental decorators.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -31,24 +31,29 @@ ranges to include both 2.x and 3.x, like `"^2.7.0 || ^3.0.0"`.
...
 
-1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2 users. As standard decorators are still immature, this is the most performant option that we recommend for production for the time being.
-2. With the TypeScript `experimentalDecorators` flag and with the `accessor` keyword. This aids migration to standard decorators. Code can add the `accessor` keyword one decorator at a time to migrate gradually.
+In Lit 3.0 there are three supported ways to use our decorators:
+
+1. With the TypeScript `experimentalDecorators` flag and properties without the `accessor` keyword. This is backwards compatible with Lit 2.
+2. With the TypeScript `experimentalDecorators` flag and using the `accessor` keyword. This aids migration to standard decorators. [Code can add the `accessor` keyword one decorator at a time to migrate gradually](#standard-decorator-migration).
```
augustjk: Can we add to 1. that `useDefineForClassField` must be false, and in 2. `useDefineForClassField` can be true (default for target 2022 or higher)?
And maybe worth stating `accessor` keyword requires TS 4.9?

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,69 @@
...
+Each supported way of using decorators is self contained. An application can mix and match libraries written in each of these configurations, and it is not generally a breaking change for a library to switch from one decorator configuration to another.
+
+To make Lit decorators consistent between both decorator modes there have been some minor breaking
+changes to Lit decorator behavior:
+
+- `@property()` and `@state()` must be used on the setter for hand-written accessors. [Decorating the getter no longer works](#decorated-getter).
+- `requestUpdate()` is called automatically for `@property()` and `@state()` decorated
+  accessors where previously that was the setters responsibility.
+- The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
+- Lit 3 decorators do not support Babel (`@babel/plugin-proposal-decorators`) with `version: "2018-09"`.
```
augustjk, line 53: I think it would be good to also state the alternative.
```suggestion
- Lit 3 decorators no longer support `version: "2018-09"` option of `@babel/plugin-proposal-decorators`. Babel users should instead [migrate](#standard-decorator-migration) to use `version: "2023-05"` with plugin version greater than 7.23.0 which follows the TC39 standard decorator spec.
```

rictic: We should also say that this requires using the `accessor` keyword too

AndrewJakubowicz:
There is some amazing decorator information being discussed here. I wonder if we're trying to fit all the decorator information into the upgrade guide, where instead we may want to split the information between "upgrading from Lit 2.0", and adding detailed docs to the v3 decorator documentation?

Here in the upgrade guide I think the most important decorator information is the behavior changes to `experimentalDecorators: true`, without `accessor`. Otherwise I think the detailed docs about decorators and the different configurations might be better suited in the v3 decorator docs?

wdyt?

augustjk:
> In a way, the upgrade guide might not even need to mention standard decorators, because users upgrading from Lit 2 don't even need to upgrade to standard decorators for Lit 3 to work well. We even don't want to encourage standard decorators yet.

This is an excellent point and I agree. Focus of the migration guide should be addressing breaking changes, not new features. Info dump can be moved to the decorators page of the docs.

AndrewJakubowicz:
I've done quite a massive change, by focussing down the upgrade guide to only cover the "breaking changes" of experimental decorators.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,57 @@
...
 of users. Most apps and libraries should be able to extend their npm version
 ranges to include both 2.x and 3.x, like `"^2.7.0 || ^3.0.0"`.
 
-Lit 2.x and 3.0 are _interoperable_: templates, base classes, directives,
-decorators, etc., from one version of Lit will work with those from another.
+Lit 2.x and 3.0 are _interoperable_: templates, base classes, and directives from one version of Lit will work with those from another.
+
+## Updates to Lit decorators
+
+The Lit 3 decorators are largely backwards compatible. If you've been using them with TypeScript, **most likely no changes are needed**. We've added forward compatibility with standard decorators, but the legacy `experimentalDecorators: true` without the `accessor` keyword is currently the more performant option.
```
justinfagnani, line 36: I think this could use a bit more context setting, even though we'll repeat this somewhat elsewhere:

```suggestion

JavaScript decorators have recently been standardized by TC39 and are at Stage 3 of the four-stage standardization process. Stage 3 is when JavaScript implementations such as VMs and compilers start implementing the stable specification. TypeScript 5.2 and Babel 7.23 have implemented the standard recently.

This means that there is more than one version of the decorator API in existence: standard decorators, TypeScript's experimental decorators, and previous proposals that Babel has implemented, such as version "2018-09".

While Lit 2 supported TypeScript experimental decorators and Babel's "2018-09" decorators, Lit 3 now supports standard decorators and TypeScript experimental decorators.

The Lit 3 decorators are mostly backwards compatible with the Lit 2 TypeScript decorators - **most likely no changes are needed**.
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,57 @@
...
+
+To make the Lit 2.x decorators consistent between both decorator modes there have been some minor breaking
+changes to Lit decorator behavior in Lit 3.0:
+
+- `requestUpdate()` is called automatically for `@property()` and `@state()` decorated accessors where previously that was the setters responsibility.
+- The value of an accessor is read on first render and used as the initial value for `changedProperties` and attribute reflection.
+- Lit 3 decorators no longer support `version: "2018-09"` option of `@babel/plugin-proposal-decorators`. Babel users should instead [migrate](#standard-decorator-migration) to use `version: "2023-05"` with plugin version greater than `7.23.0` which follows the TC39 standard decorator spec.
```
justinfagnani, line 43:
```suggestion
- Lit 3 decorators no longer support the `version: "2018-09"` option of `@babel/plugin-proposal-decorators`. Babel users should [migrate to standard decorators](#standard-decorator-migration).
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -95,10 +112,52 @@ Experimental hydration support has been moved out of core libraries and into
...
+
+@property()
+set myProperty (val) { ... }
+```
+
+### Optional: upgrade to standard decorators {#standard-decorator-migration}
+
+To use standard decorators, your decorators should add the `accessor` keyword.
+Lit 3 decorators are flexible and work as both experimental decorators, and as
+standard decorators.
```
justinfagnani, line 147:
```suggestion

While Lit 3 adds support for standard decorators, we still recommend that TypeScript users stay with experimental decorators. This is because the emitted code for standard decorators from the TypeScript and Babel compilers is quite large at the moment.

We will recommend standard decorators for production when browsers support them, or when we release decorator transform support in our new Lit compiler.

But you can try standard decorators now, and they work in TypeScript 5.2 and above and Babel 7.23 with the `@babel/plugin-proposal-decorators` plugin.

### Configuration

#### TypeScript

Install TypeScript 5.2 or later, and _remove_ the `"experimentalDecorators"` setting from your tsconfig if it's present.

#### Babel

Install Babel 7.23 or later, and [`@babel/plugin-proposal-decorators`](https://babeljs.io/docs/babel-plugin-proposal-decorators). Be sure to pass the `"version": "2023-05"` option to the plugin.

### Source updates

#### Add the `accessor` keyword to decorated fields.

Standard decorators are not allowed to change the _kind_ of class member they decorate. Decorators that need to create getters and setters must be applied to existing getters and setters. To make this more ergonomic the decorators standard adds the `accessor` keyword which creates "auto accessors" when applied to a class field. Auto accessors look and act much like class fields, but create accessors on the prototype backed by private storage.

```ts
class A {
  accessor foo = 42;
}
```ts

The `@property()`, `@state()`, `@query()`, `@queryAll()`, `@queryAssignedElements()` and `@queryAssignedNode()` decorators require the `accessor` keyword.

Example:
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -95,10 +112,52 @@ Experimental hydration support has been moved out of core libraries and into
...
+```ts
+class MyElement extends LitElement {
+  // Lit 3.0 experimental decorators, which are backwards compatible with Lit 2.0
+  @property()
+  myProperty = "initial value"
+
+  // Lit 3.0 adds support for standard decorators.
+  // When using TypeScript:
+  //  - `accessor` keyword is optional when `experimentalDecorators: true`.
+  //  - `accessor` keyword is required when `experimentalDecorators: false`.
```
justinfagnani, line 158: I would just remove all this and rely on the prose above.
```suggestion
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -95,10 +112,52 @@ Experimental hydration support has been moved out of core libraries and into
...
+  @property()
+  accessor myProperty = "initial value"
+...
+}
+```
```
justinfagnani, line 163:
```suggestion
\```
#### Move decorators from getters to setters

Standard decorators can only replace the class member they're directly applied to. Lit decorators need to intercept property setting, so the decorators must be applied to setters. This is different from the Lit 2 recommendation of applying decorators to getters.

For `@property()` and `@state()` you should also remove any `this.requestUpdate()` calls in the setter since this is done automatically now. If you need to _not_ have `requestUpdate()` called, you'll have to use the `hasChanged` property option.

Note that for `@property()` and `@state()` the _getter_ will be called by the decorator when setting the property to retrieve the old value. This means that you _must_ define both a getter and a setter.

Before:
```ts
  private _foo = 42;
  set(v) {
    const oldValue = this._foo;
    this._foo = v;
    this.requestUpdate('foo', oldValue);
  }
  @property()
  get() {
    return this._foo;
  }
}
\```

After:
```ts
  private _foo = 42;
  @property()
  set(v) {
    this._foo = v;
  }
  get() {
    return this._foo;
  }
}
\```
```

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -1,40 +1,57 @@
...
+## Overview
 
-Lit 3.0:
+Lit 3.0 has very few breaking changes from Lit 2.x:
 
-- Drops support for IE11
-- Is published as ES2021
-- Removes a few deprecated APIs
+- IE11 is no longer supported.
+- Lit's npm modules are now published as ES2021.
```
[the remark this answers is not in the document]
AndrewJakubowicz: Yes, but out of scope for this PR. Will handle separately.

AndrewJakubowicz:
This is ready for re-review!
Thank you!

Preview: https://pr1199-2799c11---lit-dev-5ftespv5na-uc.a.run.app/docs/v3/releases/upgrade/

augustjk (approved):
Looking good. Just a few minor changes.

PR #1202, opened 2023-10-02
[test PR] test revert wireit update

AndrewJakubowicz:
PR to disprove the hypothesis that bumping wireit caused the GitHub actions


AndrewJakubowicz:
Doesn't fix, but does give some nice error logs which confirm the issue is in `BlockingRender`.

Issue #1181, opened 2023-09-06
[docs] Property expression section does not provide a concrete example how to access JavaScript property in Child Web Component


projektorius96:
In reference to [Property Expression](https://lit.dev/docs/templates/expressions/#property-expressions) please provide a concrete example depicting the access to the JavaScript property in Child Web Component ; elaborating on what is given by the documentation at section [Property Expression], consider the following:

```html
html`<my-list .listItems=${this.items}></my-list>`;
```

In accordance to [this answer on stackoverflow](https://stackoverflow.com/questions/63945434/how-to-send-props-from-parent-to-child-with-litelement) the JavaScript property `.listItems` passed to custom element `my-list` in Child Web Component should be access as follows:

```js
class MyList extends LitElement {
 /* ... */
  render() {
    return html `<p>${this.listItems}</p>`
   }
/* ... */
}
customElements.define('my-list ', MyList);
```

This should be clearly outlined from day number one, without forcing to end up on sites like StackOverflow ...

Thank you.

projektorius96:
Up.

PR #1204, opened 2023-10-04
[docs] Show exactly how a property expression behaves imperatively

AndrewJakubowicz:
Part of: https://github.com/lit/lit.dev/issues/1181

It is unclear what the property expression syntax actually does. This change adds explicit documentation to show the imperative alternative to property expression binding which should help developers build a mental model of property bindings.

This documents some questions that have come up in the past.

Thank you!


augustjk (approved):
LGTM

PR #1211, opened 2023-10-09
Fixes to task and context samples

justinfagnani:
- Changes the task sample name to just "Async Tasks" so it doesn't wrap in the samples list
- Uses non-labs packages in sampels
- Adds an initial state to the task example
- Cleans up some variable names
- Uses HTML templates in other task render callbacks so it's more obvious that you can


augustjk (approved):
Nice!

packages/lit-dev-content/samples/examples/async-task/npm-info.ts
```diff
@@ -1,57 +1,73 @@
...
           alt="npm logo"
         />
       </header>
       <div>
-        ${this._fetchNpmInfoTask.render({
-          pending: () => 'Loading...',
+        ${this._npmInfoTask.render({
+          initial: () =>
+            html`<span class="initial">
+              Enter a package name to display it's npm info
```
rictic, line 47: it's -> its

packages/lit-dev-content/samples/examples/async-task/npm-info.ts
```diff
@@ -1,57 +1,73 @@
...
+   *
+   * The current state of the task is rendered below with
+   * `this._npmInfoTask.render()`.
+   */
+  private _npmInfoTask = new Task(this, {
     task: async ([pkgName], {signal}) => {
       if (pkgName === undefined || pkgName === '') {
-        throw new Error('Empty package name');
+        // This puts the task back into the INITIAL state
+        return initialState;
```
rictic, line 22: ah, nice!

rictic (approved):

packages/lit-dev-content/samples/examples/async-task/npm-info.ts
AndrewJakubowicz: This is really nice!

packages/lit-dev-content/samples/examples/async-task/npm-info.ts
```diff
@@ -1,57 +1,73 @@
...
       }
       return await fetchPackageInfo(pkgName, signal);
```
augustjk, line 24: To leave a record here, @rictic had a nice suggestion to do a debounce here by adding
```ts
// debounce
await new Promise((r) => setTimeout(r, 50));
signa.throwIfAbort();
```
Then the task can be triggered by calling on `input` event rather than `change` event.

PR #1217, opened 2023-10-10
Update bundle URLs to `3`.

bicknellr:
These are out of date as of today! 🎉 I specifically changed them to `3.0.0` rather than just `3` so that users that copy these URLs don't get surprised by the content changing out from under them when new minor and patch versions are released. If there's a way to dynamically reference the latest full version string, let me know and I'll update to that instead.

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/getting-started.md
+++ b/packages/lit-dev-content/site/docs/v3/getting-started.md
@@ -76,7 +76,7 @@ are two types of bundles:
   <dt class="paramName">core</dt>
   <dd class="paramDetails">
     <a href="https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/core/lit-core.min.js">
-      https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js
+      https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/core/lit-core.min.js
     </a>
     <br>
     <code>core</code> exports the same items as
@@ -87,7 +87,7 @@ are two types of bundles:
   <dt class="paramName">all</dt>
   <dd class="paramDetails">
     <a href="https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/all/lit-all.min.js">
-      https://cdn.jsdelivr.net/gh/lit/dist@2/all/lit-all.min.js
+      https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/all/lit-all.min.js
     </a>
     <br>
     <code>all</code> exports everything in <code>core</code> plus
--- a/packages/lit-dev-content/site/docs/v3/getting-started.md
+++ b/packages/lit-dev-content/site/docs/v3/getting-started.md
@@ -55,7 +55,7 @@ standard JavaScript modules with no dependencies - any modern browser should be
 able to import and run the bundles from within a `<script type="module">` like this:
 
 ```js
-import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/core/lit-core.min.js';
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz (approved):

packages/lit-dev-content/site/docs/v3/getting-started.md
```diff
@@ -75,7 +75,7 @@ are two types of bundles:
 <dl class="params">
   <dt class="paramName">core</dt>
   <dd class="paramDetails">
-    <a href="https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js">
+    <a href="https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/core/lit-core.min.js">
       https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js
```
bicknellr, line 79: I didn't update the actual link text!

packages/lit-dev-content/site/docs/v3/getting-started.md
```diff
@@ -86,7 +86,7 @@ are two types of bundles:
 
   <dt class="paramName">all</dt>
   <dd class="paramDetails">
-    <a href="https://cdn.jsdelivr.net/gh/lit/dist@2/all/lit-all.min.js">
+    <a href="https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/all/lit-all.min.js">
       https://cdn.jsdelivr.net/gh/lit/dist@2/all/lit-all.min.js
```
bicknellr, line 90: Also here

AndrewJakubowicz (approved):
They work great! Thank you for submitting this!

packages/lit-dev-content/site/docs/v3/getting-started.md
AndrewJakubowicz: oh oops

AndrewJakubowicz (commented):
It may be worth adding these to the `known-good-urls.txt` as well

packages/lit-dev-content/site/docs/v3/getting-started.md
bicknellr: fixed

packages/lit-dev-content/site/docs/v3/getting-started.md
bicknellr: also fixed

bicknellr:
> It may be worth adding these links to the `known-good-urls.txt` as well

Added

AndrewJakubowicz (approved):

packages/lit-dev-content/site/docs/v3/getting-started.md
```diff
@@ -55,7 +55,7 @@ standard JavaScript modules with no dependencies - any modern browser should be
 able to import and run the bundles from within a `<script type="module">` like this:
 
 ```js
-import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@2/core/lit-core.min.js';
+import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@3.0.0/core/lit-core.min.js';
```
augustjk, line 58: Do we maybe want to just specify major version here so users automatically get patches and minor non-breaking updates?
```suggestion
import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js';
```

augustjk (changes requested):
Thank you for catching this!

Maybe we should specify `@3` rather than `@3.0.0` so it stays relevant through the whole of Lit 3?

augustjk (approved):
Thank you!

PR #1219, opened 2023-10-11
docs: fix lit/react link

slorber:
As far as I know lit React graduated and the link seems to point to the old lab package instead of the new stable React package


AndrewJakubowicz (approved):
Thank you!

PR #1222, opened 2023-10-11
add v3 page redirects to each of the v1 pages

AndrewJakubowicz:
Currently these are not auto generated, so must be manually provided.

### Why

On google search, frustratingly some links still navigate to `v1` documentation.

For example, if you search: "firstUpdated", the very first link is to: `https://lit.dev/docs/v1/components/lifecycle/`.

Then when you press `v3` it doesn't take you to the right page.

### Fix

Map the v1 pages to their respective v3 pages.

### Test plan

Tested manually.

### Future work

We should consider a system which can figure out automatically how to skip versions.
We only need to provide `v1 -> v2`, and `v2 -> v3`, to compute the path from `v1 -> v3`.

### DIFF
--- a/packages/lit-dev-content/site/docs/v1/components/decorators.md
+++ b/packages/lit-dev-content/site/docs/v1/components/decorators.md
@@ -6,6 +6,7 @@ eleventyNavigation:
   order: 6
 versionLinks:
   v2: components/decorators/
+  v3: components/decorators/
 ---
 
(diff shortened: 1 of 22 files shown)


augustjk (approved):
Good catch and thank you!

augustjk:
Agreed that being able to transitively get v1 > v3 would be nice, but likely won't be needed until we do v4 docs...

Now on a separate note do we want to get rid of old docs from showing up in google search results? Maybe we want to add `<meta>` with `noindex` for older version pages..

PR #1223, opened 2023-10-13
fix: remove errant semi-colon after decorator

heyMP:
Fixes the `Declaration expected.` TS error when running demo.


justinfagnani (approved):
thanks!

PR #1225, opened 2023-10-13
prevent version selector being autocompleted by the browser

AndrewJakubowicz:
This fixes an issue where when you navigate back in history, the browser pre-fills the select with the value from just after you navigated.

Repro:

1. Navigate to lit.dev.
2. Select `v1`, and be navigated to the v1 docs.
3. Press back in history, and note the select is set to v1.

### Why is the fix worth it

This prevents confusion on navigating backwards, and having the version dropdown not reflect the page that you're on.

### Tradeoff

Safari `autocomplete="off"` doesn't work. And the workaround for Safari is a lot more involved.
This PR is an incremental improvement for Firefox and Chrome.

Tested manually across Chrome/Firefox/Safari.

### DIFF
--- a/packages/lit-dev-content/site/_includes/site-nav.html
+++ b/packages/lit-dev-content/site/_includes/site-nav.html
@@ -2,7 +2,7 @@
    <a href="{{ site.baseurl }}/docs/">Documentation</a>
 
    <litdev-version-selector>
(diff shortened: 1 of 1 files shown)


augustjk (approved):
Thank you!

PR #1227, opened 2023-10-16
fix: removed question mark from attribute assignment

paglobal:
With the question mark present, I was getting errors in the playground.


augustjk (approved):
Thank you for catching this! I added the fix to the v2 doc as well.

paglobal:
> Thank you for catching this! I added the fix to the v2 doc as well.

Nice!

Issue #1201, opened 2023-10-02
[infra] GitHub actions no longer completes a build or test


AndrewJakubowicz:
### Context

As of a couple days ago, all GitHub actions now fail on any new PR on this repo.

<img width="705" alt="Screenshot 2023-10-02 at 1 38 09 PM" src="https://github.com/lit/lit.dev/assets/15080861/81fee66f-4297-4e81-82f1-26a69a5d842f">

The site still deploys, and preview URLs still work, so this does not block content changes.
However, it increases the risk of infrastructure changes.

Augustine did some investigating as a possible starting point (rephrased from chat):

```
There is a `BlockingRender` that exists (which makes some async work sync). It uses `Atomic.wait`, and `Atomics.notify`. At some point it may be that there are so many things happening, that there are not enough cycles for the worker thread to pick up the render request. They hang until one times out.
```

AndrewJakubowicz:
Theory is that the very slow build time of lit.dev is due to the special syntax highlighting that is done.

I changed the `dev:build` command to `0x ../../node_modules/@11ty/eleventy/cmd.js --quiet` which runs eleventy and generates a flamegraph using [0x](https://www.npmjs.com/package/0x).

Here is the full flamegraph of building the site:
![Screenshot 2023-10-18 at 10 23 41 AM](https://github.com/lit/lit.dev/assets/15080861/c1ec9086-a7b4-4df7-88cd-c38c99fadc0c)

If we color all the work that uses playwright to emulate a browser, we can see a huge amount of the work is playwright related:
![Screenshot 2023-10-18 at 10 24 48 AM](https://github.com/lit/lit.dev/assets/15080861/9d2e2866-78b8-44fa-902c-a62ced31aa76)

And if we filter for `renderer.ts` or `blocking-renderer.ts`, this also reveals the length of time spent syntax highlighting:
![Screenshot 2023-10-18 at 10 27 47 AM](https://github.com/lit/lit.dev/assets/15080861/6c9cafe7-89de-4f78-8ec3-b8108fb2163b)




### Some rough numbers (on my M2 silicon mac)

Note the table should be read from top to bottom, and I am removing features using `if (!DEV) { ... }` flag.

| Features    | Time |
| -------- | ------- |
| Current dev:build time  | `17.24 seconds (46.2ms each)`    |
| Without blocking renderer |   `11.28 seconds (30.2ms each)`  |
| Without search index creation | `7.97 seconds (21.4ms each)`

[...]

PR #1228, opened 2023-10-18
remove search index generation from dev build

AndrewJakubowicz:
Part of https://github.com/lit/lit.dev/issues/1201

### Context

While profiling lit.dev, I found this as an easy performance win to make dev mode builds faster.

In the old days, the Lit.dev search was done locally, so the search index could be used and debugged locally. However now the search index is instead deployed to Algolia. Thus the several seconds it takes to create the search index can be skipped when building the site in dev mode.

### Results

Before this change, tested with `dev:build`:

`[11ty] Wrote 373 files in 15.58 seconds (41.8ms each, v2.0.1)`

After this change:

`[11ty] Wrote 373 files in 12.15 seconds (32.6ms each, v2.0.1)`

### DIFF
--- a/packages/lit-dev-content/.eleventy.js
+++ b/packages/lit-dev-content/.eleventy.js
@@ -580,7 +580,11 @@ ${content}
     );
     await Promise.all(emptyDocsIndexFiles.map((path) => fs.unlink(path)));
 
-    await createSearchIndex(ENV.eleventyOutDir);
+    if (!DEV) {
+      // Only create the search index in production when we'll be uploading it
+      // to algolia.
+      await createSearchIndex(ENV.eleventyOutDir);
+    }
 
     if (DEV) {
       // Symlink css, images, and playground projects. We do this in dev mode


packages/lit-dev-content/.eleventy.js
```diff
@@ -580,7 +580,11 @@ ${content}
     );
     await Promise.all(emptyDocsIndexFiles.map((path) => fs.unlink(path)));
 
-    await createSearchIndex(ENV.eleventyOutDir);
+    if (!DEV) {
+      // Only create the search index in production when we'll be uploading it
+      // to algolia.
+      await createSearchIndex(ENV.eleventyOutDir);
+    }
```
augustjk, line 587: Hm... can we just remove this all together? We don't actually use the locally created index file anymore as it's always going to algolia.

augustjk: Aha thanks for clarifying!

augustjk (approved):

PR #1229, opened 2023-10-18
Optimize dev edit-refresh loop by caching the highlighted files

AndrewJakubowicz:
Issue: https://github.com/lit/lit.dev/issues/1201

### Context

This PR focusses exclusively on faster edit-refresh dev mode times.

Result is 11.3s edit-refresh time (from 16.7s edit-refresh time) on my M2 silicon mac.
Or, 34.4s edit-refresh time (from 57.3s ) on intel mac.

### How

Instead of using the playwright browser process to always retrieve the syntax highlighted code, instead use a temporary dev cache so syntax highlighted code blocks can be retrieved from the file system.

This ends up being more optimal than running the code through the browser (even using synchronous file system calls).

Note: synchronous file system calls must be used as the whole API must be sync for 11ty.

### Test plan

Manually tried it with edit refresh.
Note, there is a risk of a collision. In this case the code may appear wrong **only in dev mode** where caching is enabled.

### DIFF
--- a/packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
+++ b/packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
@@ -138,21 +138,22 @@ export class BlockingRenderer {
     });
   }
 
-  private getCachedRender(lang: string, code: string): string | null {
-    if (!this.isDevMode) {
-      return null;
-    }
-    const fileName = createUniqueFileNameKey(lang, code);
-    const absoluteFilePath = pathlib.resolve(cachedHighlightsDir, fileName);
+  private getCachedRender(cachedFileName: string): string | null {
+    const absoluteFilePath = pathlib.resolve(
+      cachedHighlightsDir,
+      cachedFileName
+    );
     if (fs.existsSync(absoluteFilePath)) {
       return fs.readFileSync(absoluteFilePath, {encoding: 'utf8'});
     }
     return null;
   }
 
-  private writeCachedRender(lang: string, code: string, html: string) {
-    const fileName = createUniqueFileNameKey(lang, code);
-    const absoluteFilePath = pathlib.resolve(cachedHighlightsDir, fileName);
+  private writeCachedRender(cachedFileName: string, html: string) {
+    const absoluteFilePath = pathlib.resolve(
+      cachedHighlightsDir,
+      cachedFileName
+    );
     fs.writeFileSync(absoluteFilePath, html);
   }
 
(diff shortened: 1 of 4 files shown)


packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
```diff
@@ -70,6 +121,11 @@ export class BlockingRenderer {
       htmlBuffer: this.sharedHtml,
       notify: this.sharedNotify,
     });
+    try {
+      fs.mkdirSync(cachedHighlightsDir);
+    } catch {
+      // Directory already exists.
+    }
```
augustjk, line 128: Nit: Just in case not to swallow other errors?
```suggestion
    try {
      fs.mkdirSync(cachedHighlightsDir);
    } catch (error) {
      if (!(error as {code: string}).code !== 'EEXIST') {
        throw error;
      }
    }
```

packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
```diff
@@ -82,7 +138,44 @@ export class BlockingRenderer {
...
+  }
+
   render(lang: 'js' | 'ts' | 'html' | 'css', code: string): {html: string} {
+    if (!this.isDevMode) {
+      // In production, skip all caching.
+      return this.renderWithWorker(lang, code);
+    }
+    // In dev mode, speed up the edit-refresh loop by caching the syntax
+    // highlighted code.
+    const cachedResult = this.getCachedRender(lang, code);
```
augustjk, line 166: Nit: Can we create unique filename key above this, and have both `getCachedRender()` and `writeCachedRender()` accept the key instead of lang & code so we don't need to redo the hash?

augustjk (approved):
Only a couple nits. This is awesome!

packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
AndrewJakubowicz: Yes! Excellent performance based catch! definitely will do that.

packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
augustjk: Alternatively...
```suggestion
    try {
      fs.mkdirSync(cachedHighlightsDir);
    } catch (error) {
      if ((error as {code: string}).code === 'EEXIST') {
        // Directory already exists.
      } else {
        throw error;
      }
    }
```

AndrewJakubowicz: I like the latter that throws! Thank you!

packages/lit-dev-tools-cjs/src/playground-plugin/blocking-renderer.ts
AndrewJakubowicz: Done

augustjk (approved):

PR #1230, opened 2023-10-18
[examples] Update examples to use Lit 3

augustjk:
- Updated v3 base project json for to use Lit 3 stable (it was on prerelease).
- Update hard-coded versions in project/package.json for motion and react examples.
- Update react examples to use graduated package, and update the section heading to drop labs.


AndrewJakubowicz (approved):
Amaze! Thank you :D

PR #1231, opened 2023-10-19
chore: tighten the wireit files dependency on building unversioned docs command

AndrewJakubowicz:
### Context

Instead of watching all files under `/docs/`, the `build:unversioned-docs` command only needs to look at the files within the latest version directory.

This will result in slightly less work.

### Test plan

Unversioned docs checked in preview URL.


augustjk (approved):
Good catch!

Issue #1232, opened 2023-10-20
[example] JS version of `@lit/context` does not work


augustjk:
Related: https://github.com/lit/lit/issues/4313

Until we can do above, I believe we can provide a working JS version manually.

PR #1233, opened 2023-10-20
[examples] Add bespoke JS version of Context Basics

augustjk:
Fixes https://github.com/lit/lit.dev/issues/1232

Preview: https://pr1233-15216bd---lit-dev-5ftespv5na-uc.a.run.app/playground/#sample=examples/context-basics
(toggle to JS)

### DIFF
--- a/packages/lit-dev-content/samples/examples/context-basics/elements.js
+++ b/packages/lit-dev-content/samples/examples/context-basics/elements.js
@@ -30,7 +30,6 @@ export class ProviderEl extends LitElement {
   set data(value) {
     this._data = value;
     this._provider.setValue(value);
-    this.requestUpdate();
   }
 
   get data() {
--- a/packages/lit-dev-content/samples/examples/context-basics/elements.js
+++ b/packages/lit-dev-content/samples/examples/context-basics/elements.js
@@ -1,0 +1,64 @@
+import {html, LitElement} from 'lit';
+import {ContextProvider, ContextConsumer, createContext} from '@lit/context';
+
+import {providerStyles} from './styles.js';
+
+const contextKey = Symbol('contextKey');
+
+// Context object, which acts like a key for the context.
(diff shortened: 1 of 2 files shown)


packages/lit-dev-content/samples/examples/context-basics/elements.js
```diff
@@ -0,0 +1,64 @@
...
+   */
+  set data(value) {
+    this._data = value;
+    this._provider.setValue(value);
+    this.requestUpdate();
+  }
+
+  get data() {
+    return this._data;
+  }
```
augustjk, line 38: I considered doing
```js
  willUpdate(changedProperties) {
    if (changedProperties.has('data')) {
      this._provider.setValue(this.data);
    }
  }
```
but there was a subtle timing issue that `willUpdate()` will get called after the consumer is registered so the initial value the consumer gets could never be the data set as attribute on the `<provider-el>`. The subscriber would have to consume with `{subscribe: true}` to get the same behavior as the TS version.

AndrewJakubowicz: That's interesting!

packages/lit-dev-content/samples/examples/context-basics/elements.js
```diff
@@ -0,0 +1,64 @@
...
+    this.data = 'Initial';
+  }
+
+  /**
+   * `data` will be provided to any consumer that is in the DOM tree below it.
+   */
+  set data(value) {
+    this._data = value;
+    this._provider.setValue(value);
+    this.requestUpdate();
```
AndrewJakubowicz, line 33: With Lit 3, I think `requestUpdate` will always be called.

augustjk: Nice catch! Fixed!

PR #1235, opened 2023-10-20
[doc] Update context docs to reflect preferred options syntax for controllers

augustjk:
The multiple argument signature for the controller constructors are marked deprecated in favor of the options object. Our docs should reflect that.


AndrewJakubowicz (approved):
Cheers!

PR #1244, opened 2023-11-01
[docs] Update docs regarding class field and reactive properties

### LINKED ISSUE #1243: [doc] Add `accessor` solution option for solving the reactive updates class field error
When using TypeScript, if `useDefineForClassFields` is `true` (Note default is `true` if [target](https://www.typescriptlang.org/tsconfig#target) is ES2022 or higher, including ESNext; `false` otherwise) 

Then setting class fields will throw an error.

E.g.

```ts
@property() greeting = 'welcome';
```

Throws:

```
The following properties on element my-element will not trigger updates as expected because they are set using class fields: docsHint, count. Native class fields and some compiled output will overwrite accessors used for detecting changes. See https://lit.dev/msg/class-field-shadowing for more information
```

### Issue

The error link: https://lit.dev/msg/class-field-shadowing doesn't document that a possible solution is to use the `accessor` keyword (which is forward compatible with standard decorators).

This is a valid solution in Lit 3 with hybrid decorators.

### Proposal 

It should be documented that the error can be solved with

```
@property() accessor greeting = 'welcome';
```

This code works with all combinations of `experimentalDecorators` and `useDefineForClassFields` and is forward compatible with standard decorators.

augustjk:
Fixes #1243 

- From decorator doc, removed the separate section on class field issue, instead linking to reactive properties doc next to TS `useDefineForClassFields: false` sentence.
- For reactive properties doc
  - removed Babel plugin-proposal-class-properties solution.
  - added solution of using `accessor` keyword to avoid instance class field shadowing.
  - added code samples showing the code that breaks reactivity, as well as what each solution looks like.

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/components/properties.md
+++ b/packages/lit-dev-content/site/docs/v3/components/properties.md
@@ -120,34 +120,77 @@ An empty option object is equivalent to specifying the default value for all opt
 
 ### Avoiding issues with class fields when declaring properties {#avoiding-issues-with-class-fields}
 
-[Class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields) have a problematic interaction with reactive properties. Class fields are defined on the element instance. Reactive properties are defined as accessors on the element prototype. According to the rules of JavaScript, an instance property takes precedence over and effectively hides a prototype property. This means that reactive property accessors do not function when class fields are used. When a property is set, the element does not update.
-
-In **JavaScript** you **must not use class fields** when declaring reactive properties. Instead, properties must be initialized in the element constructor:
+[Class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields) have a problematic interaction with reactive properties. Class fields are defined on the element instance whereas reactive properties are defined as accessors on the element prototype. According to the rules of JavaScript, an instance property takes precedence over and effectively hides a prototype property. This means that reactive property accessors do not function when class fields are used such that setting the property won't trigger an element update.
 
 ```js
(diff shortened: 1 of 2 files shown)


packages/lit-dev-content/site/docs/v3/components/properties.md
```diff
@@ -120,34 +120,77 @@ An empty option object is equivalent to specifying the default value for all opt
...
+    super();
+    this.foo = 'Default';
+  }
 }
 ```
 
-For **TypeScript**, you **may use class fields** for declaring reactive properties as long as you use one of these patterns:
-* Set the `useDefineForClassFields` setting in your `tsconfig` to `false`. Note, this is not required for some configurations of TypeScript, but it's recommended to explicitly set it to `false`.
-* Add the `declare` keyword on the field, and put the field's initializer in the constructor.
+Alternatively, you may use [standard decorators with Babel](/docs/v3/components/decorators/#decorators-babel) to declare reactive properties.
```
AndrewJakubowicz, line 143: Technically, you can also use the auto accessor solution in JavaScript if you do not want the decorator piece, but do not want to initialize the fields in the element constructor.

E.g.
```
static properties = {field: {}};
accessor field = '...';
```

AndrewJakubowicz (approved):
Awesome! Thank you so much for a huge improvement to the docs!

Checked that our class field shadowing error points to this improved section (https://lit.dev/msg/class-field-shadowing).

Only nit is whether we want to include the `accessor` solution in the JavaScript section.

packages/lit-dev-content/site/docs/v3/components/properties.md
AndrewJakubowicz: Happy with that!

PR #1253, opened 2023-11-03
[learn] update tutorial thumbnail to match Lit visual identity

AndrewJakubowicz:
### Context

The visual identity of the Lit brand doesn't contain large gradients, so I think it makes the existing thumbnail stand out.

### Fix

I made two changes:
 - I removed the background radial gradient and flattened it to a similar color.
 - I added the "Lit noise" texture subtly over the icons to add visual interest.

I think this change adds cohesion to the images on the /learn page as everything remains flat, and the bit of noise ties to other video thumbnails which contain the subtle noise texture.

Thank you!


augustjk (approved):
Ooh GitHub is back. I love this! The lighter color without the gradient is much easier on my eyes.

PR #1254, opened 2023-11-03
[learn] fix error being logged on learn page

AndrewJakubowicz:
### Context

Fixes an error being thrown in the console: 

`GET https://lit.dev/images/learn/banner-blue.png 404 (Not Found)`

### Fix

We no longer use any background image because all cards contain their own explicit image. Remove the CSS that is triggering the load.

### Impact

I believe this gets the page to a lighthouse score of all 100's.


augustjk (approved):
💯

AndrewJakubowicz:
I bungled the commit message on merge. Whoops.

Issue #455, opened 2021-08-23
Add section to Tools:Requirements about multiple versions


sorvell:
Follow on to https://github.com/lit/lit/pull/2075#discussion_r693685111. As noted there, we want to cover:

1. Multiple versions should work (new feature in Lit 2)
1. We don't recommend it because while Lit is small it means you're shipping extra bytes
1. You should try npm dedupe and recommend components be published without bundling their dependencies and with as wide of version constraints as possible.

PR #1261, opened 2023-11-08
[docs] Add multiple versions of lit resolution section

AndrewJakubowicz:
Fixes: https://github.com/lit/lit.dev/issues/455

### Context

We have a developer warning that warns when multiple versions of Lit have been loaded. This doesn't currently link to documentation that gives clear actionable steps for resolving the warning.

This results in us needing to resolve the issues on a case by case basis via Discord.

### Fix

Add a section that discusses the multiple versions of Lit warning, as well as some resolution sections.

### Test plan

Docs only change. Can be tested by navigating to:

https://pr1261-1d82c04---lit-dev-5ftespv5na-uc.a.run.app/msg/multiple-versions
https://pr1261-1d82c04---lit-dev-5ftespv5na-uc.a.run.app/docs/tools/requirements/#multiple-lit-versions

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/tools/development.md
+++ b/packages/lit-dev-content/site/docs/v3/tools/development.md
@@ -91,6 +91,8 @@ If you’re publishing a library that uses Lit, follow our [publishing best prac
 
 ##### Resolving multiple versions of Lit
 
+It is possible to follow the steps below, and not be able to de-duplicate Lit, e.g., a library you depend on is bundling a specific version of Lit. In these cases the warning can be ignored.
+
 If you’re seeing a `Multiple versions of Lit loaded` development mode warning, there are a couple things you can try:
 
 1. Find out which Lit libraries have multiple versions loaded by checking the following variables in your browser console: `window.litElementVersions`, `window.reactiveElementVersions`, and `window.litHtmlVersions`.
@@ -99,9 +101,9 @@ If you’re seeing a `Multiple versions of Lit loaded` development mode warning,
 
 3. Try to use `npm dedupe` to de-duplicate Lit. Use `npm ls` to verify if the duplicated Lit package was successfully de-duped.
 
-4. If there is still duplication, you may need to delete your package lock and `node_modules`. Then install the version of `lit` you want explicitly, followed by your dependencies.
+4. It is possible to nudge `npm` to hoist particular versions of the core Lit packages by installing them as direct dependencies of your project with `npm i @lit/reactive-element@latest lit-element@latest lit-html@latest`. Replace `latest` with the version you want to de-dupe to.
 
-It is possible to follow these steps, and not be able to de-duplicate Lit, e.g., a library you depend on is bundling a specific version of Lit. In these cases the warning can be ignored.
+5. If there is still duplication, you may need to delete your package lock and `node_modules`. Then install the version of `lit` you want explicitly, followed by your dependencies.
 
 If you want to silence the warning, add the following code early in your application before Lit has been loaded to initialize the global `litIssuedWarnings` set, and preemptively add the warning to the set.
 
--- a/packages/lit-dev-content/site/docs/v3/tools/development.md
+++ b/packages/lit-dev-content/site/docs/v3/tools/development.md
@@ -107,17 +107,6 @@ If you’re seeing a `Multiple versions of Lit loaded` development mode warning,
 
 5. If there is still duplication, you may need to delete your package lock and `node_modules`. Then install the version of `lit` you want explicitly, followed by your dependencies.
 
-If you want to silence the warning, add the following code early in your application before Lit has been loaded to initialize the global `litIssuedWarnings` set, and preemptively add the warning to the set.
-
-```html
-<script>
-  window.litIssuedWarnings = new Set();
-  // By adding the Lit warning to this global set, Lit will no longer emit the multiple versions warning.
-  window.litIssuedWarnings.add(
-    'Multiple versions of Lit loaded. Loading multiple versions is not recommended. See https://lit.dev/msg/multiple-versions for more information.'
-  );
-</script>
-```
 
 ## Local dev servers { #devserver }
 
--- a/packages/lit-dev-content/site/docs/v3/tools/development.md
+++ b/packages/lit-dev-content/site/docs/v3/tools/development.md
@@ -78,6 +78,43 @@ class MyElement extends LitElement {
 
 It's best for code size if the code to control warnings is eliminated in your own production builds.
 
+#### Multiple versions of Lit warning {#multiple-lit-versions}
+
+If Lit is being used as an internal dependency of elements, elements can use different versions of Lit and are completely interoperable.
+We also take care to ensure that Lit 2 and Lit 3 are mostly compatible with each other. For example, you can pass a Lit 2 template into a Lit 3 render function and vice-versa.
+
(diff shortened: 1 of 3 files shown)


packages/lit-dev-content/site/docs/v3/tools/requirements.md
```diff
@@ -49,3 +49,38 @@ All modern browsers update automatically and users are highly likely to have a r
...
+If the warning is distracting and you want to silence it, add the following code early in your application before Lit has been loaded to initialize the global `litIssuedWarnings` set, and preemptively add the warning to the set.
+
+```html
+<script>
+  window.litIssuedWarnings = new Set();
+  // By adding the Lit warning to this global set, Lit will no longer emit the multiple versions warning.
+  window.litIssuedWarnings.add(
+    'Multiple versions of Lit loaded. Loading multiple versions is not recommended. See https://lit.dev/msg/multiple-versions for more information.'
+  );
+</script>
```
augustjk, line 85: This feels very hacky. TIL that we have a way to turn off warnings for `ReactiveElement` https://lit.dev/docs/tools/development/#development-build-runtime-warnings, so perhaps we should consider adding this globally so we can suppress `lit-html` warnings too. That is, if we really deem the 'multiple-versions' warning to be optional.

packages/lit-dev-content/site/docs/v3/tools/requirements.md
```diff
@@ -49,3 +49,38 @@ All modern browsers update automatically and users are highly likely to have a r
...
+
+If you’re seeing a `Multiple versions of Lit loaded` development mode warning, there are a couple things you can try.
+
+1. Find out which Lit libraries have multiple versions loaded by checking the following variables in your browser console: `window.litElementVersions`, `window.reactiveElementVersions`, and `window.litHtmlVersions`.
+
+2. Use `npm ls` (note, you can specify exact libraries to look for, e.g. `npm ls @lit/reactive-element`) to narrow down which dependencies are loading multiple different versions of Lit.
+
+3. Try to use `npm dedupe` to de-duplicate Lit. Use `npm ls` to verify if the duplicated Lit package was successfully de-duped.
+
+4. If there is still duplication, you may need to delete your package lock and `node_modules`. Then install the version of `lit` you want explicitly, followed by your dependencies.
```
augustjk, line 72: Before this nuclear option, it's possible to nudge npm to hoist particular versions by making core packages direct dependents of your application project with `npm i @lit/reactive-element@latest lit-element@latest lit-html@latest`

packages/lit-dev-content/site/docs/v3/tools/requirements.md
```diff
@@ -49,3 +49,38 @@ All modern browsers update automatically and users are highly likely to have a r
...
+
+It is possible to follow these steps, and not be able to de-duplicate Lit. For example, a library you depend on may bundle a specific version of Lit in a way that cannot be de-duplicated. In these cases the warning can be ignored.
```
augustjk, line 74: I think this should be mentioned up top first, that multiple versions in themselves do work, and warning can be ignored. This kind of sounds like well if you can't resolve this by deduping, just give up and ignore the warning.

augustjk (commented):
We've already discuss this might be better located elsewhere. Perhaps we do this in parallel with https://github.com/lit/lit.dev/issues/1262

packages/lit-dev-content/site/docs/v3/tools/requirements.md
AndrewJakubowicz: Agree that it feels hacky, however I think it's necessary that we document a way to resolve this warning, especially if someone has investigated the warning and cannot de-duplicate due to application specific reasons (maybe a dependency is bundling Lit).

The reasons I think we need to keep this until we have a better alternative is:

 1. If you cannot resolve warnings, they waste developer time as each developer tries to resolve them. Mentioned [here](https://github.com/lit/lit/pull/4319#issuecomment-1782389845)
 2. If warnings build up because you cannot resolve them, they start to just become background noise that'll disappear in production.
 
This solution is fragile, but it is compatible with Lit 2 and Lit 3, and a very small amount of code.

packages/lit-dev-content/site/docs/v3/tools/requirements.md
AndrewJakubowicz: That is a great point! Moving this up top.

packages/lit-dev-content/site/docs/v3/tools/requirements.md
AndrewJakubowicz: Added this. Thank you!

packages/lit-dev-content/site/docs/v3/tools/development.md
```diff
@@ -78,6 +78,45 @@ class MyElement extends LitElement {
 
 It's best for code size if the code to control warnings is eliminated in your own production builds.
 
+#### Multiple versions of Lit warning {#multiple-lit-versions}
+
+If Lit is being used as an internal dependency of elements, elements can use different versions of Lit and are completely interoperable.
```
augustjk, line 83: We should be clear what triggers the warning at the top of this section, before going into why it's okay with Lit's behavior.

```md
A dev mode only warning is triggered when multiple versions, or even multiple copies of the same version, of any of the Lit core packages – `lit-html`, `lit-element`, `@lit/reactive-element` – are detected.
```

AndrewJakubowicz: Wonderful! Added verbatim.

packages/lit-dev-content/site/docs/v3/tools/requirements.md
augustjk: Now that I think of this, I wish we hadn't done the deduping of warnings by the message string. We'd be opting out of the dedupe across versions if we want to change the message.

Still, writing out our "internal" mechanism for deduping warnings just feels wrong. Feels like documenting a way to hack our own package. I'd almost prefer it if users patch `console.warn` to skip any message that includes `'https://lit.dev/msg/multiple-versions'`.

I'm not happy, but I don't wish to block the rest of this excellent addition on account of this.

augustjk (approved):

packages/lit-dev-content/site/docs/v3/tools/requirements.md
AndrewJakubowicz: I agree that we probably should not have de-duped based on the string contents, but instead some stable error code that we've invented. E.g. Rust error code index: https://doc.rust-lang.org/error_codes/error-index.html

We should probably think about adopting some unique error string, or maybe we could use the first argument that becomes the link to lit.dev to track duplication.

> I'm not happy, but I don't wish to block the rest of this excellent addition on account of this.

Your message swayed me with the note that we can easily hack around this without relying on Lit globals/internals. Since the `console.warn` hack can easily behave in the same way I agree with you to not document the specific hack.


I removed it. WDYT about adding a note that you can silence it by patching `console.warn`? Or should we leave that to users?

AndrewJakubowicz: Discussed offline–documenting hacks in official docs is probably not a great idea.

augustjk: Leaning towards not mentioning and leaving it up to users. We can revisit if users really want a way to suppress it. And perhaps we can provide a longer term solution similar the reactive-element's disable warning feature though that'll only work for future versions from that point on. we have to start somewhere!

augustjk (approved):
👯‍♂️

Issue #1269, opened 2023-11-20
[infra] Update Lit.dev to Lit 3


AndrewJakubowicz:
### Context

Lit.dev should be running the most recent version of Lit. Lit 3.0
It currently is still on Lit 2.

Issue can be closed when Lit.dev is on v3.

PR #1270, opened 2023-11-20
[infra] update Lit and the Lit eleventy plugin

AndrewJakubowicz:
Part of issue: https://github.com/lit/lit.dev/issues/1269

### Context

Instead of updating Lit to 3 at once, this change updates us to the latest version of Lit 2. The change to Lit 3 will follow.

I also updated the `@lit-labs/eleventy-plugin-lit` to pull in the module resolution change.

### Test plan

Bugs in the Lit upgrade should be caught in the playground or screenshot tests. I tested preview link manually.
Bugs in `eleventy-plugin-lit` would result in catastrophic build errors. Tested locally & also covered by checking preview link.


AndrewJakubowicz:
Looks like this naive upgrade resulted in duplicated Lit which is creating property minifaction errors in the eleventy lit plugin.

augustjk (approved):

PR #1271, opened 2023-11-20
[docs] mention that server should import custom element definitions to server render them

AndrewJakubowicz:
Context: https://github.com/lit/lit/discussions/4406

While investigating the user issue, I noticed that we don't seem to explicitly mention or show that LitElements should be defined on the server.

`@lit-labs/ssr` patches `customElements.define` to be able to reference the class for a given tag. If the tag doesn't have a `LitElement` defined for it, then the element will not have it's declarative shadow DOM rendered on the server.

### Fix

Add a small change to the documentation which explicitly shows an import to the custom element on the server.


augustjk (approved):
Thank you!

Issue #1273, opened 2023-11-21
Property 'blockId' has no initializer and is not definitely assigned in the constructor.


andyjessop:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

I'm defining a property like this:

```ts
import { css, html, LitElement } from 'lit';
import { customElement, property } from 'lit/decorators.js';

@customElement('rune-block-actions')
export class RuneBlockActions extends LitElement {
  @property()
  blockId: string;
```

I get this TS error in the editor:
```
Property 'blockId' has no initializer and is not definitely assigned in the constructor.
```

### Reproduction

The code is unfortunately closed-source, and I cannot reproduce on the playground.

### Workaround

I have not found a workaround.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

3.0.2

### Browser/OS/Node environment

Browser: Chrome 119.0.6045.105
OS: Mac 14
Node: 18.6.1
npm: 9.5.1

PR #1274, opened 2023-11-21
[docs] Fix 'property' has no initializer issue in TypeScript property samples

AndrewJakubowicz:
Fixes: https://github.com/lit/lit.dev/issues/1273

### Context

When looking at our documentation, the very first TypeScript example shows: https://lit.dev/docs/components/properties/

```ts
  @property()
  name: string;
}
```

This results in a TypeScript error: `Property 'name' has no initializer and is not definitely assigned in the constructor.`.

This can be globally resolved by relaxing the TypeScript compiler via `"strictPropertyInitialization": false`, but we should prefer documentation that is as explicit/strict as possible.

We have [an example](https://lit.dev/playground/#sample=examples/properties-custom-converter) in the playground where it makes sense to not initialize a property. So I've added that syntax to our `properties.md` docs.

Fix:
```ts
  @property()
  name?: string;
}
```

This clearly states that the property should be a string, but it can also be undefined if no-one initialized it.

### Risk & Tests

This is a low risk content only change – tested by checking out preview URL.

### DIFF
--- a/packages/lit-dev-content/site/docs/v2/components/properties.md
+++ b/packages/lit-dev-content/site/docs/v2/components/properties.md
@@ -16,7 +16,7 @@ Lit components receive input and store their state as JavaScript class fields or
 ```ts
 class MyElement extends LitElement {
   @property()
-  name: string;
+  name?: string;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


augustjk (approved):
It's okay for now.. optional won't be valid once we start using `accessor` but we can cross that bridge later. Alternative would be providing defaults but I can't think of good ones at the moment.

PR #1277, opened 2023-11-29
[docs] Fix context provide decorator api doc

augustjk:


packages/lit-dev-content/site/docs/v2/data/context.md
```diff
@@ -369,7 +369,7 @@ The `ValueType` type parameter is the type of value that can be provided by this
 
 ### `@provide()`
 
-A property decorator that adds a ContextConsumer controller to the component which will try and retrieve a value for the property via the Context API.
+A property decorator that adds a ContextProvider controller to the component making it respond to any `context-request` events from its children consumer.
```
AndrewJakubowicz, line 372: Nit: thoughts about `children consumer` -> `child consumers`?

AndrewJakubowicz (approved):
My comment is non blocking because this is a pretty glaring error you have fixed. Thank you!

packages/lit-dev-content/site/docs/v2/data/context.md
augustjk: I took it straight from the JSDoc here
https://github.com/lit/lit/blob/f60a3a2c994f41fc3df1bd8a76451ea185b66e11/packages/context/src/lib/decorators/provide.ts#L19-L20

While I don't oppose to the new wording itself, I think it's minor enough that I'd rather keep it in sync with what's on the source code. Thanks for the suggestion though!

AndrewJakubowicz: Great call!

PR #1279, opened 2023-12-11
fix import of createContext in context.md

38elements:
This is the section of `createContext`.


augustjk (approved):
whoa, weird mistake there must've been some bad copy paste that got missed. thank you very much for the fix!

PR #1280, opened 2023-12-14
Improve localize documentation

julienvier:
Add `localeCodesModule` in lit-localize.json


augustjk (approved):
Thank you for the PR.
Technically it's an optional config but all of our examples do use it as good practice for keeping the locales in code in sync with those listed in the config so this seems fine to add.

PR #1283, opened 2023-12-16
Write an article on Reactive Controllers and Redux

e111077:


packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+- [`requestUpdate()`](/docs/api/ReactiveElement/#ReactiveElement.requestUpdate)
+- [`updateComplete: Promise<boolean>`](/docs/api/ReactiveElement/#ReactiveElement.updateComplete)
+
+The `addController()` method takes in the controller that you want to hook into the host’s lifecycle.
+
+{% aside "info" "no-header" %}
+
+If the host is already attached to the DOM or rendered onto the page, then `hostConnected()` will be called after attaching the Reactive Controller via `addController()`.
+
+{% endaside %}
```
augustjk, line 104: This is specific to `@lit/reactive-element`'s implementation of `addController()`. Perhaps it could be rephrased as a recommended way to implement `addController()` or removed?

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+  private host: ReactiveControllerHost;
+
+  constructor(host: ReactiveControllerHost, options: MyOptions) {
+    this.host = host;
+    this.host.addController(this);
+  }
+}
+```
+
+The `removeController()` method is used less frequently than the other callbacks. It is useful when you do not want the controller to update with the host, such as: the host updates too often, the `hostUpdate[d]()` methods have slow or expensive logic, or you do not need the controller to run its updates while the component has been removed from the document.
```
augustjk, line 119:
```suggestion
The `removeController()` method is used less frequently than the other callbacks. It is useful when you do not want the controller to update with the host, such as: the host updates too often, the `hostUpdated()` methods have slow or expensive logic, or you do not need the controller to run its updates while the component has been removed from the document.
```

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+
+The read-only `updateComplete` property is often used in conjunction with `requestUpdate()` method. A `ReactiveControllerHost`’s update lifecycle is assumed to be asynchronous, so the `updateComplete` property is a promise that resolves when the host’s update lifecycle has completed. This is useful for controllers that need to update the DOM and then read from it. For example, imagine a controller that resizes a DOM element and needs to then read its new dimensions. This controller would update a property, call `requestUpdate()`, await `host.updateComplete`, and then read the DOM.
+
+## Redux Patterns
+
+Redux has a bit of verbosity associated with it in order to enforce the Redux state management patterns. In this article we will be making a simple component that renders circles and squares, renders how many of them exist, and can increment or decrement the amount of circles and squares. This component will have its state managed by Redux.
```
augustjk, line 127: Commenting on this whole section: The pattern shown here is an older redux pattern that is more verbose than needed. I think the goal of this article shouldn't be to teach Redux so we just use the most concise way to create actions and reducers, which is `createSlice` from `@redux/toolkit`.
https://redux.js.org/tutorials/essentials/part-2-app-structure#the-counter-example-app

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+<img
+    src="/images/articles/redux-reactive-controllers/application-sans-labels.webp"
+    loading="lazy"
+    width="550px"
+    title="Sketch of the redux application"
+    alt="A vector box diagram of the entire redux application. There are 3 sections. The first is the shape-dials component – it has 3 rows. The first row is a minus button, the label 'squares', followed by a plus button. The second row is a minus button, the label 'circles', followed by a plus button. The final row is a long 'reset' button. The next section to the right of the shape-dials section the '<shape-count>' component. It is a black box with 3 rows of text. The first row is '2 circles'. The second row is '2 squares', and the third row is '4 total'. Below the shape-dials and shape-count components is a long rectangle which is the '<shape-list>' section. This is a long, horizonal, black rectangle and inside of it is a red square followed by a blue circle, blue circle, and another red square.">
+
+### Initial State
+
+We need to define an initial state, so in a <code>store.<ts-js></ts-js></code> file we will give the initial state the following shape:
```
augustjk, line 138: The code sample below isn't switchable so I don't see a reason to switch on the file extension either. Since these are not docs, I think it's fine to stick to TS all the way, unless you really want to show the JS code sample too.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+  host: ReactiveControllerHost;
+
+  constructor(host: ReactiveControllerHost) {
+    this.host = host;
+    host.addController(this);
+  }
+}
+```
+
+Next, let's accept the following options: the Redux store in which we would like to subscribe to, as well as the Selector we would like to use to select what data we would like to render from the store:
```
augustjk, line 354: Data can be selected for other reasons than rendering
```suggestion
Next, let's accept the following options: the Redux store in which we would like to subscribe to, as well as the Selector we would like to use to select what data we would like to pick out from the store:
```

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+        this.host.requestUpdate();
+      }
+    });
+  }
+}
+```
+
+By first comparing the previous state to the current state, we can avoid re-rendering components that don't need to be re-rendered which can improve performance. Nice!
+
+In conclusion, we need to improve our component so that it does not re-render when the component is disconnected from the page and the store’s state changes. Redux’s `Store.subscribe()` method returns an `unsubscribe()` function. Let’s keep track of this and unsubscribe from the store’s changes when the component disconnects.
```
augustjk, line 472: Logically it feels like unsubscribe should be implemented before the equality check as a subscribe/unsubscribe and connected/disconnected generally should be considered together. Equality check is an optimization that can be done afterwards.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+export type Selector<State, Result> = (state: State) => Result;
+
+export class SelectorController<
+    State,
+    Action extends RAction,
+    Result = unknown
+> implements ReactiveController {
+  host: ReactiveControllerHost;
+  store: Store<State, Action>;
+  selector: Selector<State, Result>;
```
augustjk, line 369: Nit: It probably makes sense to mark all of these as private and do either `#` or `_` prefix.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+
+```ts
+...
+export class SelectorController<
+    State,
+    Action extends RAction,
+    Result = unknown
+> implements ReactiveController {
+  ...
+  selected: Result;
```
augustjk, line 395: Nit: I think `value` makes more sense for this field which also should be marked public.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+          <td>${shapeList.length}</td>
+          <td>total</td>
+        </tr>
+      </table>
+    `;
+  }
+}
+```
+
+Pressing the buttons in `shape-dials` should now update the count!
```
augustjk, line 588: It's a bit hard to follow this part since we haven't actually seen `shape-dials`.. maybe that's okay for now.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+
+To accomplish this, we initialized the `SelectorController` with the shared Redux store and rendered the entire state.
+
+### shape-list
+
+`shape-list` should be a component that only subscribes and reads the `state.shapeList` state from the store. Let's create a custom element with boilerplate, and render an array of `<div>`s with classes set to the shape name. Our pre-provided CSS styles will render the squares and circles based on the class name.
```
augustjk, line 594: Nit: adding "s" after code text just looks really bad with our style, especially after a closing `>`.
```suggestion
`shape-list` should be a component that only subscribes and reads the `state.shapeList` state from the store. Let's create a custom element with boilerplate, and render an array of `<div>` elements with classes set to the shape name. Our pre-provided CSS styles will render the squares and circles based on the class name.
```

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,731 @@
...
+
+## What is a Reactive Controller?
+
+Reactive Controllers are a programming pattern that makes it easy to share logic between components by hooking into a component’s reactive update lifecycle. They achieve this by expecting an object that exposes an interface rather than having to create a new component or subclassing like you would with a mixin.
+
+One advantage of the Reactive Controller pattern is that it creates a **with** relationship rather than an **is** relationship. For example, a component that uses a Reactive Controller that incorporates Redux logic is a **component with Redux selector abilities**, whereas a mixin that does the same would mean that the **component is a Redux selector component**. This type of composability results in code that is more portable, self-contained, and easier to refactor. This is because components that inherit via subclassing are more closely coupled with the logic they inherit.
```
augustjk, line 29: Is it worth mentioning for readers familiar with React that Reactive Controllers are like hooks whereas mixins are like higher order components? It also allows parallels below how the `SelectorController` is much like the `useSelector` hook from `react-redux`.

augustjk (commented):
Is there a playground link with the finished example or is that not included on purpose?

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
augustjk: Also I'm debating whether configurable equality check is even needed. With redux recommending immutable patterns, triple equal should be good enough and it removes some complexity from the basic demonstration here.

e111077: `===` would cause the array to re-render every state change in the shapes-list. a shallow equals would be useful here. Actually they offer a [shallowEqual](https://github.com/reduxjs/react-redux/blob/master/src/utils/shallowEqual.ts) in their react redux library

augustjk: For redux state, the array should only be a new reference if any items inside it has changed so `===` works. The `shallowEqual` in `react-redux` is used for comparing state/props across React component re-renders which would always be a new object.

I don't disagree that if we were to actually vend this as a package, we should make the equality check configurable. I wanted to raise whether it's worth the complication in a reactive controller demonstration.

Ultimately, I think it's fine to keep as an example of what things could be options for a controller. I do still think the order should be flipped with the disconnected.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
augustjk: Found the old playground example this article is based from. Here's a stab and refactoring to use `createSlice`. Auto generated action creators remove the need for the verbose enumerated action types. https://lit.dev/playground/#gist=4c07268fb5ec5fddf0280e7882d39bc0

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
e111077: Does the image in the previous section nor the image in the Actions section not help?

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
e111077: removed

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
augustjk: I think my uneasiness with this sentence is more that this describes what clicking the buttons in `shape-dials` does which isn't the what the previous code is about. It's about subscribing to the store and updating UI on state change.

Maybe it can be something like
```suggestion
The counts are now pulled from the state in Redux and any updates to the state will update the component!
```

e111077: ah this makes sense

e111077:
Thanks for leading this blind man through Redux! All suggestions should be addressed. Though I think it def needs another pass and a bit more discussion about structure. Personally I think it's okay to be a bit Lit specific since it's hosted on Lit.dev, but i agree it's a bit lazy to talk about the lit implementation for a specific part and the generalized use case in the rest of it

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+
+[Redux](https://redux.js.org/) is a mature library that introduces patterns to manage state across a JavaScript application. The Lit team does not have a particular endorsement for a single state management library, but we will be using Redux as an example for creating a Reactive Controller, because the patterns used in integrating Redux into Lit with a Reactive Controller may be used to integrate for other popular libraries.
+
+There are three general concepts that Redux describes:
+
+1. Stores
+2. Actions
+3. Reducers
+
+Stores contain the current application state. Actions describe what kind of change to make to the state, and reducers take actions and apply them to the current state to return a new state. Here is a diagram derived from the [official Redux documentation](redux.js.org/tutorials/essentials/part-1-overview-concepts#redux-application-data-flow ?) that depicts the interaction pattern between these concepts:
```
AndrewJakubowicz, line 50: This markdown link doesn't work, and seems to render in full.

See preview url: https://pr1283-b46091e---lit-dev-5ftespv5na-uc.a.run.app/articles/redux-reactive-controllers/#:~:text=%5Bofficial%20Redux%20documentation

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+  ...
+  value: Result;
+
+  constructor(
+    host: ReactiveControllerHost,
+    store: Store<State, Action>,
+    selector: Selector<State, Result>,
+  ) {
+    ...
+    this.value = selector(store.getState());
```
AndrewJakubowicz, line 366: These incremental additions are a bit hard to follow without a playground open to provide a full picture. No action required here except I think this would be solved with a playground.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+        this._host.requestUpdate();
+      }
+    });
+  }
+}
+```
+
+Great! Now the controller will update its value and tell the host element to update whenever the state changes. Additionally, by first comparing the previous state to the current state, we can avoid re-rendering components that don't need to be re-rendered which can improve performance. Nice!
+
+In conclusion, we need to improve our component so that it does not re-render when the component is disconnected from the page and the store’s state changes. Redux’s `Store.subscribe()` method returns an `unsubscribe()` function. Let’s keep track of this and unsubscribe from the store’s changes when the component disconnects.
```
AndrewJakubowicz, line 400: Nit: Starting with `In conclusion,` makes me feel like we're about to wrap up the article.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+    });
+  }
+
+  hostDisconnected() {
+    this._unsubscribe();
+  }
+}
+```
+
+We now have a Redux `SelectorController` Reactive Controller that listens to a Redux store for state changes, selects a value from the state, and updates the host element whenever that state value changes!
```
AndrewJakubowicz, line 425: Nice section! Imho this section is the juicy part of the article that shows off the actual integration. We can maybe celebrate this more.
Could use this location to hammer home the benefits of having a `SelectorController`.

"The SelectorController can be reused across components that need to access some state in the redux store, and all the logic for automatically and efficiently re-rendering those components is encapsulated in this controller."

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+    alt="A vector box diagram of the entire redux application. There are 3 sections. The first is the shape-dials component – it has 3 rows. The first row is a minus button, the label 'squares', followed by a plus button. The second row is a minus button, the label 'circles', followed by a plus button. The final row is a long 'reset' button. This component is surrounded by a red dashed-square labeled '<shape-dials>'. The next section to the right of the shape-dials section is labeled by a red label that says '<shape-count>'. It is a black box with 3 rows of text. The first row is '2 circles'. The second row is '2 squares', and the third row is '4 total'. Below the shape-dials and shape-count components is a long rectangle labeled '<shape-list>'. This is a long, horizonal, black rectangle and inside of it is a red square followed by a blue circle, blue circle, and another red square.">
+
+### shape-count
+
+`shape-count` should be a component that only subscribes and reads from the store. Let us create a custom element and render the table:
+
+```ts
+import { LitElement, html} from 'lit';
+import { customElement} from 'lit/decorators.js';
```
AndrewJakubowicz, line 451: Nit: The spaces are imbalanced between the `{` and `}`.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+          <td>${0}</td>
+          <td>total</td>
+        </tr>
+      </table>
+    `;
+  }
+}
+```
+
+Next we want to render actual counts rather than just `0`. In this case we will need all values from the state:
```
AndrewJakubowicz, line 476: Nit: `In this case we will need all values from the state:`.
Maybe:

`In this case we will need to pull those values from the store`.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+
+- `state.circles`
+- `state.squares`
+- `state.shapeList`
+
+To achieve this, we will need a broad Redux selector that selects the entire state:
```
AndrewJakubowicz, line 482: Maybe, "that selects the entire store"?

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+}
+```
+
+The counts are now pulled from the state in Redux and any updates to the state will update the component!
+
+To accomplish this, we initialized the `SelectorController` with the shared Redux store and rendered the entire state.
+
+### shape-list
+
+`shape-list` should be a component that only subscribes and reads the `state.shapeList` state from the store. Let's create a custom element with boilerplate, and render an array of `<div>` elements with classes set to the shape name. Our pre-provided CSS styles will render the squares and circles based on the class name.
```
AndrewJakubowicz, line 522: Nit: can `pre-provided` be removed?

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+    const shapeList = [];
+
+    return shapeList.map(
+        shape => html`<div class="${shape}"></div>`
+    );
+  }
+}
+```
+
+Next, let’s initialize our `SelectorController` to hook into Redux and render the `shapeList`:
```
AndrewJakubowicz, line 540: Worth calling out that this time the selector is focussing a specific value (`shapeList`), instead of selecting the whole store with an identity function?

The difference here is that this component will only re-render if specifically `shapeList` changes.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+import { SelectorController } from './selector-controller.js';
+import { store } from './store.js';
+
+@customElement('shape-list')
+export class ShapeList extends LitElement {
+  private sc = new SelectorController(this, store, (state) => state.shapeList);
+
+  render() {
+    const shapeList = this.sc.value;
+    ...
```
AndrewJakubowicz, line 553: I'm not sure it's worth omitting the returned line here, since it shows that it is fine to re-render lists and it's efficient.

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+  }
+}
+```
+
+The `shape-list` component should now be responsive to changes in the Redux store!
```
AndrewJakubowicz, line 558: `responsive` makes me think window size. Dare I say, I think I prefer the word "reactive" instead. (But this is a nit, not required).

Another small nit: Can the word "should now be" be replaced with "is now"? I read "should" as "probably", and sure hope the shape-list is reactive to the redux store now!

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+
+We were able to accomplish this by initializing the `SelectorController` with the shared Redux store. We then selected only the `shapeList` from the state and updated the host element only when the arrays were truly not equal.
```
AndrewJakubowicz, line 560: Nit: Did this opt-into past tense?

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+
+### Preventing invalid inputs in shape-dials
+
+To prevent invalid inputs, we will use our `SelectorController` to disable the buttons in the `shape-dials` component. For example, we want to disable the decrement buttons when the respective shape count is 0, and we want to disable the reset button when the length of the `shapeList` is 0.
+
+We will be using the entire state object again, so the selector will be broad. Let’s add the SelectorController to our `shape-dials` component.
```
AndrewJakubowicz, line 566: Nit: Can maybe remove "so the selector will be broad".

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+
+### Your Own Bespoke Behavior
+
+Reactive Controllers are generally a good way to share logic across components and we cannot cover every possible use case here. For example [Material Design’s Material Components](https://material-web.dev) have written their own bespoke controllers such as the [SurfacePositionController](https://github.com/material-components/material-web/blob/v1.1.1/menu/internal/controllers/surfacePositionController.ts) which can position a popup surface next to an anchor or [TypeaheadController](https://github.com/material-components/material-web/blob/v1.1.1/menu/internal/controllers/typeaheadController.ts) which can automatically select an item from a list just by typing the first few letters of the item like an autocomplete.
+
+Reactive Controllers are flexible, focused, and a great way to integrate libraries into your Lit project or any framework of your choice.
```
AndrewJakubowicz, line 633: This whole ending section is awesome. It's a great upbeat section full of awesome examples and ideas.

Could also add a call to action to inspire readers to go and implement some controllers where they see the requirement to share behavior.

AndrewJakubowicz (commented):
This post is awesome! I had a lot of fun going through it. Most of my comments are minor suggestions that could be ignored. I left a comment anywhere that I needed to leave the article to pursue other knowledge.

What is the assumed knowledge of the reader?
 - I ask this question because it can drive some of the comment resolutions that I left. E.g., if this is focussed on quite technical readers who can follow quite sophisticated TypeScript, then we can leave in the more complex types.

Similarly there is quite a nuanced discussion of ReactiveControllers that may be fine to leave here, but we should update the documentation pages with your excellent information.

The only blocking change I'd like to see is a working playground example at the top of the article (or bottom?). Because it's a bit hard to follow the code samples out of context.

Elliott, this was an amazing read. Thank you so much for writing this!

packages/lit-dev-content/site/articles/article/redux-reactive-controllers.md
```diff
@@ -0,0 +1,633 @@
...
+    title="Sketch of the redux application"
+    alt="A vector box diagram of the entire redux application. There are 3 sections. The first is the shape-dials component – it has 3 rows. The first row is a minus button, the label 'squares', followed by a plus button. The second row is a minus button, the label 'circles', followed by a plus button. The final row is a long 'reset' button. The next section to the right of the shape-dials section the '<shape-count>' component. It is a black box with 3 rows of text. The first row is '2 circles'. The second row is '2 squares', and the third row is '4 total'. Below the shape-dials and shape-count components is a long rectangle which is the '<shape-list>' section. This is a long, horizonal, black rectangle and inside of it is a red square followed by a blue circle, blue circle, and another red square.">
+
+### Initial State
+
+We need to define an initial state, so in a <code>store.<ts-js></ts-js></code> file we will give the initial state the following shape:
+
+```ts
+export type Shape = 'square' | 'circle';
+export type State = Record<`${Shape}s`, number> & {shapeList: Shape[]};
```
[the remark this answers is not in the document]
augustjk: I fully take blame for introducing this and agree it should be simplified or removed.

vikjung:
Two things will help make this more useful -  

1. provide code in javascript
2. add how it could replace e.g. used with the pwa starter kit example which used lazyReducerEnhancer and thunk

PR #1288, opened 2024-01-10
docs: add labs lowdown videos and small material animation to thumbnails

e111077:
uses standard material cubic bezier animation


augustjk (commented):
The thumbnail animation is cute and I love it!

I don't know about the thumbnails for the lowdown videos though. The off center gradient with no discernible border is messing with my eyes. Ahh I realized those are the actual thumbnails from the youtube videos. Would it be bad if we made different thumbnails for these?

It also would be lovely if we could add video length metadata for these cards like we have for tutorials.

packages/lit-dev-content/site/_data/videos.js
```diff
@@ -80,6 +80,54 @@ const loadVideoData = () =>
...
     },
+    {
+      title: 'Lit Labs: Task (graduated)',
+      summary: `This video covers @lit-labs/task, a package under development in Lit Labs. Watch for an overview, demos, package status and more! Originally streamed as part of the Lit Labs Lowdown event on October 24, 2022.`,
+      youtubeId: 'niWKuGhyE0M',
+      date: 'Jan 19 2023',
+    },
+    {
+      title: 'Lit Labs: Context (graduated)',
+      summary: `In this video, we cover @lit-labs/context, a package under development in Lit Labs. Watch for an overview, an explanation of the underlying Web Components Context Protocol, use cases, demos, package status and more! Originally streamed as part of the Lit Labs Lowdown event on October 24, 2022.`,
```
AndrewJakubowicz, line 91: What do you think about shortening these summaries. They're a bit long on the preview.

AndrewJakubowicz (commented):
Thank you! This is awesome.

The new thumbnails blend in with the page background and the gradient is mis-aligning. I wonder if it's worth somehow separating the thumbnails from the background. We can either update the thumbnails, or add some contrast, maybe with shadow, or border, or applying a filter that changes their color.

e111077:
I've added backgrounds to the sections. do you think it looks too "heavy"? I can also add borders to the images, but if I do that without the bg, it seems odd

e111077:
only problem with border is that it seems there's a small issue with overflow + our aspect ratio enforcement

e111077:
added durations:

image

e111077:
also my position on this is that it's a hassle manually generating the thumbnails. I'd rather have a fix that doesn't require custom thumbnail generation. Currently experimenting with BG and border. Long term solution would be a server-side auto-generation of thumbnail 1x + 2x sizes

augustjk:
Yeah. I agree we shouldn't manually generate thumbnails for videos if we can avoid it. It's just unfortunate for those series of videos. They also look bad on youtube.com in light mode.

The background makes those look tolerable but then makes the lit 3 launch video thumbnail and the light blue background thumbnails weird. :( It's not perfect but I'm no better at design so I'm okay with how it is.

Durations are awesome, thanks!

augustjk (approved):

AndrewJakubowicz (approved):
Nice!

I similarly wonder if there is a better option than the gray background, but I am also not a designer so don't want to block on that. Everything else is awesome!

Also agree that thumbnails should be automatically generated. The script would:

 1. Fetch from youtube using video id.
 2. Process the images into the right sizes and locations.

e111077:
@AndrewJakubowicz I think there is, but we currently don't have a precedent for a lighter bg color on the site.

Dark mode is such a TODO, but it requires tokenizing all of our colors

e111077:
Also @augustjk FWIW, Material Design doesn't really have anything to create contrast between a non-fullbleed image and a card

https://m3.material.io/components/cards/guidelines#b1f0df76-3c86-4b32-982a-43102374e091
https://m3.material.io/components/cards/guidelines#5608bee0-e976-43c3-843d-382ace2dd980

PR #1291, opened 2024-01-12
[docs/task] Fix code sample for manual run

augustjk:
The args need to be provided to `.run()` call in an array.


AndrewJakubowicz (approved):
Thank you! Great catch!

Issue #1285, opened 2023-12-20
Inconsistency between `All` package on CDN and NPM modules


digitalsadhu:
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

We've noticed a bit of an inconsistency between the Lit all CDN package and using the packages via npm. My expectation is that the CDN URL would essentially be a drop in replacement, however there's an edge case due to aliasing of `html` from the `static-html` package.

This:
```js
import { unsafeStatic, html } from "lit/static-html.js";
```
and this:
```js
import { unsafeStatic, html } from "https://cdn.jsdelivr.net/gh/lit/dist@3/all/lit-all.min.js";
```
are not the same since.

In the former you get `html` from the lit static package where as in the latter you get `html` from the core package. 

This has proved quite a gotcha for us and as far as I'm aware isn't well documented. 

If you'd like me to PR a documentation change, just point me in the right direction.

### Reproduction

No reproduction should be necessary

### Workaround

No workaround found.
This should probably be documented.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

All versions

### Browser/OS/Node environment

Environment not relevant

augustjk:
Hope that worked! I've transferred the issue to lit.dev repo as we should document the aliases where we mention the bundle.

PR #1295, opened 2024-01-13
[docs] Add note about static tag aliasing for lit-all bundle

augustjk:
Fixes https://github.com/lit/lit.dev/issues/1285


AndrewJakubowicz (approved):
Nice!

PR #1297, opened 2024-01-14
[SSR] (docs) show Lit `html` dependency in SSR docs examples

thescientist13:
Not sure if intentional, but noticed that the SSR examples weren't showing where the `html` tagged template comes from, which presumably was meant to be from the lit package, e.g.
```js
import {html} from 'lit';
```

> _As a side-effect, there was already an `const html = ...` variable being used in the examples, so just to avoid confusion, renamed it to `const contents = ...`, but I am happy to change it whatever you would prefer._


augustjk (approved):
Excellent additions! Thank you very much!
Looking good in the preview: https://pr1297-0fcbd41---lit-dev-5ftespv5na-uc.a.run.app/docs/ssr/server-usage/

PR #1301, opened 2024-01-31
[docs/localization] Update info on messages that map to the same id

augustjk:
Issue found while looking into https://github.com/lit/lit/issues/4502


AndrewJakubowicz (approved):
Well written!

PR #1303, opened 2024-02-05
fixed typo in upgrade.md

davie-robertson:
"transform Lit 3 code to work with Webpack 5" should be Webpack 4 as Lit 3 already works with Webpack 5


augustjk (approved):
Good catch, thank you!

PR #1305, opened 2024-02-16
fix(docs): remove duplicate line in code sample

maxpatiiuk:
This PR fixes a small issue in the docs:
- a code sample line duplicated twice

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/data/context.md
+++ b/packages/lit-dev-content/site/docs/v3/data/context.md
@@ -28,7 +28,6 @@ Context definition (`logger-context.ts`):
 ```ts
 import {createContext} from '@lit/context';
 import type {Logger} from 'my-logging-library';
-export type {Logger} from 'my-logging-library';
 export const loggerContext = createContext<Logger>('logger');
 ```
 


packages/lit-dev-content/site/docs/v3/data/context.md
```diff
@@ -28,7 +28,6 @@ Context definition (`logger-context.ts`):
 ```ts
 import {createContext} from '@lit/context';
 import type {Logger} from 'my-logging-library';
-export type {Logger} from 'my-logging-library';
```
augustjk, line 31: This line is actually not a duplicate and is meant to be a re-export of the type from this module so that consumer won't need to import it from `'my-logging-library'` package, and rather import it from this context module instead as shown in the consumer code example below.

augustjk (commented):
Thank you for the PR but the line is actually not a duplicate. I also remember doing a double take when I first saw it. I don't know of a way in TS to both import and re-export with a single line while still making use of the imported value/type in the current module.

packages/lit-dev-content/site/docs/v3/data/context.md
maxpatiiuk: ah, right!
sorry, not sure how I overlooked that

PR #1306, opened 2024-02-16
docs(controllers): remove needless non-null assertion operator

maxpatiiuk:
This PR fixes a small issue in the docs:
- a needless non-nullish assertion (since the this.task property is assigned in the constructor, `!` is not necessary)

Btw, I really like Lit controllers! I am primarily working with Stencil, and the need for code reuse lead me to create a pattern I called "Stencil Controllers" - that was before I knew that Lit had a very similar feature, with a very similar name - what a coincidence!


augustjk (commented):
Thank you for the PR!
Could you also make the same change to `packages/lit-dev-content/samples/v3-docs/controllers/names/names-controller.ts`?
We have copies of docs for v2 and v3 and this change can apply to both versions.

augustjk (approved):
Thanks!

PR #1308, opened 2024-02-28
feat: implement dark mode

e111077:
Large CL that implements dark mode using mostly material 3's token system generated in monochrome with some site-specific additions. Some color refactors were necessary since they were either random magic colors or did not fit into the token system.

This Also adds reddot to the list of users.

Also there are some... color choices I had to make. The BG is very dark so that the tokens would match light and dark more more correctly. If I tried one-offing the tokens for light mode vs dark mode, there were lots of inconsistencies that looked bad in dark but good in light. So now we have quite contrasty blacks and saturated primary, lit blues around a lot of the site unfortunately


sorvell:
Amazing, love it!

* the contrast on some of the header text seems too subtle. I wonder if all the text should just be light, rather than have a blue-ish tint, e.g. the main header/footer text on the front page.
* the button background seems a little too purple-ish.

justinfagnani:
The diagram on https://pr1308-9c4b190---lit-dev-5ftespv5na-uc.a.run.app/docs/composition/component-composition/#passing-data-up-and-down-the-tree is difficult to see because it has a transparent background and black lines.

e111077:
Tests are gonna fail because of screenshots, but i think this might be good for another pass.

I added the [`/design/`](https://pr1308-2a73c2a---lit-dev-5ftespv5na-uc.a.run.app/design/) page for your convenience as well.

@sorvell Thanks for looking into it, here's my responses.

> - Lit logo is invisible until hover on front page.

This is the default behavior on prod

> - the contrast on some of the header text seems too subtle. I wonder if all the text should just be light, rather than have a blue-ish tint, e.g. the main header/footer text on the front page.

Contrast should all be AA+ rated on this. I removed the dimmest one which was AA and turned h4's into the same color as h3 which is AAA rated.

> - the button background seems a little too purple-ish.

Changed color PTAL

> - learn page pill buttons don't seem to match the other colors.

They should now with the color change. PTAL

> - what about adding a (can be short) blog post in conjunction with this?

I'll write one up after this lands depending on how our comms should be considered going forward

e111077:
ping for another review. All comments should have been addressed in some respect

packages/lit-dev-content/site/home/5-who-is-using.html
```diff
@@ -21,6 +21,18 @@ <h1>
...
+          svg-role="img"
+          label="Reddit"
+          href="{{ site.baseurl }}/images/logos/reddit.svg#full"
+          loading="visible">
+        </lazy-svg>
+        <div>
+          <span>Frontend UI</span>
+        </div>
+      </a>
+
```
augustjk, line 35: Nit: I think the rest of the companies here are all alphabetically ordered, except JSTOR which was added to the end. Should it all be re-ordered?

augustjk (commented):
Thanks for working on this! This looks good to me.

packages/lit-dev-content/site/home/5-who-is-using.html
e111077: Good point, WDYT @justinfagnani

e111077: alphabetized!

augustjk (approved):
LGTM

e111077:
will happily address further suggestions in another PR

PR #1309, opened 2024-03-11
[docs] Update url for swapi api in controller example

augustjk:
Fixes issue reported here: https://github.com/lit/lit/discussions/4577


e111077 (approved):
we have swapi.dev at home!

PR #1315, opened 2024-03-12
[playground] New context example with element that both consumes and provides

augustjk:
Example is taken from the React doc https://react.dev/learn/passing-data-deeply-with-context
 which is also an excellent resource for describing what context does and when to use it.

My only concern with this is that it's such a direct copy of the React doc example, even if I attribute to them at the top of the preview page.

`my-heading.ts` gets a manually made `.js` file because we don't have a good transform for the `@consume()` decorator.

latest preview link: http://localhost:5415/playground/#sample=examples/context-consume-provide

### DIFF
--- a/packages/lit-dev-content/samples/examples/context-consume-provide/my-section.ts
+++ b/packages/lit-dev-content/samples/examples/context-consume-provide/my-section.ts
@@ -5,20 +5,36 @@ import {levelContext} from './level-context.js';
 
 @customElement('my-section')
 export class MySection extends LitElement {
-  // Serve as a context provider with an initial value of 1.
+  // Serve as a context provider with an initial level 1 and numbering prefix
+  // based on its numbering among siblings.
   private _provider = new ContextProvider(this, {
     context: levelContext,
-    initialValue: 1,
+    initialValue: {
+      level: 1,
+      prefix: String(this._siblingNumber) + '.',
+    },
   });
 
   // Consumes level context from a parent provider. If a parent provider is
-  // found, update own provider value to be 1 greater than provided value.
+  // found, update own provider level to be 1 greater than provided value and
+  // append its sibling number.
   private _consumer = new ContextConsumer(this, {
     context: levelContext,
-    callback: (value) => {
-      this._provider.setValue(value + 1);
+    callback: ({level, prefix}) => {
+      this._provider.setValue({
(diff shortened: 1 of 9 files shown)


e111077:
Discussed offline that I think it's not 100% best practice for a user to define a class field `_consumer` that is not used in the component, but we think it's best not to introduce a "new concept" of constructor + super() call when all the other examples are using class members

e111077 (approved):

augustjk:
Synced with @justinfagnani offline on how we might make the example more interesting and added a numbering prefix to all the headings.

We had originally talked about doing alternate numbering system based on level like roman numerals, alphabet, arabic number, etc. but I came across this https://typographyforlawyers.com/hierarchical-headings.html and figured simple is better.

packages/lit-dev-content/samples/examples/context-consume-provide/my-section.ts
```diff
@@ -5,30 +5,45 @@ import {levelContext} from './level-context.js';
...
-      this._provider.setValue(value + 1);
+    callback: ({level, prefix}) => {
+      this._provider.setValue({
+        level: level + 1,
+        prefix: prefix + String(this._siblingNumber) + '.',
+      });
     },
   });
 
+  private get _siblingNumber() {
```
e111077, line 31: I'd be a bit weary using this over an `<ol>` even if it's for the sake of example.

augustjk: Is it possible to do this kind of numbering with `<ol>`s?
image

augustjk: Ah, I guess it is possible... https://stackoverflow.com/a/56430931/19498660

e111077: AH sorry the page didn't finish rendering and I thought it was simple numbering. This looks good

augustjk: The point of this is you can do it across sections of your application though. The example doesn't show it but presumably there'll be content in between the headings and making that entire thing a `<li>` doesn't seem right? But of course the numbering can still be done with css counter so that does make this example less compelling.

PR #1320, opened 2024-04-02
theme-switcher is now a select

### LINKED ISSUE #1318: Dark Mode trigger icons are reversed if already in prefers-color-scheme(dark)
I think it should be a moon icon to say "switch to dark" and sun icon for "switch to light" but they appear to be reversed

e111077:
fixes #1318

theme switcher is now a select rather than a trinary toggle button

brotli bundle of first page goes from 419kb -> 433kb


augustjk (approved):
Nice

PR #1329, opened 2024-04-24
Updates task.md to include typescript `as const` tip.

collin:
Using `as const` with the return value of the args list creates a more useful, and narrower tuple type.

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/data/task.md
+++ b/packages/lit-dev-content/site/docs/v3/data/task.md
@@ -442,3 +442,41 @@ class MyElement extends LitElement {
 ```
 
 {% endswitchable-sample %}
+
+### TypeScript Tips
+
+Consider the following task, with two arguments.
+
+```ts
+class MyElement extends LitElement {
+  @property() myNumber = 10;
+  @property() myText = "Hello world";
+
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -442,3 +442,41 @@ class MyElement extends LitElement {
 ```
 
 {% endswitchable-sample %}
+
+### TypeScript Tips
```
justinfagnani, line 446: Since this just has one tip for now, I'd suggest:

```suggestion
### More accurate argument types in TypeScript
```

packages/lit-dev-content/site/docs/v3/data/task.md
```diff
@@ -442,3 +442,41 @@ class MyElement extends LitElement {
...
+
```
justinfagnani, line 447: I'd add an introduction sentence:

```suggestion
Task argument types can sometimes be inferred too loosely by TypeScript. This can be fixed by casting argument arrays with `as const`.
```

justinfagnani (approved):
Thanks @collin!

PR #1337, opened 2024-05-21
[example/until] replace call to jsonplaceholder with a static text

kadoshms:
Currently the `until` directive playground example is broken due to the reliance on jsonplaceholder which is not stable.
Instead, simply await a timeout of 2 seconds, and return some static text.


augustjk (approved):
Thank you for the PR! jsonplaceholder seems to be working fine for me for the moment, and I do like examples to have even a semblance of something close to a real example.. but reducing an external dependency probably is a good idea.

Perhaps down the line we can come up with an example that uses some async web API that doesn't depend on an external resource. Or we just add some mock data endpoint to lit.dev since we do have other examples that want to show case network requests.

PR #1338, opened 2024-05-23
[docs] make code samples in component lifecycle consistent

odanado:
In the description of `disconnectedCallback`, `window.removeEventListener` is used. However, the prefix `window.` is omitted in the `connectedCallback` description. To improve consistency between the two sample codes, `window.` has been added to the `connectedCallback` description.


augustjk (approved):
Thanks! I love me some consistency.

PR #1348, opened 2024-07-03
Add notes about separate unbundled bundles for ESbuild

KonnorRogers:
Hi there! I recently got Lit SSR setup with Web Awesome / Shoelace and wanted to share some snags I hit. Happy to document these somewhere else or change wording, just wanted to add some things I was missing until I found @e111077 's 11ty SSR Example and their ESBuild config.

https://github.com/e111077/11ty-lit/blob/204a7d3a060acced0573745d57147cbfe95abb1e/esbuild.config.mjs#L76-L86


augustjk:
Thanks for the PR! I reworded and reorganized the content a bit. We already mentioned conditional exports at the end of the browser-only code section so made that into a subheading and added the package.json example there.

Made the bundler section a bit more clear that the ideal path is to mark everything external, which won't require separate entrypoints (unless your own library code has node-specific things you want to do). And if you must bundle Lit into, that's when the two entrypoints are required.

/gcbrun

augustjk:
preview: https://pr1348-c911d2a---lit-dev-5ftespv5na-uc.a.run.app/docs/ssr/authoring/

KonnorRogers:
@augustjk thanks for that!! And hmm 🤔 maybe I gotta double check what I was doing with ESBuild if I was marking the initial components as having no externals and bundling in Lit. I'll report back when I get a shot to look again.

KonnorRogers:
@augustjk was extending another config and forgot to set `packages: "external"` in my esbuild config, which is why `platform: "node"` worked 🤦

augustjk (approved):

PR #1355, opened 2024-07-19
Lit cheat sheet

e111077:
See it here:

https://pr1355-e52fd31---lit-dev-5ftespv5na-uc.a.run.app/articles/lit-cheat-sheet/#adding-listeners-to-global-nodes

This PR also adds lazy loading to lit-example via intersection observer so that we do not load the 300kb TS worker a million times per page and so that it can cache.


justinfagnani:
Whoa, awesome!!!

I'm reading through now and my first/only comment is that some of the example are really long compared to their main topic. I think the Templating section has the most of this. eg, the event listener expression example could just be a single button and an output element (similar to the common counter example) without validation, etc.

It would be great to get in small Task and Context example too, and may an ElementInternals thing (custom state?).

justinfagnani:
I can definitely contribute to refining this branch! This is such a good start.

e111077:
let me debug get this working in the preview first!

e111077:
lol it was a real bug and not a flake. good job, tests!

filimon-danopoulos:
I've read through it and it will be a gold mine. I have the same thought as Justin about section length. 

One thing that stod out was the mention of running on the server in the lifecycle section. Seems a little premature given that the SSR package is still in labs.  I think that the lifecycle section could be compacted into a table instead. 

Would it be reasonable to include a short glossary section?

I have a couple of smaller suggestions with the text but I get the impression that we don't want to put these here but in a separate PR.

e111077:
Thanks for the look-over Filimon! The nice thing about articles is that they have more of a chance to live as a living document, so lifecycle reference can change over time. Though I do think that this information is very useful for people trying to actually use our SSR package and we should strive to make it easier to use. I also agree with you and think that there needs to be some sort of "labs" warning near that

JosefJezek:
Could you write a page about Lit vs React? With comparisons all Lit + Signal solutions instead of React hooks.

resources:
- https://codelabs.developers.google.com/codelabs/lit-2-for-react-devs#0
- https://react.dev/reference/react/hooks
- https://github.com/proposal-signals/signal-utils

e111077:
Can I get another review? I'd like to get this out before thanksgiving or by cyber monday

filimon-danopoulos:
I have looked it over but not at every detail as it is a lot to go through but this looks **really** good.

A common misconception is around converters is that they are run when the property is set. Your example does not make it worse as it is clear enough but maybe add a warning/note making that extra clear? 

I think that a pattern around derived state is probably nice to have to. Maybe with `get/set` and with lifecycle? By no means a requirement though and easy enough to add eventually.

I have not worked enough with signals to have meaningful feedback. 

Really well done!

e111077:
Thanks for the suggestion @filimon-danopoulos, I added that note in custom attribute converters and I also added 4 examples for:

- Derived Read-Only State
  - getters (nothing on setters for now)
- Reconcile Values Between Reactive Properties
  - `willUpdate()`
- Sync Reactive Property with Browser Features
  - `update()` RP + localstorage
- Reconcile Values Between Reactive Properties and DOM
  - `updated()`

sorvell:
> Sync Reactive Property with Browser Features
> update() RP + localstorage

Why update and not updated?

e111077:
@sorvell it happens before render and does not rely on the DOM. In particular the example uses `localStorage` and does some simple two-way syncing between the two

e111077:
ping @justinfagnani @sorvell @augustjk for a review

If ready to merge, we should also prepare socials as I no longer have access

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+
+### Turn off Shadow DOM
+
+You can turn off the Shadow DOM by overriding the `createRenderRoot()` method and setting the render root to the element itself.
+
+{% playground-ide "articles/lit-cheat-sheet/turn-off-shadow-dom", true %}
+
+{% aside "warn" %}
+
+By turning off shadow DOM you lose the benefits of encapsulation, DOM scoping, and `<slot>` elements.
```
sorvell, line 439:
```suggestion
This is generally not recommended, but it may sometimes be worthwhile for integration with older systems or libraries that may not be updated to work with Shadow DOM. 

By turning off shadow DOM you lose the benefits of encapsulation, DOM scoping, and `<slot>` elements.
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+
+Since the Shadow root no longer exists, Lit will no longer handle the `static styles` property for you. You must decide how to handle your styles.
```
sorvell, line 441:
```suggestion
Since the Shadow root no longer exists, `<slot>` does not work and Lit will no longer handle the `static styles` property for you. You must decide how to handle your styles.
```

packages/lit-dev-content/samples/articles/lit-cheat-sheet/reconcile-update/my-element.ts
```diff
@@ -0,0 +1,46 @@
...
+
+@customElement('my-element')
+export class MyElement extends LitElement {
+  @property() savedString = '';
+
+  update(changed: PropertyValues<this>) {
+    // localStorage is only available in the browser so we put this reactive
+    // property reconciliation in an update() method.
+    if (changed.has('savedString')) {
+      localStorage.setItem('savedString', this.savedString);
```
sorvell, line 13:
```suggestion
      window.localStorage.setItem('savedString', this.savedString);
```

packages/lit-dev-content/samples/articles/lit-cheat-sheet/reconcile-update/my-element.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    if (localStorage.get('savedString') !== this.savedString) {
```
sorvell, line 16:
```suggestion
    if (window.localStorage.getItem('savedString') !== this.savedString) {
```

packages/lit-dev-content/samples/articles/lit-cheat-sheet/reconcile-update/my-element.ts
```diff
@@ -0,0 +1,46 @@
...
+      this.savedString = localStorage.getItem('savedString') ?? '';
```
sorvell, line 17:
```suggestion
      this.savedString = window.localStorage.getItem('savedString') ?? '';
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+  possible.
+
+#### `disconnectedCallback`
+
+- Is called *AFTER* the element is removed from the DOM
+- A good place to clean up event listeners
+
+### Waiting for an update
+
+All Lit elements have asynchronous lifecycles. You need to wait for the `updateComplete` promise to resolve before you can be sure that the element has finished updating its DOM.
```
sorvell, line 780:
```suggestion
All Lit elements have asynchronous lifecycles. The reason for this is so that property changes (e.g. el.foo = 1; el.bar = 2;) are batched for efficiency and correctness. 

You need to wait for the `updateComplete` promise to resolve before you can be sure that the element has finished updating its DOM.
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+- Do **NOT** require parameters in the constructor or modify DOM
+
+#### `connectedCallback`
+
+- *May* run on the server. This has not been finalized.
+- Runs when the element is added to the DOM via:
+  - `element.appendChild(element)`
+  - `element.innerHTML = '<my-element></my-element>'`
+  - etc.
+- Can run multiple times, but a good place to set up event listeners
```
sorvell, line 744:
```suggestion
- Can run multiple times, but a good place to set up event listeners to external elements like `document`.
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+**Related Documentation & Topics:**
+
+- [Asynchronous Tasks](/docs/data/task/)
+- [Lit Labs – Task](https://www.youtube.com/watch?v=niWKuGhyE0M) (video)
+  - NOTE: This video was published before Task had graduated from Labs.
+- [Reactive Controllers – Asynchronous Tasks](/docs/composition/controllers/#asynchronous-tasks)
+
+## Events
+
+### Adding listeners to host element or globally
```
sorvell, line 821: I don't think we should encourage setting up event listeners on host in `connectedCallback`. This should generally be done in `constructor` and doesn't need cleanup. Promoting this potentially feeds FUD about memory leaks. Recommending changing the text here and the associated example to focus on listeners to external elements.

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+
+### Pass data down
+
+The simplest way to pass data down is to use properties and attributes.
+
+For example, you can pass data down to child components using property bindings like this:
+
+`.name=${'Steven'}`
+
+For boolean properties, use a question mark instead of a period, like this:
```
sorvell, line 844:
```suggestion
For boolean attributes, use a question mark instead of a period, like this:
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+
+### Other Methods to Get An Element Reference
+
+The `ref()` directive is a lit-html-specific method to acquire an element
+reference. The `ref()` directive is a good alternative when:
+
+- You can't use the `@query` decorator (or its JS equivalent)
+- You cannot determine when an element will be rendered
+- You need to pass element references from a child to a parent component (not common)
+- You're migrating from another framework like React
```
sorvell, line 928:
```suggestion
- You're migrating from another framework like React
- You need to run a function when the referenced element changes
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+effects, but signal changes used in `render()` or the Lit lifecycle will trigger
+the Lit lifecycle like any other Reactive Property. You can trigger effects in
+the Lit lifecycle by using signals in the same way you would use Reactive
+Properties.
+
+{% playground-ide "articles/lit-cheat-sheet/signals-effect", true %}
+
+{% aside "warn" %}
+
+Effects implemented in this way are not compatible with the `watch()` directive.
```
sorvell, line 1079: What does this mean? 

I think this section is confusing and should be limited to just referencing the effect in `signal-utils`.

packages/lit-dev-content/samples/articles/lit-cheat-sheet/bind-tag-name/input-or-textfield.ts
```diff
@@ -0,0 +1,35 @@
...
+export class MyElement extends LitElement {
+  // attribute is false because this is a value that can't be serialized to an
+  // HTML attribute
+  @property({ attribute: false }) tagLiteral: StaticValue|null = null;
+  @property() value = '';
+
+  render() {
+    return html`
+      ${
+        // NOTE: the live() directive prevents setting the .value property if
```
sorvell, line 16: I would remove `live` here and the comment. As long as the tag doesn't change, the expense is minimal.

packages/lit-dev-content/samples/articles/lit-cheat-sheet/bind-tag-name/input-or-textfield.ts
```diff
@@ -0,0 +1,35 @@
...
+        // the live value of the input / textfield already matches this.value.
+        // This is important since static html templates should not be thrashed
+        // due to performance concerns.
+        staticHTML`
+          <${this.tagLiteral}
```
sorvell, line 21: Why doesn't this use `unsafeStatic` or `literal`?

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1118 @@
...
+In rare cases, you need to bind to an HTML tag name to change the rendered
+element. You can do this safely with the `static-html` module and the `literal`
+template tag.
+
+{% playground-ide "articles/lit-cheat-sheet/bind-tag-name", true %}
+
+
+{% aside "warn" %}
+
+Be careful when using static HTML templates as changing values of a static
```
sorvell, line 219: They are cached but the caching is slightly more expensive. Instead I would just draw attention to the cost of changing element tags.

sorvell (changes requested):

packages/lit-dev-content/samples/articles/lit-cheat-sheet/bind-tag-name/input-or-textfield.ts
e111077: This uses literal in the parent element. Also unsafeStatic is handled in the next example

e111077:
@sorvell thanks for the review. I think the only part that _really_ needs a re-review are the changes to `effect()` since it's leaky and TIL about the `signal-utils` package

e111077:
quick ping @sorvell

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1131 @@
...
+
+### Common Signal Setup (SignalWatcher)
+
+The most common way to use signals in Lit is to use the `SignalWatcher` mixin.
+When an accessed signal value changes, `SignalWatcher` will trigger the Lit
+element update lifecycle. This includes signals read in `shouldUpdate()`,
+`willUpdate()`, `update()`, `render()`, `updated()`, `firstUpdated()`, and
+reactive controller's `hostUpdate()` and `hostUpdated()`.
+
+{% playground-ide "articles/lit-cheat-sheet/signal-watcher", true %}
```
augustjk, line 1022: the examples in the signal section that rely on `console.log()` to demonstrate when render is called or when effects are running are hard to appreciate without actually opening up the dev console.

how about splitting up the preview pane and "log" to a light DOM container?
[example](https://lit.dev/playground/#project=W3sibmFtZSI6Im15LWVsZW1lbnQudHMiLCJjb250ZW50IjoiaW1wb3J0IHtMaXRFbGVtZW50LCBodG1sfSBmcm9tICdsaXQnO1xuaW1wb3J0IHtjdXN0b21FbGVtZW50fSBmcm9tICdsaXQvZGVjb3JhdG9ycy5qcyc7XG5pbXBvcnQge1NpZ25hbFdhdGNoZXIsIHNpZ25hbCwgd2F0Y2h9IGZyb20gJ0BsaXQtbGFicy9zaWduYWxzJztcbmltcG9ydCB7bG9nfSBmcm9tICcuL2xvZ2dlci5qcyc7XG5cbmNvbnN0IGNvdW50ID0gc2lnbmFsKDApO1xuXG5AY3VzdG9tRWxlbWVudCgnbXktZWxlbWVudCcpXG5leHBvcnQgY2xhc3MgTXlFbGVtZW50IGV4dGVuZHMgU2lnbmFsV2F0Y2hlcihMaXRFbGVtZW50KSB7XG4gIHJlbmRlcigpIHtcbiAgICBsb2coJ3JlbmRlciBjYWxsZWQnKTtcblxuICAgIHJldHVybiBodG1sYFxuICAgICAgPHA-VGhlIGNvdW50IGlzICR7Y291bnQuZ2V0KCl9PC9wPlxuICAgICAgPGJ1dHRvbiBAY2xpY2s9JHt0aGlzLiNvbkNsaWNrfT5JbmNyZW1lbnQ8L2J1dHRvbj5cbiAgICBgO1xuICB9XG5cbiAgI29uQ2xpY2soKSB7XG4gICAgY291bnQuc2V0KGNvdW50LmdldCgpICsgMSk7XG4gIH1cbn1cblxuIn0seyJuYW1lIjoiaW5kZXguaHRtbCIsImNvbnRlbnQiOiI8c2NyaXB0IHR5cGU9XCJtb2R1bGVcIiBzcmM9XCIuL215LWVsZW1lbnQuanNcIj48L3NjcmlwdD5cblxuPHN0eWxlPlxuICBtYWluIHtcbiAgICBkaXNwbGF5OiBmbGV4O1xuICB9XG4gIFxuICBtYWluID4gKiB7XG4gICAgZmxleDogMVxuICB9XG4gIFxuICBwIHtcbiAgICBtYXJnaW46IDA7XG4gICAgZm9udC1mYW1pbHk6IG1vbm9zcGFjZTtcbiAgfVxuPC9zdHlsZT5cbjxtYWluPlxuPG15LWVsZW1lbnQ-PC9teS1lbGVtZW50PlxuPGRpdiBpZD1cImxvZ1wiPlxuICA8Yj5Mb2c6PC9iPlxuICA8aHIvPlxuPC9kaXY-XG48L21haW4-In0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMi4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjQuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4zLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9LHsibmFtZSI6ImxvZ2dlci50cyIsImNvbnRlbnQiOiJjb25zdCBsb2dDb250YWluZXIgPSBkb2N1bWVudC5nZXRFbGVtZW50QnlJZCgnbG9nJykhO1xuXG5leHBvcnQgZnVuY3Rpb24gbG9nKHRleHQ6IHN0cmluZykge1xuICBjb25zdCBwID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgncCcpO1xuICBwLmlubmVyVGV4dCA9IHRleHQ7XG4gIGxvZ0NvbnRhaW5lci5hcHBlbmRDaGlsZChwKTtcbn0ifV0)

augustjk (commented):
this is incredible work! a lot of great stuff here.

packages/lit-dev-content/samples/articles/lit-cheat-sheet/signals-effect/my-element.ts
```diff
@@ -0,0 +1,33 @@
...
+// NOTE: `html` is imported from the `@lit-labs/signals` package.
+import {SignalWatcher, signal} from '@lit-labs/signals';
+import {effect} from 'signal-utils/subtle/microtask-effect';
+
+const count = signal(0);
+
+@customElement('my-element')
+export class MyElement extends SignalWatcher(LitElement) {
+  @state() reactiveProp = 0;
+  #loggerEffect = effect(() => {
```
sorvell, line 12:
```suggestion
  #disposeLoggerEffect = effect(() => {
```

packages/lit-dev-content/samples/articles/lit-cheat-sheet/signals-effect/my-element.ts
```diff
@@ -0,0 +1,33 @@
...
+    console.log('Do effect here!');
+    console.log('The current count is:', count.get());
+  });
+
+  render() {
```
sorvell, line 17:
```suggestion
  disconnectedCallback() {
    this.#disposeLoggerEffect();
  }

  render() {
```

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
```diff
@@ -0,0 +1,1131 @@
...
+`effect()` function that allows you to react to signal changes and run side
+effects.
+
+{% playground-ide "articles/lit-cheat-sheet/signals-effect", true %}
+
+{% aside "warn" %}
+
+The `effect()` function from the `signal-utils` package is experimental.
+
+The current implementation as of December 2024 does not have a way to clean up
```
sorvell, line 1094: I don't think this is the case, see https://github.com/proposal-signals/signal-utils/blob/main/src/subtle/microtask-effect.ts#L38.

You just have to call the returned function yourself. Let's change this to remind users to do this in disconnected. I also changed the example to do this.

sorvell (approved):
Approving but please do update the section on effect.

packages/lit-dev-content/site/articles/article/lit-cheat-sheet.md
e111077: huh idk why I was running into this issue before

packages/lit-dev-content/samples/articles/lit-cheat-sheet/event-listeners/my-element.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@customElement('my-element')
+export class MyElement extends LitElement {
+  @state() inputValue = '';
+  @query('input') inputEl!: HTMLInputElement;
+
+  render() {
+    return html`
+      <div>
+        <input .value=${this.inputValue} />
```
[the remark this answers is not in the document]
e111077: oops!

packages/lit-dev-content/samples/articles/lit-cheat-sheet/with-logger.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "/samples/base.json",
+  "files": {
+    "_visual-logger.ts": {
+      "hidden": true,
+      "content": "import {LitElement, html, css} from 'lit';\nimport {customElement} from 'lit/decorators.js';\nimport {SignalWatcher, signal, watch} from '@lit-labs/signals';\n\nconst log = signal<unknown[][]>([]);\n\nconst oldLog = window.console.log\nwindow.console.log = (...values: unknown[]) => {\n  log.set([...log.get(), values]);\n  oldLog(...values)\n};\n\nconst globalStyles = new CSSStyleSheet();\n\nglobalStyles.replaceSync(`body {\n  display: flex;\n}\n  \nbody > * {\n  flex: 1\n}`)\n\ndocument.adoptedStyleSheets.push(globalStyles);\n\n@customElement('visual-logger')\nexport class MyElement extends SignalWatcher(LitElement) {\n  render() {\n    return html`\n      <div id=\"log\">\n        <b>Log:</b>\n        <hr/>\n          ${log.get().reverse().map(val => html`<p>${val.join(' ')}</p>`)}\n      </div>\n    `;\n  }\n  \n  static styles = css`\n    #log {\n      display: block;\n      max-height: 100dvh;\n    }\n      \n    p {\n      margin: 0;\n      font-family: monospace;\n    }\n  `;\n}"
```
augustjk, line 6: i was seeing some ordering issue in the logs for the effect example because the array was being mutated with `.reverse()`. assuming you want latest log to be on the top, using `.toReversed()` does not mutate the original array.
```suggestion
      "content": "import {LitElement, html, css} from 'lit';\nimport {customElement} from 'lit/decorators.js';\nimport {SignalWatcher, signal, watch} from '@lit-labs/signals';\n\nconst log = signal<unknown[][]>([]);\n\nconst oldLog = window.console.log\nwindow.console.log = (...values: unknown[]) => {\n  log.set([...log.get(), values]);\n  oldLog(...values)\n};\n\nconst globalStyles = new CSSStyleSheet();\n\nglobalStyles.replaceSync(`body {\n  display: flex;\n}\n  \nbody > * {\n  flex: 1\n}`)\n\ndocument.adoptedStyleSheets.push(globalStyles);\n\n@customElement('visual-logger')\nexport class MyElement extends SignalWatcher(LitElement) {\n  render() {\n    return html`\n      <div id=\"log\">\n        <b>Log:</b>\n        <hr/>\n          ${log.get().toReversed().map(val => html`<p>${val.join(' ')}</p>`)}\n      </div>\n    `;\n  }\n  \n  static styles = css`\n    #log {\n      display: block;\n      max-height: 100dvh;\n    }\n      \n    p {\n      margin: 0;\n      font-family: monospace;\n    }\n  `;\n}"
```

augustjk (approved):
🎉

PR #1356, opened 2024-07-19
feat: add highlight colors!

e111077:
shamelessly stolen from angular docs site

Yea or nay?!?!?!? It doesn't penetrate shadow roots


augustjk (approved):
Neat! I really like it. The different colors at different orders is a nice touch. The only downside is I can't tell if leading or trailing whitespaces are selected or not but who cares!

soumyasr-google:
> The different colors at different orders is a nice touch

FWIW, I find this very confusing; I spent an inordinate amount of time tracking down this PR to see if there was an actual reason for having multiple different selection colours that very from paragraph to paragraph, from list item to list item.

image

PR #1380, opened 2024-10-28
Mention that updating to TypeScript 5.2 is necessary in the upgrade guide

rictic:


rictic:
Based on feedback from the model viewer folks as part of their Lit 3 update.

packages/lit-dev-content/site/docs/v3/releases/upgrade.md
```diff
@@ -23,8 +23,9 @@ Lit 3.0 has very few breaking changes from Lit 2.x:
 - APIs marked deprecated in Lit 2.x releases have been removed.
 - SSR hydration support modules have moved to the `@lit-labs/ssr-client` package.
 - Type only: Type of `ReactiveElement`'s `renderRoot` and `createRenderRoot()` have been updated.
-- Decorator behavior has been unified between TypeScript experimental decorators and standard decorators.
 - Support was removed for Babel decorators version "2018-09".
+- Decorator behavior has been unified between TypeScript experimental decorators and standard decorators.
+  - As a consequence, if you use TypeScript you'll need to upgrade to at least TypeScript v5.2, to get the updated types for both kinds of decorators.
```
justinfagnani, line 28: Or use `"skipLibCheck": true`, right? This is just because of the reference to the decorator types in Lit?

rictic: Yeah, exactly.

You could use `skipLibCheck: true`, but I wouldn't recommend that. Easy to get implicit `any` when it encounters a type it doesn't know about.

justinfagnani: I was just wondering if that should be mentioned as a workaround in the upgrade guide. Up to you.

PR #1390, opened 2025-01-10
[docs] Add documentation for SSR event handling

kyubisation:
This PR adds the relevant documentation for the functionality in lit/lit#4755

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/data/context.md
+++ b/packages/lit-dev-content/site/docs/v3/data/context.md
@@ -281,6 +281,21 @@ and the ContextConsumer controller:
   );
 ```
 
+### SSR
+
+Basic support for server side rendering can be enabled by assigning
+`globalThis.litSsrCallConnectedCallback = true`. This will configure `@lit-labs/ssr` to call
+`connectedCallback()` and `ReactiveController.hostConnected()`, which allows `@lit/context`
+to register the necessary event handlers for SSR.
+
+As an alternative to `document.documentElement` or `document.body` (which are expected to be undefined
+in the server environment) for global event listeners (e.g. for `@lit/context ContextProvider`), you can
+use the global variable `globalThis.litServerRoot` which is (only) available during SSR.
+
+```ts
+  new ContextProvider(isServer ? globalThis.litServerRoot : document.body, {...});
+```
+
 ## Example Use Cases
 
 ### Current user, locale, etc.
--- a/packages/lit-dev-content/site/docs/v3/ssr/authoring.md
+++ b/packages/lit-dev-content/site/docs/v3/ssr/authoring.md
@@ -103,7 +103,7 @@ Whether a method is called on the server is subject to change while Lit SSR is p
 | Method | Called on server | Notes |
 |-|-|-|
 | `constructor()` | Yes ⚠️ | |
-| `connectedCallback()` | No | |
+| `connectedCallback()` | No | Can be enabled with `globalThis.litSsrCallConnectedCallback = true` |
 | `disconnectedCallback()` | No | |
 | `attributeChangedCallback()` | No | |
 | `adoptedCallback()` | No | |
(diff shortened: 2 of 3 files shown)


packages/lit-dev-content/site/docs/v3/data/context.md
```diff
@@ -281,6 +281,21 @@ and the ContextConsumer controller:
...
+`globalThis.litSsrCallConnectedCallback = true`. This will configure `@lit-labs/ssr` to call
+`connectedCallback()` and `ReactiveController.hostConnected()`, which allows `@lit/context`
+to register the necessary event handlers for SSR.
+
+As an alternative to `document.documentElement` or `document.body` (which are expected to be undefined
+in the server environment) for global event listeners (e.g. for `@lit/context ContextProvider`), you can
+use the global variable `globalThis.litServerRoot` which is (only) available during SSR.
+
+```ts
+  new ContextProvider(isServer ? globalThis.litServerRoot : document.body, {...});
```
bennypowers, line 296: the overloaded constructor signature for ContextProvider is

```ts
  constructor(host: HostElement, options: Options<T>);
  /** @deprecated Use new ContextProvider(host, options) */
  constructor(host: HostElement, context: T, initialValue?: ContextType<T>);
  constructor(
    host: HostElement,
    contextOrOptions: T | Options<T>,
    initialValue?: ContextType<T>
  ) { /*...* }
```
ContextRoot does not have any constructor params.

Do you mean something like this?

```ts
  const root = new ContextRoot();
  if (!isServer) {
    root.attach(document.body);
  } else {
    root.attach(globalThis.litServerRoot);
  }
```

kyubisation: No, I am referring to the use case of registering a ContextProvider globally. (See https://www.npmjs.com/package/@lit/context and search for `document.body`)
ContextRoot _should_ not be necessary in an SSR environment, as classes are loaded in order (from my understanding).

But thank you for pointing out, that my example is incomplete and easily misunderstood. I will extend it. 😄 👍

packages/lit-dev-content/site/docs/v3/data/context.md
```diff
@@ -281,6 +281,21 @@ and the ContextConsumer controller:
   );
 ```
 
+### SSR
```
e111077, line 284: I think this section needs to focus a bit more on the information of "I am googling around and I just want to get this thing done where can I copy paste a solution" and then follow up with the explanation. So like questions a user might have:

- Does Context work in SSR?
- If so how do I do it?
- I am running into X error when I try to use Context in SSR, how do I fix it?
- Why is this the solution? (probably lower priority question they have)

So maybe rephrasing a bit backwards might help e.g.

> // answer the questions
> Lit SSR provides basic support for global and component-scoped context. Context in SSR requires that `globalThis.litSsrCallConnectedCallback = true` to be enabled. (insert link to docs on that)
>
> // give a copy-pastable example
> ```ts
> // server.ts
> globalThis.litSsrCallConnectedCallback = true; // or however / wherever you're supposed to set this
> ...
>
> // client.ts
> import {isServer} from 'lit';
> import {ContextProvider} from '@lit/context';
> const globalRoot = isServer ? globalThis.litServerRoot : document.body;
> new ContextProvider(globalRoot, {...});
> ```
>
> // then include wonky explanation why
> As DOM nodes like `document.body` are not typically available in the server environment, Lit SSR will provide `globalThis.litServerRoot` for event-emitting purposes. Additionally, Lit Context provides setup in the native web component `connectedCallback` lifecycle callback, so `globalThis.litSsrCallConnectedCallback` needs to be set to true to enable Lit SSR to include that callback in the server lifecycle.
>
> To use Lit Context in a component, you can do the following:
>
> // copy-pastable example
> ```ts
> // I actually don't know, please provide
> ```
>
> // wonky explanation why the code above is like that

packages/lit-dev-content/site/docs/v3/ssr/authoring.md
```diff
@@ -103,7 +103,7 @@ Whether a method is called on the server is subject to change while Lit SSR is p
 | Method | Called on server | Notes |
 |-|-|-|
 | `constructor()` | Yes ⚠️ | |
-| `connectedCallback()` | No | |
+| `connectedCallback()` | No | Can be enabled with `globalThis.litSsrCallConnectedCallback = true` |
```
e111077, line 106: Can you please give an example of this in this document?

e111077 (changes requested):

PR #1394, opened 2025-01-29
Add task behavior of empty args array

TownCube:
It wasn't clear to me how to create a no arguments task that auto runs.

I found a reference for how it works in https://github.com/lit/lit/pull/2336 and thought it would be worth including in the documentation to clarify the different handling between no arguments and an empty array of arguments.


e111077 (approved):
awaiting preview

PR #1396, opened 2025-02-04
Add docs for defaultValue

sorvell:
Adds docs for https://github.com/lit/lit/pull/4895. Publish on after that release.

### DIFF
--- a/packages/lit-dev-content/samples/properties/attributereflect/my-element.ts
+++ b/packages/lit-dev-content/samples/properties/attributereflect/my-element.ts
@@ -1,4 +1,4 @@
-import {LitElement, html, css} from 'lit';
+import {LitElement, html, css, PropertyDeclaration} from 'lit';
 import {customElement, property} from 'lit/decorators.js';
 
 @customElement('my-element')
@@ -6,19 +6,35 @@ class MyElement extends LitElement {
   @property({type: Boolean, reflect: true})
   active: boolean = false;
 
+  @property({type: String, reflect: true, defaultValue: 'normal'} as PropertyDeclaration)
+  variant!: string|null;
+
   static styles = css`
     :host {
-      display: inline-block;
+      display: inline-block; padding: 4px;
     }
-
     :host([active]) {
-      border: 1px solid red;
+      font-weight: 800;
+    }
+    :host([variant]) {
+      outline: 4px solid green;
+    }
+    :host([variant="special"]) {
+      border-radius: 8px; border: 4px solid red;
     }`;
 
   render() {
     return html`
-      <span>Active: ${this.active}</span>
-      <button @click="${() => this.active = !this.active}">Toggle active</button>
+      <div><label>active: <input type="checkbox"
+        .value="${this.active}"
+        @change="${(e: Event) => this.active = (e.target! as HTMLInputElement).checked}">
+        ${this.active}
+      </label></div>
+      <div><label>variant: <input type="checkbox"
+        .value="${this.variant === 'special'}"
+        @change="${(e: Event) => this.variant = (e.target! as HTMLInputElement).checked ? 'special' : 'normal'}">
+        ${this.variant}
+      </label></div>
     `;
   }
 }
--- a/packages/lit-dev-content/site/docs/v3/components/properties.md
+++ b/packages/lit-dev-content/site/docs/v3/components/properties.md
@@ -129,10 +129,13 @@ class MyElement extends LitElement {
 }
 ```
 
-In **JavaScript**, you **must not use class fields** when declaring reactive properties. Instead, properties must be initialized in the element constructor:
+In **JavaScript**, you **must not use class fields** when declaring reactive properties. Instead, properties must be either initialized in the element constructor or set with the `defaultValue` option:
 ```js
 class MyElement extends LitElement {
-  static properties = {foo: {type: String}}
+  static properties = {
+    foo: {type: String},
+    bar: {type: String, defaultValue: 'Default'}
+  }
   constructor() {
     super();
     this.foo = 'Default';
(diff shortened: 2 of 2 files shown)


packages/lit-dev-content/samples/properties/attributereflect/my-element.ts
```diff
@@ -1,24 +1,40 @@
-import {LitElement, html, css} from 'lit';
+import {LitElement, html, css, PropertyDeclaration} from 'lit';
 import {customElement, property} from 'lit/decorators.js';
 
 @customElement('my-element')
 class MyElement extends LitElement {
   @property({type: Boolean, reflect: true})
   active: boolean = false;
 
+  @property({type: String, reflect: true, defaultValue: 'normal'} as PropertyDeclaration)
```
justinfagnani, line 9: Is this cast only needed until https://github.com/lit/lit/pull/4895 is released?

packages/lit-dev-content/site/docs/v3/components/properties.md
```diff
@@ -129,10 +129,13 @@ class MyElement extends LitElement {
 }
 ```
 
-In **JavaScript**, you **must not use class fields** when declaring reactive properties. Instead, properties must be initialized in the element constructor:
+In **JavaScript**, you **must not use class fields** when declaring reactive properties. Instead, properties must be either initialized in the element constructor or set with the `defaultValue` option:
```
justinfagnani, line 132: While you're here "Instead" reads a little unclearly to me. I think saying "to initialize properties" would help somewhere. Also, "JavaScript" now has decorators via Babel at least, so maybe we update that phrase too.

Maybe:

```suggestion
If you are declaring reactive properties with the static properties class field, and not using decorators, you **must not use class fields** for those properties. Class fields will overwrite the properties in the static properties class field, disabling reactivity. To initialize a reactive property, do so in the element constructor or set the `defaultValue` option:
```

justinfagnani (approved):
🎉

packages/lit-dev-content/site/docs/v3/components/properties.md
```diff
@@ -553,11 +566,16 @@ If this behavior doesn't fit your use case, there are a couple of options:
 
 You can configure a property so that whenever it changes, its value is reflected to its [corresponding attribute](#observed-attributes). Reflected attributes are useful because attributes are visible to CSS, and to DOM APIs like `querySelector`.
 
+Setting `useDefault` to true prevents the property's default value from initially reflecting to the corresponding attribute. All subsequent changes will be reflected. This matches web platform behavior for attributes like `id`. For example, the default value of an element's `id` property is `''` (an empty string) and initially it does not have an `id` attribute, but if the `id` property is set (even to an empty string), the appropirate `id` attribute is reflected. In addition, if the attribute is removed, the property will be reset to its default value.
```
justinfagnani, line 569: Need to add a bit about how removing an attribute restores the property to the default. It'd be good to sync the language here and the jsdocs for the option a bit.

justinfagnani (changes requested):

packages/lit-dev-content/site/docs/v3/components/properties.md
sorvell: I think the preview may be broken because the property options in the example use `useDefualt` and that's not published yet.

packages/lit-dev-content/site/docs/v3/components/properties.md
```diff
@@ -551,29 +564,44 @@ If this behavior doesn't fit your use case, there are a couple of options:
...
+
+To ensure elements behave as expected and perform well, try to follow these best practices when reflecting attributes:
+
+* Attributes should generally be considered input to the element from its owner, rather than under control of the element itself, so reflecting properties to attributes should be done sparingly. Consider instead using the [`:state` pseudo selector](https://wicg.github.io/custom-state-pseudo-class/) and the [Accessibility Object Model](https://wicg.github.io/aom/spec/) where possible.
+
+* Reflecting properties should typically also set `useDefault` since this helps match expected platform behavior.
+
+* Reflecting properties of type object or array is not recommended. This can cause large objects to serialize to the DOM which can result in poor performance and consume excess memory when `useDefault` is used.
+  
+* In general user property settings should not be changed, but if you want to do so, define a property setter. Setting `useDefault` will only reset the property value to the default when the corresponding attribute is removed.
```
justinfagnani, line 603: "In general user property settings" reads a little ambiguously to me. "Settings" sounds like a noun like "app settings", not like a gerund.

```suggestion
* The property decorator does not alter any values assigned to the reactive property, which is considered a best practice for custom accessors. Sometimes native elements restrict properties to certain valid values, for instance, and if an invalid value is assigned to a property, the property will be set to a default instead. `useDefault: true` does not do this - it only restores the default when the attribute is removed. If you'd like to alter the property value on property assignments, define and decorate a custom property setter.
```

packages/lit-dev-content/site/docs/v3/components/properties.md
sorvell: Replaced cast to fix prview

packages/lit-dev-content/site/docs/v3/components/properties.md
```diff
@@ -269,6 +271,17 @@ When converting a string-valued attribute into a property, Lit's default attribu
...
 
+</dd>
+<dt id="use-default">
+
+`useDefault`
+
+</dt>
+<dd>
+
+Set to true to prevent initial attribute reflection for the default value when `reflect` is set to true, and to reset the property to its default value when its corresponding attribute is removed. For more information, see [Enabling attribute reflection](#reflected-attributes).
```
justinfagnani, line 282: Do we want a specific note here that the initial value is used as the default, and that it's retained in memory?

sorvell: Added

PR #1397, opened 2025-02-07
[search] add search keyword modifiers

e111077:
Add a method to attach arbitrary keywords or search terms to specific URLs in search index. See changes to CONTRIBUTING.md for authoring details. We used to have AI synonyms in Algolia, but we downgraded from algolia pro to algolia basic because we did not use the majority of the AI features because we did not send click events back to them. This is close-ish enough

Also added a few custom entries to the search index based on algolia usage stats.

To see output, check:

 https://pr1397-460b025---lit-dev-5ftespv5na-uc.a.run.app/search-modifiers/keywords.json
 
 to see keywords output and see
 
 https://pr1397-460b025---lit-dev-5ftespv5na-uc.a.run.app/searchIndex.json
 
 and CMD+F for `"keywords":`.
 
 I also tested this works on prod by editing the algolia search data on the live site with "keywords" and it didn't break anything and things started showing up as expected (I've since reverted those edits manually)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,6 +87,30 @@ Sometimes we want to include something that is not on lit.dev in the search inde
 }
 ```
 
+### How to associate keywords to a search item
+
+Sometimes the text of a search item does not include certain keywords you'd like to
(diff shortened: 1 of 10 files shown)


augustjk (approved):
neat!

PR #1398, opened 2025-02-07
[search] update algolia version and send clicks

e111077:
Updated algolia search version, and introduced search-insights for anonymized click analytics. Specifically does not use cookies and does not store any identifiable data. Reason for introduction is because I realized I am only updating our site based on searches with no results, but not searches with no clicks.

This causes the problem that we had results for `componentDidMount` which linked to this page https://lit.dev/docs/frameworks/react/#how-it-works which is a bad result. Something like that could have been caught with click data.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -327,6 +327,50 @@
         "@algolia/cache-common": "4.14.2"
       }
     },
+    "node_modules/@algolia/client-abtesting": {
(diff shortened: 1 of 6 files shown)


augustjk (approved):
looks like existing search hasn't broken and i see click event from search result getting sent.

PR #1403, opened 2025-02-26
[docs] update ContextRequestEvent signature

nkabrown:
Document changes to `ContextRequestEvent` in @lit/context 

https://github.com/lit/lit/pull/4734
https://github.com/lit/lit/blob/main/packages/context/src/lib/context-request-event.ts#L51-L65

### DIFF
--- a/packages/lit-dev-content/site/docs/v3/data/context.md
+++ b/packages/lit-dev-content/site/docs/v3/data/context.md
@@ -503,6 +503,10 @@ The `context-request` bubbles and is composed.
 
     The context object this event is requesting a value for
 
+- `readonly contextTarget: Element`
+
+    The original context target of the requester
+
 - `readonly callback: ContextCallback<ContextType<C>>`
 
     The function to call to provide a context value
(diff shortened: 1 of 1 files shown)


packages/lit-dev-content/site/docs/v3/data/context.md
```diff
@@ -503,6 +503,10 @@ The `context-request` bubbles and is composed.
 
     The context object this event is requesting a value for
 
+- `readonly contextTarget: Element`
+
+    The original context target of the requester
```
e111077, line 508: Personally not a fan of using `context target` to describe `contextTarget`. What about something like:

> The DOM element that initiated the context request

e111077 (commented):
Thanks for the submission, quick suggestion

e111077:
Also don't mind the failing CI checks. Pretty sure this happens with forks

packages/lit-dev-content/site/docs/v3/data/context.md
nkabrown: I will open a PR on @lit/context to update the param description there so this won't get overwritten when the API docs are automatically generated.

e111077 (approved):

e111077:
Thanks for the quick turnaround! 🚀

e111077:
deployed!

PR #1408, opened 2025-03-15
fix banner icon size on safari

e111077:
In safari the icon is HUGE

image


sorvell (approved):

e111077:
thanks! just gotta update the screenshot test

PR #1417, opened 2025-05-02
update wireit and github action cache setup version

augustjk:
addresses https://github.com/google/wireit/issues/1297


justinfagnani (approved):
Thanks!!

PR #1433, opened 2025-09-03
Update ing.svg

DieBauer:
Make sure the correct coloring (contrast) is used on the black-white logo of ING.

Before: 
image

After:
image


DieBauer:
> heya, do you mind checking if you signed the Google CLA?

Checked that thanks for the heads-up

e111077 (approved):

PR #1437, opened 2025-10-01
fix(playground): fix playground issues on safari

e111077:
Fix the playground issues on safari by upgrading to codemirror 6 which is playground elements ~0.21


e111077:
oh did i not push that change up?

e111077:
https://github.com/google/playground-elements/pull/433

augustjk (approved):
works good from my testing on Safari 26.0.1

PR #1442, opened 2025-10-17
Fix the manual task examples to specify `autoRun: false`

julienw:
I think this was missing, tell me if _I_'m missing anything :-)


julienw:
(started the CLA dance internally)

Issue #1429, opened 2025-08-21
Playground  Motion "Todos List Transitions" is broken


timker:
https://lit.dev/playground/#sample=examples/motion-todos

seems to be missing/looking for the wrong versions of node_modules/@material components

it's looking for 0.25.1
https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/samples/examples/motion-todos/project.json

but maybe 0.27.0 is installed
https://github.com/lit/lit.dev/blob/main/packages/lit-dev-content/package.json

binlabs:
Just noticed this as well, still broken. I tried MWC 0.27.0 in the playground as well, and that also didn't work. I think I might try to get this running locally and see if I can tell what's going on.

PR #1450, opened 2025-12-24
Fix motion todo example by removing mwc components

binlabs:
The [Motion: Todo List Transitions example](https://lit.dev/playground/#sample=examples/motion-todos) has broken for several months now due to 404s that are occurring when pulling in the Material Web components that this example uses.

See #1429

I tried bumping the version of those packages but faced the same error.

In this PR I simply replaced the MWC elements with their native equivalents. I tried to make the minimum possible change to get this example working again.

Screenshot 2025-12-23 at 10 44 25 PM

Tested and working on Mac in recent versions of Firefox, Safari, Edge and Chrome.

I did have two questions:
1. This example was the only place in the entire repo that uses mwc checkbox, formfield, and textfield. Should I remove those 3 dependencies and regenerate the main package lock?
2. Should I add a little styling to make it look a tiny bit better (e.g. styling to make it more obvious what the remove button does?

### DIFF
--- a/packages/lit-dev-content/samples/examples/motion-todos/motion-todos.ts
+++ b/packages/lit-dev-content/samples/examples/motion-todos/motion-todos.ts
@@ -4,12 +4,6 @@ import {classMap} from 'lit/directives/class-map.js';
 import {repeat} from 'lit/directives/repeat.js';
 import {animate, fadeOut, flyBelow} from '@lit-labs/motion';
 import {styles} from './styles.js';
-import {TextField} from '@material/mwc-textfield';
-import {Checkbox} from '@material/mwc-checkbox';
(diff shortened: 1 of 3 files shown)


binlabs:
It just occurred to me the reason why the MWC components weren't working - they probably just need to be updated to use the newer `@material/web` package.

If you'd rather I do that instead of use native elements like I did in this PR, let me know and I can make that change.

justinfagnani:
Thanks for looking at this @binlabs!

I'm not sure why existing packages would break like this, that's unfortunate.

I personally would go for fewer dependencies so there's less likely to be a break like this in the future. We can add some minimal styles to make the native controls look nice for the demo (which maybe you did already).

@sorvell wdyt?

